PR #3, opened 2017-11-19
Initial README – Take 2!

gr2m:
I’ve addressed the points mentioned in #1. I’ll move forward with it unless you have any objections

# [Preview](https://github.com/gr2m/octokit-webhooks/blob/initial-readme-take-2/README.md) 👀 

ping @bkeepers, @anglinb, @JasonEtco, @hiimbex, @zeke

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,669 @@
-# @octokit/webhooks.js
+# @octokit/webhooks
 
-> GitHub webhooks toolset
+> GitHub webhook events toolset for Node.js
+
+[![Build Status](https://travis-ci.org/octokit/webhooks.js.svg?branch=master)](https://travis-ci.org/octokit/webhooks.js)
+[![Coverage Status](https://coveralls.io/repos/octokit/webhooks.js/badge.svg?branch=master)](https://coveralls.io/github/octokit/webhooks.js?branch=master)
+[![Greenkeeper badge](https://badges.greenkeeper.io/octokit/webhooks.js.svg)](https://greenkeeper.io/)
+
+[GitHub webhooks](https://developer.github.com/webhooks/) can be registered in multiple ways
+
+1. In repository or organization settings on [github.com](https://github.com/).
+2. Using the REST API for [repositories](https://developer.github.com/v3/repos/hooks/) or [organizations](https://developer.github.com/v3/orgs/hooks/)
+3. By installing a [GitHub App](https://developer.github.com/apps/).
+
+`@octokit/webhooks` helps to handle webhook events received from GitHub.
+
+Note that while setting a secret is optional on GitHub, it is required to be set in order to use `@octokit/webhooks`. Content Type must be set to `application/json`, `application/x-www-form-urlencoded` is not supported.
+
+## Example
+
+```js
+// install with: npm install @octokit/webhooks
+const WebhooksApi = require('@octokit/webhooks')
+const webhooks = new WebhooksApi({
+  secret: 'mysecret'
+})
+
+webhooks.on('*', (data, {id, name}) => {
+  console.log(name, 'event received')
+})
+
+require('http').createServer(webhooks.middleware).listen(3000)
+// can now receive webhook events at port 3000
+```
+
+## API
+
+1. [Constructor](#constructor)
+2. [webhooks.sign()](#webhookssign)
+3. [webhooks.verify()](#webhooksverify)
+4. [webhooks.receive()](#webhooksreceive)
+5. [webhooks.on()](#webhookson)
+6. [webhooks.removeListener()](#webhooksremoveListener)
+7. [webhooks.middleware()](#webhooksmiddleware)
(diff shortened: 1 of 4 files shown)


bkeepers (approved):

README.md
```diff
@@ -1,3 +1,636 @@
...
+
+Returns a `requestListener` (or _middleware_) method which can be directly passed to [`http.createServer()`](https://nodejs.org/docs/latest/api/http.html#http_http_createserver_requestlistener), <a href="http://expressjs.com/">Express</a> and other compatible Node.js server frameworks.
+
+Can also be used [standalone](middleware/).
+
+### List of all event names
+
+See the full list of [event types with example payloads](https://developer.github.com/v3/activity/events/types/).
+
+If there are actions for a webhook, events are emitted for both, the webhook name as well as a combination of the webhook name and the action, e.g. `installation` and `installation.created`. A `*` event is always emitted. If an event handler throws an error, and `error` event is triggered.
```
JasonEtco: Typo here:

> error, and `error` event

Should be "an `error` event"

JasonEtco:
@gr2m this all looks great ✨ 

I saw you used HTML tables - I guess you prefer that over the Markdown table syntax? You could reduce the size of the README, might be a better way to go about it.

gr2m:
I use both. I think the markdown syntax does not allow for `valign` and `align`? And sometimes it just becomes to complex / too hard to read when everything is in one line. But when I use Markdown I like using http://www.tablesgenerator.com/markdown_tables and leave a comment like 

```
<!-- use http://www.tablesgenerator.com/markdown_tables to update the table below -->
```

JasonEtco:
Cool 👍 Whatever suits you best.

gr2m:
I got tests running with probot, I’ll push the code and start a PR there soon.

As I worked on it, I’ve run into the current `robot.receive` method. I like "receive" better than "handle". But renaming it would be a bit confusing when using the [standalone receiver](https://github.com/gr2m/octokit-webhooks/tree/initial-readme-take-2/receiver) module.

To work around that I’m thinking to rename that module to `event-handler` and rename the method from `.handle` to `.receive`. So the standalone "receiver" usage would change from

```js
const Receiver = require('@octokit/webhooks/receiver')
const receiver = new Receiver({secret: 'mysecret'})

receiver.on('installation', asyncInstallationHook)

// put in webhooks route handler
receiver.handle({
  name: request.headers['X-GitHub-Event'],
  data: request.body,
  signature: request.headers['X-Hub-Signature']
}).catch(handleErrorsFromHooks)
```

to

```js
const EventHandler = require('@octokit/webhooks/event-handler')
const eventHandler = new EventHandler({secret: 'mysecret'})

eventHandler.on('installation', asyncInstallationHook)

// put in webhooks route handler
eventHandler.receive({
  name: request.headers['X-GitHub-Event'],
  data: request.body,
  signature: request.headers['X-Hub-Signature']
}).catch(handleErrorsFromHooks)
```

We could call also just call the standalone module "handler" instead of "event-handler". But really we don’t "handle" webhooks but events sent by webhooks. I find it hard sometimes to distinguish between the terms "Webhook" and "Event", making "event" part of the name and having a good documentation in the READMEs makes things more clear, I think.

🤔 Naming is hard :tm: Thanks for bearing with me. I’ll sleep over this 💤

gr2m:
here is the PR for the changes in Probot: https://github.com/probot/probot/pull/335

README.md
```diff
@@ -1,3 +1,636 @@
...
+
+Returns a `requestListener` (or _middleware_) method which can be directly passed to [`http.createServer()`](https://nodejs.org/docs/latest/api/http.html#http_http_createserver_requestlistener), <a href="http://expressjs.com/">Express</a> and other compatible Node.js server frameworks.
+
+Can also be used [standalone](middleware/).
+
+### List of all event names
+
+See the full list of [event types with example payloads](https://developer.github.com/v3/activity/events/types/).
+
+If there are actions for a webhook, events are emitted for both, the webhook name as well as a combination of the webhook name and the action, e.g. `installation` and `installation.created`. A `*` event is always emitted. If an event handler throws an error an `error` event is triggered.
```
hiimbex: When I first started with probot listening on `*` events was a confusing concept for me, so I think maybe this wording could be expanded to explain listening on * listens to all events. Not sure if it should be included in the chart though?

hiimbex (approved):
This looks amazing!! Great work @gr2m! Left one comment about * events :)

README.md
gr2m: Thanks for bringing this up. I agree we should make this more clear. The difference between `*` and `error` event and all the others is that these two are implemented on the library level, all the others come from GitHub webhooks. I’ll try to make this more clear by adding a section "Special events", I’m open to suggestions :)

hiimbex: 👍 I think a "Special Events" category would cover it well, and allow you to explain in more detail how `*` and `error` work.

PR #9, opened 2017-11-27
Timing attach vulnerability

gr2m:


verify/index.js
```diff
@@ -9,8 +10,9 @@ function verify (secret, eventPayload, signature) {
...
   }
 
-  return Buffer.compare(
-    Buffer.from(signature),
-    Buffer.from(sign(secret, eventPayload))
-  ) === 0
+  const signatureBuffer = Buffer.from(signature)
+  const verificationBuffer = Buffer.from(sign(secret, eventPayload))
+
+  return signatureBuffer.length === verificationBuffer.length &&
```
bkeepers: This allows an attacker to quickly guess the length of the signature. I'm not sure if it's that big of a deal though since GitHub signatures are probably a known & consistent length. If this comparison could be avoided, I'd recommend it, but it's probably fine if it can't.

gr2m: it does the same thing in `buffer-equal-constant-time`: https://github.com/salesforce/buffer-equal-constant-time/blob/master/index.js#L18

bkeepers (approved):
👍  Thanks for the quick fix

PR #21, opened 2018-03-15
Unknown events: log warning instead of throwing error

### LINKED ISSUE #20: Asserting webhook names is not future-proof
https://github.com/octokit/webhooks.js/blob/c4f495f14a4f23d590c22b4aef2d0d464445ba78/event-handler/on.js#L12

Limiting the events that can be received to this list has two issues:

1. This library can't be used by beta testers for webhooks that are not yet unreleased
2. I think there are some undocumented webhooks for GitHub Enterprise. I can't find the list right now, but will try to locate it

It'd be great if the assertions could be turned off, or turned into just warnings.

gr2m:
### Breaking change

Before, this code would throw an error

```js
webhooks.on('woodstock', handler)
```

Now it logs a warning: `"woodstock" is not a known webhook name (https://developer.github.com/v3/activity/events/types/)`

---

closes #20

### DIFF
--- a/event-handler/on.js
+++ b/event-handler/on.js
@@ -9,7 +9,7 @@ function receiverOn (state, webhookNameOrNames, handler) {
   }
 
   if (webhookNames.indexOf(webhookNameOrNames) === -1) {
-    throw new Error(`${webhookNameOrNames} is not a valid webhook name`)
+    console.warn(`"${webhookNameOrNames}" is not a known webhook name (https://developer.github.com/v3/activity/events/types/)`)
   }
 
   if (!state.hooks[webhookNameOrNames]) {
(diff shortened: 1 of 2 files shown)


event-handler/on.js
```diff
@@ -9,7 +9,7 @@ function receiverOn (state, webhookNameOrNames, handler) {
   }
 
   if (webhookNames.indexOf(webhookNameOrNames) === -1) {
-    throw new Error(`${webhookNameOrNames} is not a valid webhook name`)
+    console.warn(`"${webhookNameOrNames}" is not a known webhook name (https://developer.github.com/v3/activity/events/types/)`)
```
gr2m, line 12: @bkeepers okay like that?

bkeepers (approved):
Perfect, thanks!

PR #22, opened 2018-03-31
handling request.body

dylancarlone:
This handles the case where request.body already exists. I was running into trouble using this package on Firebase, as it parses the body automatically, causing request.on('data') not to fire.

### DIFF
--- a/middleware/middleware.js
+++ b/middleware/middleware.js
@@ -6,7 +6,7 @@ const verifyAndReceive = require('./verify-and-receive')
 
 const debug = require('debug')('webhooks:receiver')
 function middleware (state, request, response, next) {
-  function handlePayload(payload) {
+  function handlePayload (payload) {
     verifyAndReceive(state, {
       id: id,
       name: eventName,
@@ -63,14 +63,14 @@ function middleware (state, request, response, next) {
       response.statusCode = 500
       response.end(error.toString())
     })
-  
+
     request.on('data', (chunk) => {
       dataChunks.push(chunk)
     })
-  
+
     request.on('end', () => {
       const payload = Buffer.concat(dataChunks).toString()
       handlePayload(JSON.parse(payload))
-    }) 
+    })
   }
 }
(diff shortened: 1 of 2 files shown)


gr2m:
Hey @dylancarlone, thanks for looking into solving the problem you run into, that’s great!

Could you please check the failing linting errors? Could you please add a failing test?

I’d also need a way to reproduce that myself. In what circumstance is the request body already set? Could you please create a repository with a minimal, reproducible test case?

dylancarlone:
Sure, I’ll put that together later this evening. Is there any reason not to
use the body-parser package to handle the request body instead of the data
event? I am using Firebase cloud functions which parses the body into
request.body automatically before any of my code runs. See
https://cloud.google.com/functions/docs/writing/http#parsing_http_requests

gr2m:
> Sure, I’ll put that together later this evening

Awesome, thanks!

> Is there any reason not to use the body-parser package to handle the request body instead of the data event

Good question :) I don’t see any reason not to use the body-parser right now, I’d like to investigate why we didn’t use it in the first place

> I am using Firebase cloud functions which parses the body into
request.body automatically before any of my code runs

Good to know, thanks! Could you please add a comment to the code so people wondering in future will know why we did what we did?

dylancarlone:
I fixed the linting errors, added a comment and added a test case where request.body is present. I'm also happy to submit a separate pull request adding a test that fails when request.body is present in the current code. I believe it will also fail when used as a middleware in an app that uses the body-parser package, so this problem should be bigger than just a Firebase implementation.

gr2m:
> I'm also happy to submit a separate pull request adding a test that fails when request.body is present in the current code

I’d appreciate it a lot, thank you. It’d be 💯 if you then pushed the commit with the failing test first so that we get the ❌ on the commit from CI, then push the fix. That’s a dream come true for reviewers :)

dylancarlone:
Sure, I'm going to refork and start from the beginning to avoid making a mess with commit order. So I may have to open a new pull request for the fix.

dylancarlone:
Closing in favor #23

PR #23, opened 2018-04-03
adding failing test for request.body

### LINKED ISSUE #40: webhook hangs when used in combination to bodyparser.json
Great project!

I'm using body-parser to transform requests to json form. Seems like this will consume the body stream before webhooks gets a chance to look at it.

Webhooks will then hang the request until the express timeout triggers and cancels the request. No error messages are thrown.

Perhaps a quick note in the README that this package must come before any body consumers such as body parser? Maybe an error thrown by checking if req.body already exists?

I can put together a PR as well.

dylancarlone:
This test shows that the middleware will hang when request.body is created by consuming the request data stream. 

closes #40

### DIFF
--- a/middleware/middleware.js
+++ b/middleware/middleware.js
@@ -2,6 +2,7 @@ module.exports = middleware
 
 const isntWebhook = require('./isnt-webhook')
 const getMissingHeaders = require('./get-missing-headers')
+const getPayload = require('./get-payload')
 const verifyAndReceive = require('./verify-and-receive')
 
 const debug = require('debug')('webhooks:receiver')
@@ -39,48 +40,23 @@ function middleware (state, request, response, next) {
 
   debug(`${eventName} event received (id: ${id})`)
 
-  const dataChunks = []
-  request.on('error', (error) => {
-    response.statusCode = 500
-    response.end(error.toString())
-  })
+  getPayload(request)
 
-  request.on('data', (chunk) => {
-    dataChunks.push(chunk)
-  })
-
-  request.on('end', () => {
-    const data = Buffer.concat(dataChunks).toString()
-    let payload
-
-    try {
-      payload = JSON.parse(data)
-    } catch (error) {
-      response.statusCode = 400
-      response.end('Invalid JSON')
-      return
-    }
-
-    verifyAndReceive(state, {
-      id: id,
-      name: eventName,
-      payload,
-      signature
+    .then((payload) => {
+      return verifyAndReceive(state, {
+        id: id,
+        name: eventName,
+        payload,
+        signature
+      })
     })
 
-      .then(() => {
-        response.end('ok\n')
-      })
+    .then(() => {
+      response.end('ok\n')
+    })
 
-      .catch(error => {
-        response.statusCode = error.status || 500
-        response.end(error.toString())
-      })
-  })
-
-  // refiring data and end event if request.body exists since they won't fire on their own
-  if (request.body) {
-    request.emit('data', Buffer.from(JSON.stringify(request.body)))
-    request.emit('end')
-  }
+    .catch(error => {
+      response.statusCode = error.status || 500
+      response.end(error.toString())
+    })
 }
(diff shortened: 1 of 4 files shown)


dylancarlone:
I was actually able to get this to work just fine by wrapping the middleware and firing events after the middleware has subscribed:

```
function (request, response) {
  webhooks.middleware(request, response);

  if (request.body) {
    request.emit('data', Buffer.from(JSON.stringify(request.body)));
    request.emit('end');
  }
}
```

This would be an easy fix, whether or not it's implemented inside the middleware itself.

gr2m:
sounds good to me 👍 sorry for the long response times, I’m mostly out until April 22nd

dylancarlone:
I added the fix and the tests are now passing. If you don't think this is the way to do it or would prefer to hold off, I'm happy just refiring the events in my code and leaving the package be.

BerkeleyTrue:
Well, just opened an issue for this. (#40)

ping @dylancarlone

dylancarlone:
@BerkeleyTrue I had success duplicating the logic I added to this PR in my own app:
https://github.com/octokit/webhooks.js/blob/41c0930737c635e926f8c360a1a531c5e6fd5a24/middleware/middleware.js#L70

My problem initially came from FireBase parsing the request body before any of my own code ran.

dylancarlone:
@BerkeleyTrue I had success duplicating the logic I added to this PR in my own app:
https://github.com/octokit/webhooks.js/blob/41c0930737c635e926f8c360a1a531c5e6fd5a24/middleware/middleware.js#L70

My problem initially came from FireBase parsing the request body before any of my own code ran.

nbransby:
@dylancarlone do you mean firebase functions? Im getting a timeout in firebase functions on calling verifyAndReceive, is this related?

gr2m:
I’ve rebased the PR, so this is ready to merge. I can’t think of any disadvantages of merging this. Does anyone have any objections or concerns?

nbransby:
Not from me.. I'm waiting for it!

middleware/middleware.js
```diff
@@ -77,4 +77,10 @@ function middleware (state, request, response, next) {
         response.end(error.toString())
       })
   })
+
+  // refiring data and end event if request.body exists since they won't fire on their own
+  if (request.body) {
+    request.emit('data', Buffer.from(JSON.stringify(request.body)))
+    request.emit('end')
+  }
```
gr2m: I wonder, is there a need to emit the events, why not just call `verifyAndReceive` directly? 

```js
  if (request.body) {
    verifyAndReceive(state, {
      id: id,
      name: eventName,
      payload: request.body,
      signature
    })

      .then(() => {
        response.end('ok\n')
      })

      .catch(error => {
        response.statusCode = error.status || 500
        response.end(error.toString())
      })
  }
```

gr2m: See my change at https://github.com/octokit/webhooks.js/pull/23/commits/b380f87c53e51cc9164f309e76becca4bd5cf16b which omits re-emitting the "data" event on the `request` object. Test is still passing, so unless anyone objects I’ll go ahead with that :)

gr2m:
@dylancarlone @nbransby  I’ve updated the implementation. Could you check if it works with Firebase functions now?

nbransby:
Works for me

nbransby:
Sorry but it would appear its still hanging on master for me my firebase function is defined as:

```js
export const webhook = https.onRequest((req, resp) => webhooks.verifyAndReceive({
    id: req.headers['x-github-delivery'],
    name: req.headers['x-github-event'],
    payload: req.body,
    signature: req.headers['x-hub-signature']
}));
```

results in

`Function execution took 60002 ms, finished with status: 'timeout'`

gr2m:
This seems unrelated to this pull request. You use the `. verifyAndReceive` method directly, while this PR was about the http middleware.

I don’t know firebase functions, but from these docs: https://firebase.google.com/docs/functions/http-events it looks like you still have to send a response, in your code example you don’t do anything with the `resp` parameter.

Maybe try this instead:

```js
export const webhook = https.onRequest(webhooks.middleware);
```

PR #34, opened 2018-08-02
Trigger error handler when missing headers

### LINKED ISSUE #29: Error handler is not called in case of missing headers
Hi there 👋 
I am opening this regarding some findings from the issue probot/probot#568

While debugging to understand the error, I noticed that if the header `X-Hub-Signature` is missing (or any header actually) a 400 error response is sent, but the error is never propagated to the error handler (in that case, probot):
https://github.com/octokit/webhooks.js/blob/17ec6e3dc434d39caf542efba017509339be8a52/middleware/middleware.js#L25-L32

Sorry if I got it wrong, but shouldn't it be calling the error handler like the case with an invalid signature?
https://github.com/octokit/webhooks.js/blob/17ec6e3dc434d39caf542efba017509339be8a52/middleware/verify-and-receive.js#L8-L15

bkeepers:
This fixes #29 by triggering an error when headers are missing. 

cc @hiimbex @mwebler

### DIFF
--- a/middleware/middleware.js
+++ b/middleware/middleware.js
@@ -24,11 +24,13 @@ function middleware (state, request, response, next) {
 
   const missingHeaders = getMissingHeaders(request).join(', ')
   if (missingHeaders) {
-    debug(`ignored: ${request.method} ${request.url} due to missing headers: ${missingHeaders}`)
+    const error = new Error(`Required headers missing: ${missingHeaders}`)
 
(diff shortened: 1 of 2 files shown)


hiimbex (approved):
LGTM

mwebler:
Looks good 👍  Thanks Brandon!

gr2m (approved):

PR #41, opened 2018-11-26
Fix header name in readme

nbransby:
Maybe its because im using GitHub Apps but this header name is wrong

-----
[View rendered README.md](https://github.com/nbransby/webhooks.js/blob/master/README.md)


gr2m (approved):
You are right, thank you for sending the PR!

PR #50, opened 2018-12-11
remove obsolete and invalid events & actions

gr2m:


lib/webhook-names.json
```diff
@@ -101,15 +96,12 @@
   "pull_request.assigned",
   "pull_request.closed",
   "pull_request.edited",
-  "pull_request.false",
   "pull_request.labeled",
-  "pull_request.merged",
```
gr2m, line 110: I double checked that this action really doesn’t exist. But in fact when merging a PR the action is "closed"

PR #53, opened 2018-12-20
Type definitions

gimenete:
This adds type definitions for TypeScript.

To do:

- [x] I think it's a good idea to track `index.d.ts` in git and generate it before publishing the package. I've tracked the file but left to do is running the script to generate it on `prepublish`
- [x] There are a few things commented out in the type definition file
- [x] Test this with probot. PR here https://github.com/probot/probot/pull/832

Screenshots:

screen shot 2018-12-20 at 12 33 07
screen shot 2018-12-20 at 12 33 25
screen shot 2018-12-20 at 12 33 49
screen shot 2018-12-20 at 12 34 01


gimenete:
@julien-c I've just made a commit to define the payload types first and then define the method signatures so you can use the types outside those methods signatures. Example of payload definition: https://github.com/octokit/webhooks.js/pull/53/files#diff-b52768974e6bc0faccb7d4b75b162c99R8

Nevertheless maybe you can take a look to the `generate-types.js` file to see how types are generated and do something similar. Basically it reads the `examples` from the JSON file and uses `@gimenete/type-writer` to generate typescript definitions from those examples.

generate-types.js
```diff
@@ -0,0 +1,60 @@
+const TypeWriter = require('@gimenete/type-writer')
+const webhooks = require('@octokit/webhooks-definitions/index.json')
```
gr2m:
```suggestion
const webhooks = require('@octokit/webhooks-definitions')
```

That will allow to move the `index.json` file in the future. As long as the `main` property in `@octokit/webhooks-definitions`’s package.json file points to the new file it will continue to work

gr2m:
> I think it's a good idea to track index.d.ts in git and generate it before publishing the package. I've tracked the file but left to do is running the script to generate it on prepublish

We could run the script automatically when `@octokit/webhooks` is updated:
https://github.com/octokit/webhooks.js/blob/cd16d7dbf38dca77d24b7f00739dc6e54e39853b/.travis.yml#L35

I don’t mind tracking the `index.d.ts` file but please add a very big comment on top making clear that this file is generated, otherwise folks will start sending pull requests against it :)

generate-types.js
```diff
@@ -0,0 +1,60 @@
...
+    ...actions.map(action => `'${name}.${action}'`)
+  ].join(' | ')
+  types.push(`type ${typeName} = ${tw.generate('typescript', { inlined: true })}`)
+  signatures.push(`
+    public on (event: ${events}, callback: (event: Webhooks.WebhookEvent<${typeName}>) => void)
+  `)
+}
+
+const definition = `
+// import { Application } from 'express'
```
gr2m: We don’t use express, is this obsolete?

gimenete: That was copied from https://github.com/probot/probot/blob/master/src/%40types/%40octokit/webhooks/index.d.ts

I commented it out because maybe probot needs it. But now I thinking that if needed probot can just extend the type. So I'll delete that.

generate-types.js
```diff
@@ -0,0 +1,60 @@
...
+  path: string,
+  secret: string
+}
+
+${types.join('\n\n')}
+
+declare class Webhooks {
+  // public middleware: Application
+
+  constructor (Options)
```
wolfy1339: Ref: probot/probot#798
```suggestion
  constructor (options: Options)
```

wolfy1339 (changes requested):

generate-types.js
```diff
@@ -0,0 +1,60 @@
...
+  const tw = new TypeWriter()
+  const typeName = `WebhookPayload${pascalCase(name)}`
+  tw.add(examples, { rootTypeName: typeName })
+  const events = [
+    `'${name}'`,
+    ...actions.map(action => `'${name}.${action}'`)
+  ].join(' | ')
+  types.push(`type ${typeName} = ${tw.generate('typescript', { inlined: true })}`)
+  signatures.push(`
+    public on (event: ${events}, callback: (event: Webhooks.WebhookEvent<${typeName}>) => void)
```
wolfy1339:
```suggestion
    public on (event: ${events}, callback: (event: Webhooks.WebhookEvent<${typeName}>) => void): void
```

generate-types.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+${types.join('\n\n')}
+
+declare class Webhooks {
+  // public middleware: Application
+
+  constructor (Options)
+
+  public on (event: 'error', callback: (event: Error) => void)
```
wolfy1339:
```suggestion
  public on (event: 'error', callback: (event: Error) => void): void
```

generate-types.js
```diff
@@ -0,0 +1,60 @@
...
+  public on (event: '*' | string[], callback: (event: Webhooks.WebhookEvent<any> | Error) => void)
```
wolfy1339:
```suggestion
  public on (event: '*' | string[], callback: (event: Webhooks.WebhookEvent<any> | Error) => void): void
```

wolfy1339 (changes requested):

wolfy1339:
I've been looking around at the examples on GitHub's docs, and some examples come with an example that doesn't include everything.

For example, in the `check_run` event's payload the `check_suite.pull_requests` property is incomplete, and thus would lead to incomplete types here.

gimenete:
@wolfy1339 that's not great... Nevertheless you can always cast to `any` and it will be better than no having type definitions at all, I think.

@gr2m other than that, I think this is ready. I have also a local branch of probot that works with these changes.

generate-types.js
```diff
@@ -0,0 +1,53 @@
...
+  constructor (Options)
+
+  public on (event: 'error', callback: (event: Error) => void): void
+  public on (event: '*' | string[], callback: (event: Webhooks.WebhookEvent<any> | Error) => void): void
+  ${signatures.join('\n')}
+
+  public sign (data: any): string
+}
+
+export interface WebhookEvent<T> {
```
wolfy1339: Please move this before the definition of the class. It's always best practice to declare things before you use them.

wolfy1339: Also, this needs to be moved into a namespace
```suggestion
export namespace Webhooks {
    export interface WebhookEvent<T> {
```

wolfy1339 (changes requested):

gr2m:
@gimenete 

> Test this with probot, removing the need for this file
> https://github.com/probot/probot/blob/master/src/%40types/%40octokit/webhooks/index.d.ts

Could you check @wolfy1339’s feedback and update / remove the file above? It’d be great to make this update as part of the Probot v8 beta release: https://github.com/probot/probot/pull/826

scripts/generate-types.js
```diff
@@ -0,0 +1,65 @@
...
+${tw.generate('typescript', { inlined: false }).replace(/type /g, 'export type ')}
+
+export namespace Webhooks {
+  export interface WebhookEvent<T> {
+    id: string
+    name: string
+    payload: T
+    protocol?: 'http' | 'https'
+    host?: string
+    url?: string
```
gr2m, line 50: @gimenete where does `protocol`, `host`, `url` come from?

gimenete: I don’t know. I copied that from the definitions probot had

PR #55, opened 2019-01-26
fix(typescript): Add missing class functions

wolfy1339:


wolfy1339:
This should be ready to go @gr2m

scripts/generate-types.js
```diff
@@ -54,9 +56,16 @@ export namespace Webhooks {
 
     public on (event: 'error', callback: (event: Error) => void): void
     public on (event: '*' | string[], callback: (event: Webhooks.WebhookEvent<any>) => void): void
+    public on (event: '*' | string[], callback: (event: Webhooks.WebhookEvent<any>) => Promise<void>): void
     ${signatures.join('\n')}
 
     public sign (data: any): string
+    public verify (eventPayload: WebhookPayloadWithRepository, signature: string): boolean
```
gr2m: Sorry, I’m not sure what `WebhookPayloadWithRepository` is? Not all event payload has a repository key, e.g. https://developer.github.com/v3/activity/events/types/#installationevent has `repositories`.

But if that works, can we use the same type for `sign()`? Both `sign()` and `verify()` accept the same payload type as first argument

scripts/generate-types.js
```diff
@@ -54,9 +56,16 @@ export namespace Webhooks {
...
+    public verifyAndReceive (options: { id: string, name: string, payload: any, signature: string }): Promise<void>
+    public receive (options: { id: string, name: string, payload: any }): Promise<void>
+    public removeListener (event: string, callback: (event: Webhooks.WebhookEvent<any>) => void): void
+    public removeListener (event: string, callback: (event: Webhooks.WebhookEvent<any>) =>Promise<void>): void
```
gr2m: `event` can be a `string` or an event of strings. In line 66, I think the return type is incorrect/

```diff
- (event: Webhooks.WebhookEvent<any>) => void)
+ (event: Webhooks.WebhookEvent<any>) => Promise<void>)
```

gr2m (commented):
Thanks for working on this!

scripts/generate-types.js
wolfy1339: I had taken these from a PR of mine from Probot. I'll fix that. I didn't notice it

gr2m (approved):
Awesome, thank you!

PR #56, opened 2019-01-26
Typescript: export Webhooks explicitly

gr2m:
This is how we do it in @octokit/rest. I don’t know if that’s the right way to do it, but this change was necessary in order to get https://github.com/probot/probot/pull/832 to work


wolfy1339:
You shouldn't need to export the namespace and the class anymore as they're exported via the CommonJS export in TS

PR #58, opened 2019-01-27
fix(typescript): Fix errors from linters and an update for signatures

wolfy1339:
You can't use ESModule and CommonJS exports in the same file (you could theoretically, but it's bad practice), so we have to move the definitions into the namespace.

This does not affect the way to import the types, you can still do this:
```ts
import { WebhookPayloadStatus } from '@octokit/webhooks';
```

In the `Options` type, replace the `T` parameter with `any` since it was not defined.

### DIFF
--- a/scripts/generate-types.js
+++ b/scripts/generate-types.js
@@ -40,7 +40,7 @@ type Options = {
   transform?: (event: Webhooks.WebhookEvent<any>) => Webhooks.WebhookEvent<any> & { [key: string]: any }
 }
 
-namespace Webhooks {
+declare namespace Webhooks {
   ${tw.generate('typescript', { inlined: false })}
 
   interface WebhookEvent<T> {
@@ -53,8 +53,8 @@ namespace Webhooks {
   }
 }
 
-class Webhooks {
-  constructor (options: Options)
+declare class Webhooks {
+  constructor (options?: Options)
 
   public on (event: 'error', callback: (event: Error) => void): void
   public on (event: '*' | string[], callback: (event: Webhooks.WebhookEvent<any>) => Promise<void> | void): void
--- a/scripts/generate-types.js
+++ b/scripts/generate-types.js
@@ -34,7 +34,7 @@ const definition = `
 
 import http = require('http')
 
-export type Options = {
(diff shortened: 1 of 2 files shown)


scripts/generate-types.js
```diff
@@ -24,7 +24,8 @@ webhooks.forEach(({ name, actions, examples }) => {
     ...actions.map(action => `'${name}.${action}'`)
   ].join(' | ')
   signatures.push(`
-    public on (event: ${events}, callback: (event: Webhooks.WebhookEvent<${typeName}>) => void): void
+    public on (event: ${events}, callback: (event: Webhooks.WebhookEvent<Webhhooks.${typeName}>) => void): void
```
gr2m:
```suggestion
    public on (event: ${events}, callback: (event: Webhooks.WebhookEvent<Webhooks.${typeName}>) => void): void
```

gr2m (commented):
could you run `npm run generate-types` to also update the index.d.ts file?

scripts/generate-types.js
```diff
@@ -24,7 +24,8 @@ webhooks.forEach(({ name, actions, examples }) => {
     ...actions.map(action => `'${name}.${action}'`)
   ].join(' | ')
   signatures.push(`
-    public on (event: ${events}, callback: (event: Webhooks.WebhookEvent<${typeName}>) => void): void
+    public on (event: ${events}, callback: (event: Webhooks.WebhookEvent<Webhooks.${typeName}>) => void): void
+    public on (event: ${events}, callback: (event: Webhooks.WebhookEvent<Webhooks.${typeName}>) => Promise<void>): void
```
gr2m: I wonder, can you combine the two lines above?

```ts
public on (event: ${events}, callback: (event: Webhooks.WebhookEvent<Webhooks.${typeName}>) => void | Promise<void>): void
```

wolfy1339: I'm not sure if It could work, I'll go test

wolfy1339:
@gr2m This is ready

gr2m (approved):
Thanks for the ping!

PR #66, opened 2019-03-29
fix(typescript): Update signature for middleware function

### LINKED ISSUE #64: Readme Example Fails to Compile with TypeScript
Hello! I attempted to build a new TypeScript project that consumes this module. I created a `package.json` and installed the following dependencies:
```json
{
  "dependencies": {
    "@octokit/webhooks": "5.3.3"
  },
  "devDependencies": {
    "@types/node": "^11.12.0",
    "typescript": "^3.3.4000"
  }
}
```

I took the example from the readme and fixed the imports to work with TypeScript but it fails to compile:
```ts
// TypeScript default import syntax
import * as WebhooksApi from '@octokit/webhooks';
import * as http from 'http';

const webhooks = new WebhooksApi({
  secret: 'mysecret'
});

webhooks.on('*', ({id, name, payload}) => {
  console.log(name, 'event received');
});

http.createServer(webhooks.middleware).listen(3000);
//                ~~~~~~~~~~~~~~~~~~~
// TS2559: Type '(request: ClientRequest, response: ServerResponse, next: (err?: any) => void) => (request: IncomingMessage, response: ServerResponse) => void' has no properties in common with type 'ServerOptions'.
```

I would expect that this code example would successfully compile. What is going on here? It looks like `middleware` is not a `http.RequestListener`, but a function that returns a `http.RequestListener`. Is this merely an issue with the definition of `middleware` in `index.d.ts`? 

Thanks for your help!

wolfy1339:
The bug was introduced in #55
Fixes #64

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3876,10 +3876,10 @@ declare class Webhooks {
     callback: (event: Webhooks.WebhookEvent<any>) => Promise<void>
   ): void;
   public middleware(
-    request: http.ClientRequest,
+    request: http.IncomingMessage,
     response: http.ServerResponse,
     next: (err?: any) => void
-  ): (request: http.IncomingMessage, response: http.ServerResponse) => void;
+  ): void | Promise<void>;
 }
 
 export = Webhooks;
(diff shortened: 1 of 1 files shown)


gr2m (approved):
Thanks!

PR #68, opened 2019-04-01
fix(typescript): Properly fix #64

### LINKED ISSUE #64: Readme Example Fails to Compile with TypeScript

wolfy1339:
I had totally forgot about it. I'm used to having a template file lie other repositories xD

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3926,10 +3926,10 @@ declare class Webhooks {
     callback: (event: Webhooks.WebhookEvent<any>) => Promise<void>
   ): void;
   public middleware(
-    request: http.ClientRequest,
+    request: http.IncomingMessage,
     response: http.ServerResponse,
     next: (err?: any) => void
-  ): (request: http.IncomingMessage, response: http.ServerResponse) => void;
+  ): void | Promise<void>;
 }
 
 export = Webhooks;
--- a/index.d.ts
+++ b/index.d.ts
@@ -3928,7 +3928,7 @@ declare class Webhooks {
   public middleware(
     request: http.IncomingMessage,
     response: http.ServerResponse,
-    next: (err?: any) => void
+    next?: (err?: any) => void
   ): void | Promise<void>;
 }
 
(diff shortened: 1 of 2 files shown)


gr2m:
> I'm used to having a template file lie other repositories

what do you mean?

gr2m (approved):
Looks good to me 👍

gr2m:
you want to wait for https://github.com/octokit/webhooks.js/pull/69 or shall I merge this in?

wolfy1339:
Better to merge this now

wolfy1339:
> 
> 
> > I'm used to having a template file lie other repositories
> 
> what do you mean?

Like this file in the `@octokit/rest` repo: https://github.com/octokit/rest.js/blob/master/scripts/templates/index.d.ts.tpl

gr2m:
There is [scripts/generate-types.js](https://github.com/octokit/webhooks.js/blob/master/scripts/generate-types.js) and a warning on top of the file
https://github.com/octokit/webhooks.js/blob/245c40d3f0c218a5a18db905415fa923c5a2a7f7/index.d.ts#L1-L2

:)

gr2m:
I think we had the idea to split the `index.d.ts` into a static part we can edit and a generated one that is clearly declared as such, e.g. by living in a path such as `/generated/definitions.d.ts`, then import `generated/definitions` in index.d.ts?

PR #72, opened 2019-04-11
normalize escape sequences after JSON.stringify()

### LINKED ISSUE #71: verifyAndReceive fails depending on json details
# Summary
`verifyAndReceive` uses a parsed representation of the passed json data. This can lead to false negative signature verifications. One Example is the signature verification failure for ANSI/ISO-6429 codes a.k.a. escape sequences. If Any GitHub event body contains an escape sequence, the verification will always fail, even though everything is valid.

## Details
When the `ESC` sequence is json-encoded it might be encoded as `\u001b` OR `\u001B`. 
The node json parser will always encode this sequence with a lowercase `b`, which means that the verification will fail if the signature was generated with a capital `B` (which is already the case).
Since this (and other encoding details) are ambiguous in the json standard, the only safe option is to verify the signature with the raw body we receive and not with the parsed representation, since `JSON.stringify(JSON.parse(x))` does not necessarily yield `x`, thus causing a false negative. 
This however is only one example, that already happens. In general, the only way that will guarantee that we properly verify the signature header is, using the data in the format it was encoded in the first place. This means, before parsing it.

# Suggestion
In order to solve the problem, [middleware/middleware](https://github.com/octokit/webhooks.js/blob/master/middleware/middleware.js) might need to be adjusted, together with `get-payload` and `verify-and-receive`.
I was already working on that when a realised the following compatibility issues:

# Compatibility concerns

1. `verifyAndReceived` is part of the public API, meaning people are using it directly and outside of the middleware context. (I saw it in some PR discussions). This means this function has to stay and still work with parsed data instead of json.
1. Since [this PR](https://github.com/octokit/webhooks.js/pull/23) octokit is supporting the usage in combination with [body-parser](https://www.npmjs.com/package/body-parser), which will also store parsed information in `request.body`

It seems as if we need to keep the verification based on parsed data, if we want to support the mix of `body-parser` and webhook.js. We could also store the raw body in the request object.

A compromise could be:
testing if `req.body` is defined and if that is the case, we could use the current implementation.
Also keep the current implementation of `verifyAndReceive` exposed and (maybe?) deprecate it. 
In order to fix our bug, we could just add a second verify and receive implementation that does the verification with the raw data. 
This however would duplicate a lot of code and I am not sure what you think about this.

I hope we can figure out a way to fix this bug. This is just my view on things, maybe somebody has a better suggestion on how to address this issue.

# Related issues
* This issue might be related to: #61
* There are also several slack discussions raising this topic (Non ANSI code support)

/cc @gr2m

gr2m:
quick-fix for #71

- [x] verify the fix works as expected with a production setup

I confirmed this fix works using this server code on glitch.com

```js
const Middleware = require('@octokit/webhooks/middleware')
const middleware = new Middleware({
  secret: 'development',
  path: '/'
})

middleware.on('*', console.log)
middleware.on('error', console.log)

require('http').createServer(middleware).listen(process.env.PORT)
```

And this curl command

```sh
curl -i   -H "Authorization: token [REDACTED]"   -H "Content-Type: application/vnd.github.v3+json"   -d '{"body":"Foo\n\u001B[34mbar: \u2665\u2665\u2665\u2665\u2665\u2665\u2665\u2665\nthis-is-lost\u001b[0m\u001B[2K" }'   https://api.github.com/repos/gr2m/sandbox/issues/78/comments
```

On `@octokit/rest@5.02` the server response with

> signature does not match event payload and secret

Installing from `octokit/webhooks#71/verify-json-error` resolves the error

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,10204 @@
+{
+  "name": "@octokit/webhooks",
+  "version": "0.0.0-development",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "^7.0.0"
+      }
+    },
+    "@babel/generator": {
+      "version": "7.4.0",
+      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.4.0.tgz",
+      "integrity": "sha512-/v5I+a1jhGSKLgZDcmAUZ4K/VePi43eRkUs3yePW1HB1iANOD5tqJXwGSG4BZhSksP8J9ejSlwGeTiiOFZOrXQ==",
+      "dev": true,
+      "requires": {
+        "@babel/types": "^7.4.0",
+        "jsesc": "^2.5.1",
+        "lodash": "^4.17.11",
+        "source-map": "^0.5.0",
+        "trim-right": "^1.0.1"
+      }
+    },
+    "@babel/helper-function-name": {
+      "version": "7.1.0",
+      "resolved": "https://registry.npmjs.org/@babel/helper-function-name/-/helper-function-name-7.1.0.tgz",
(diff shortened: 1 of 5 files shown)


dennissivia (approved):
LGTM 👍 
I also like the simplification of the safe crypto comparison.
I think all follow up long term changes or ideas can be discussed elsewhere, so I am happy 👏

dennissivia (approved):
LGTM 👏

gr2m:
@scepticulous thanks for the review!

@gimenete could you have a look?

tcbyrd:
@gr2m Any opinions on how best to backport this into Probot 7.x? Some of our integrations haven't made the jump yet to Probot 9.x and will need this fix.

gimenete (approved):
Looks good to me! 👍 🎉

gr2m:
@tcbyrd could you make a separate issue so I don’t forget, I’m afk right now

PR #85, opened 2019-06-16
Correct missing external reference to GitHub docs

azerella:
The `README.md` is missing the correct URL to the GitHub docs

-----
[View rendered README.md](https://github.com/adamzerella/webhooks.js/blob/master/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,7 +351,7 @@ webhooks.on(eventNames, handler)
     </td>
     <td>
       <strong>Required.</strong>
-      Name of the event. One of <a target="_blank" rel="noopener noreferrer" href="https://developer.github.com/webhooks/#events">GitHub’s supported event names</a>.
+      Name of the event. One of [GitHub’s supported event names](https://developer.github.com/webhooks/#events).
     </td>
   </tr>
   <tr>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -351,7 +351,7 @@ webhooks.on(eventNames, handler)
     </td>
     <td>
       <strong>Required.</strong>
-      Name of the event. One of <a href="#listofalleventnames">GitHub’s supported event names</a>.
+      Name of the event. One of <a target="_blank" rel="noopener noreferrer" href="https://developer.github.com/webhooks/#events">GitHub’s supported event names</a>.
```
gr2m: Can you leave out the `target` and `rel` properties? GitHub doesn’t add these either when it transpiles Markdown links to HTML

wolfy1339: Might as well use Markdown syntax here instead

azerella: No worries, fixed 👍

README.md
```diff
@@ -351,7 +351,7 @@ webhooks.on(eventNames, handler)
     </td>
     <td>
       <strong>Required.</strong>
-      Name of the event. One of <a href="#listofalleventnames">GitHub’s supported event names</a>.
+      Name of the event. One of [GitHub’s supported event names](https://developer.github.com/webhooks/#events).
```
gr2m: Markdown won’t render correctly

image

I think it’s because of the indentation. If you remove the indentation and add empty lines above and below, then it works. But I think it remains more readable when you stick with the `<a>` tags

azerella: @gr2m I've reverted it back to anchor tags, should be good now.

I still question why we can't use `target="_blank"` and `rel="noopener noreferrer"` - These are very common and a recommended [security practice](https://developers.google.com/web/tools/lighthouse/audits/noopener)

gr2m: The docs say

> In general, when you use `target="_blank"`, always add `rel="noopener"` or `rel="noreferrer"`.

So if we don’t add `target="_blank"` we should be good?

gr2m (approved):
Thanks 👍

PR #89, opened 2019-07-18
Add changes to PR

just-be-dev:
This updates `WebhookPayloadPullRequest` to have a changes field. It converts both the PR payload and the issue payload `changes` field to be optional. They're only present on an edited webhook. Isn't perfect, but it's better than not having it.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1856,6 +1856,7 @@ declare namespace Webhooks {
     pull_request: WebhookPayloadPullRequestPullRequest;
     repository: PayloadRepository;
     sender: WebhookPayloadPullRequestSender;
+    changes?: WebhookPayloadIssuesChanges;
   };
   type WebhookPayloadPublicSender = {
     login: string;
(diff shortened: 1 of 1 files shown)


gr2m:
Hi @zephraph, the `index.d.ts` file is generated in https://github.com/octokit/webhooks.js/blob/master/scripts/generate-types.js, we don’t edit it manually. All changes would get overwritten the next time https://developer.github.com/v3/activity/events/types/ is changed.

Usually the right approach would be to ask the docs team to update the example payloads in https://developer.github.com/v3/activity/events/types/. But it this case it’s more complicated 🤔 

I wonder if we there would be a way where we could make a merge between what’s generated & updated automatically by parsing https://developer.github.com/v3/activity/events/types/#pullrequestevent and an additional file where we can maintain manual overrides until we maybe some day we get official JSON Schemas from GitHub 🙏 

Would you be interested in looking into that?

just-be-dev:
Ahh, that make sense... I don't know that I have the time currently, but let's create an issue for it so we keep track.

PR #95, opened 2019-09-30
Fix removeListener Iteration

aidant:
The current implementation of [EventHandler#removeListener](https://github.com/octokit/webhooks.js/blob/27bdc45d150be445c36cba19ab686cd9a4cad0a6/event-handler/remove-listener.js#L17) never iterates to index `0` in the state.

---

Example code to reproduce this issue:

```js
const Octokit = require('@octokit/rest')
const Webhook = require('@octokit/webhooks')
const http = require('http')

const GITHUB_ORGANIZATION_NAME = ''
const GITHUB_ORGANIZATION_WEBHOOK_ID = 0
const GITHUB_ORGANIZATION_WEBHOOK_PATH = '/'
const GITHUB_ORGANIZATION_WEBHOOK_PORT= 8080
const GITHUB_ORGANIZATION_WEBHOOK_SECRET = ''
const GITHUB_PERSONAL_ACCESS_TOKEN = ''

const octokit = new Octokit({
  auth: GITHUB_PERSONAL_ACCESS_TOKEN
})

const webhook = new Webhook({
  path: GITHUB_ORGANIZATION_WEBHOOK_PATH,
  secret: GITHUB_ORGANIZATION_WEBHOOK_SECRET
})

const entrypoint = async () => {
  const listener = (event) => {
    webhook.removeListener('*', listener)
    console.log(`Received event: "${event.name}".`)
  }

  webhook.on('*', listener)

  await octokit.orgs.pingHook({
    org: GITHUB_ORGANIZATION_NAME,
    hook_id: GITHUB_ORGANIZATION_WEBHOOK_ID,
  })

  await octokit.orgs.pingHook({
    org: GITHUB_ORGANIZATION_NAME,
    hook_id: GITHUB_ORGANIZATION_WEBHOOK_ID,
  })
}

http.createServer(webhook.middleware)
  .listen(GITHUB_ORGANIZATION_WEBHOOK_PORT, () => entrypoint().catch(console.error))
```

Without the fix:
```
Received event: "ping".
Received event: "ping".
```
With the fix:
```
Received event: "ping".
```

### DIFF
--- a/event-handler/remove-listener.js
+++ b/event-handler/remove-listener.js
@@ -14,7 +14,7 @@ function receiverListener (state, webhookNameOrNames, handler) {
 
   // remove last hook that has been added, that way
   // it behaves the same as removeListener
-  for (let i = state.hooks[webhookNameOrNames].length; i > 0; i--) {
+  for (let i = state.hooks[webhookNameOrNames].length - 1; i >= 0; i--) {
     if (state.hooks[webhookNameOrNames][i] === handler) {
       state.hooks[webhookNameOrNames].splice(i, 1)
       return
--- a/test/unit/event-handler-remove-listener-test.js
+++ b/test/unit/event-handler-remove-listener-test.js
@@ -1,6 +1,6 @@
 const test = require('tap').test
 
-const receiverListener = require('../../event-handler/remove-listener')
+const removeListener = require('../../event-handler/remove-listener')
 
 test('remove-listener: single listener', t => {
   const push = () => {}
@@ -12,7 +12,7 @@ test('remove-listener: single listener', t => {
   }
 
   t.doesNotThrow(() => {
-    receiverListener(state, 'push', push)
+    removeListener(state, 'push', push)
   })
   t.deepEqual(state, { hooks: { push: [] } })
   t.end()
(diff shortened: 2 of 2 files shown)


event-handler/remove-listener.js
```diff
@@ -14,7 +14,7 @@ function receiverListener (state, webhookNameOrNames, handler) {
 
   // remove last hook that has been added, that way
   // it behaves the same as removeListener
-  for (let i = state.hooks[webhookNameOrNames].length; i > 0; i--) {
+  for (let i = state.hooks[webhookNameOrNames].length - 1; i >= 0; i--) {
```
gr2m, line 17: Wow, good catch, thanks so much for investigating and sending a pull request, with tests 😻

test/unit/event-handler-remove-listener-test.js
```diff
@@ -0,0 +1,42 @@
+const test = require('tap').test
+
+const receiverListener = require('../../event-handler/remove-listener')
```
gr2m: have you meant to call this variable `removeListener`? I think that would make the tests a little more clear!

gr2m (commented):
Wonderful pull request, thank you so much! I just have one little nitpick question

test/unit/event-handler-remove-listener-test.js
aidant: I agree it does look cleaner to call it `removeListener`. I've updated the test with the new name.

gr2m (approved):
Perfect, thank you!

PR #96, opened 2019-10-07
Do not uppercase quoted escape sequences

hugopeixoto:
If the payload contains text that starts with "\u", the replacer kicks
in and signature validation fails.

Attempt at improving the fix to #71


gr2m (approved):
Great pull request, thanks a bunch 💐

gr2m:
does that fix #71 for you?

hugopeixoto:
> does that fix #71 for you?

It does! thank you for merging this so fast.

PR #113, opened 2020-02-17
feat(TypeScript): Webhook events are now an enum instead of a list of strings. Webhook event name types can be imported as EventNames

### LINKED ISSUE #112: [question/proposal] use of enums for event names?
[❓question] Would have sense to use [`enums`](https://www.typescriptlang.org/docs/handbook/enums.html) for event names? 
https://github.com/octokit/webhooks.js/blob/5ca6c8aba911e5888add0b1fa2d96a359596368b/index.d.ts#L4402-L4417

A use case where it would be valuable is to make sure the events to be listened `webhooks.on([<eventName>])` are the right ones on [compile time](http://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time).

If you think is a valid proposal I would be more than happy to propose a pull-request for it! :)

oscard0m:
## 📝 Proposal

- To move event names to a collection of `enum` so they can be reused and centralized
- To place them in a different file so they are easy to re-import from another project and `index.d.ts` becomes more readable

Fixes #112 

## ❓/ 💬 Questions / Discussion

Before editing `scritps/generate-types.js` I would like to agree which should be the output. 
What do you think about the proposed changes?
* Independent file for enums
* Name convention
* Any other feedback or suggestion?

### Features

- **TypeScript**: Webhook events are now an enum instead of a list of strings. Webhook event name types can be imported as `EventNames`

### TypeScript changes

- `Webhooks.WebhookEvent` type is now just `WebhookEvent`
- `Webhooks.WebhookPayload*` types are now `EventPayloads.WebhookPayload*`


gr2m:
Thanks for getting the PR going 👍 


> Independent file for enums

Good idea 👍 

> Name convention

I use dashes, not camelCase, for file names.

I like putting generated files into a `generated/` folder. Example: https://github.com/octokit/plugin-rest-endpoint-methods.js/tree/master/src/generated

Maybe we could leave the static parts of `index.d.ts` in there, and move all the generated parts into a `lib/generated/` folder? E.g. `lib/generated/event-names.d.ts`, `lib/generated/event-payloads.d.ts`, `lib/generated/api.d.ts`. Or combine all three into one if that's easier.

> Any other feedback or suggestion?

This repo is actually the last repository still written in JavaScript, I'd like to move it to TypeScript, I'd be happy to collaborate on that if you are interested.

oscard0m:
> I use dashes, not camelCase, for file names.

Sure, let's maintain the convention
 
> I like putting generated files into a `generated/` folder. Example: https://github.com/octokit/plugin-rest-endpoint-methods.js/tree/master/src/generated
> 
> Maybe we could leave the static parts of `index.d.ts` in there, and move all the generated parts into a `lib/generated/` folder? E.g. `lib/generated/event-names.d.ts`, `lib/generated/event-payloads.d.ts`, `lib/generated/api.d.ts`. Or combine all three into one if that's easier.

I like the approach, let me commit on this draft PR this `/generated/` folder proposal and from there I can start the development.
 
> This repo is actually the last repository still written in JavaScript, I'd like to move it to TypeScript, I'd be happy to collaborate on that if you are interested.

I would be happy to contribute on the migration. I'm starting to learn TypeScript so yes! Let's go for it 💪

lib/generated/api.d.ts
```diff
@@ -0,0 +1,371 @@
+import http = require("http");
```
oscard0m: by `api.d.ts` you meant all the different methods I placed in  this file right? I was not 100% sure

index.d.ts
```diff
@@ -1,4592 +1,4 @@
...
-    starred_url: string;
-    subscriptions_url: string;
-    organizations_url: string;
-    repos_url: string;
-    events_url: string;
-    received_events_url: string;
-    type: string;
-    site_admin: boolean;
-  };
-  type WebhookPayloadMetaHookConfig =
```
oscard0m: Is it ok to rely on a `lib/generated` folder from root level file? (even though the file is also generated?)

oscard0m (commented):
Let's try to agree on the skeleton and I can draft a PR from here!

Thanks

index.d.ts
gr2m: Yes, sure! I depend on generated code in many of my libraries. I only put them in a `generated/` folder to clearly signal that these files are generated and should not be edited manually

lib/generated/api.d.ts
gr2m: Yes

oscard0m:
- `/generated` files on last commit are generated already by generate-scripts
- used `PascalCase` for naming
- should we implement a kind of test to assure the old `index.d.ts` is equivalent to the new generated files? Or with `typescript-validate.ts` check is enough?

### TypeScript migration Questions
- Which `tsconfig.json` should I use, my own proposal? To take [this one](https://github.com/octokit/rest.js/blob/master/tsconfig.json) as example?
- Should it be done on the same PR or in a different one (taking this branch as base-branch for example)
- Should we implement unit tests for `/scripts/` folder? If so, jest with `.test.ts` as it's done [here](https://github.com/octokit/rest.js/tree/master/test/scenarios)?

wolfy1339:
The TypeScript build should be handled by pika, just like in the other Octokit repositories. Just copy the configuration, and maybe add some necessary additions.

If you need any help, just ping gr2m

gr2m:
What @wolfy1339 says. Also don't worry about cleaning up the commit history, it will be easier to follow your changes if you keep the separate commits. When the PR is ready I can just squash & commit

oscard0m:
Sorry for laaaate commit here @gr2m and @wolfy1339, I hope is a good time to follow up on this Pull Request. Would you mind to validate my last commit so I can migrate the rest of files to TypeScript?

I would like you to check:
* `tsconfig.json`: If you like the config, if you miss or you think we should remove something
* comments I left inside _Changed files_
* any other valuable feedback :)

Thanks!

---

These last changes depend on: https://github.com/octokit/webhooks/pull/83 and https://github.com/octokit/webhooks/pull/82

tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+		"resolveJsonModule": true,
+		"strict": true,
+		"target": "es2018"
+	},
+	"include": [
+		"event-handler/**/*",
+		"middleware/**/*",
+		"scripts/**/*",
+		"sign/**/*",
+		"verify/**/*"
```
dominguezcelada, line 18: Should it be considered to move all the project to `src/*` so we can do this?
```json
   "src/**/*"
```

tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+	]
```
oscard0m: [❓ ] Should it be considered to move all the project to `src/*` so we can do this?
```json
   "include": ["src/**/*"]
```

scripts/scripts.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module '@gimenete/type-writer';
+declare module 'table-builder';
```
oscard0m: [❓ ] I'm not sure if this is the right approach when depending on 3rd party dependencies which are not exposing its own declaration. Is this correct?

wolfy1339: Not exactly, this would make everything have type of `any` or throw errors because there are no exported properties.

You'll need to document all exports from the imported modules used in the codebase

tsconfig.json
wolfy1339: That would greatly simplify things

scripts/scripts.d.ts
oscard0m: Naming for declaration file?
* `scripts/sccript.d.ts`
* `scripts/index.d.ts`
* same but in root level of the project?

wolfy1339: You could name it `external-deps.d.ts`

oscard0m:
I would need some help or opinion here @wolfy1339. Trying to migrate `event-handler` to TypeScript I'm getting errors when running `tap` with tests migrated to `.ts`. I drafted a PR [here](https://github.com/dominguezcelada/webhooks.js/pull/7) where you can checkout and check:

`> tsc --noEmit` ✅ 
but
`> npx tap 'test/unit/event-handler-on-test,ts'` 🔴 

Should it be invested effort on migrating these tests to TypeScript? I did not find a way to make `tap`, should it work out of the box in theory (https://node-tap.org/docs/using-with/#using-tap-with-typescript).

Thanks for your time and help once again, and sorry for not being much autonomous here.

wolfy1339:
Not a problem, asking lots of questions is how you will learn for the next time you contribute.

I'm not exactly sure why we're using tap here, since we use jest in the other repos.
I would not rewrite the tests in TypeScript as part of this PR, that can be done as another PR after this one

oscard0m:
> I would not rewrite the tests in TypeScript as part of this PR, that can be done as another PR after this one

The problem on this is that having files in TypeScript and being imported in .js with `tap` make the test fail too. 

Ideas that came to my mind are:
* Create a first PR, coming before this one, migrating to `jest` for example. For this I guess we would need to understand why `tap` was the solution chosen. @gr2m 
* Invest more effort on digging why `tap` is not playing ok with this TypeScript migration
   * Finding a workaround/loader to make `.test.js` files play with imported `.ts` files
   * Solving why test do not compile when using `.test.ts`

tsconfig.json
gr2m: Yes, I did the same with all the other repos as part of the migration to TypeScript

gr2m:
> I'm not exactly sure why we're using tap here, since we use jest in the other repos

I migrated all the other repos to Jest as part of the mgiration to TypeScript, feel free to do the same

gr2m:
what I did during the migration is to keep the tests as they are, written in JavaScript using `tap` or mocha or whatever I used for testing, but require the library from the `pkg/` folder (built by pika). When all the previous tests passed, push the current status, then rewrite all tests in TypeScript/Jest

oscard0m:
> what I did during the migration is to keep the tests as they are, written in JavaScript using `tap` or mocha or whatever I used for testing, but require the library from the `pkg/` folder (built by pika). When all the previous tests passed, push the current status, then rewrite all tests in TypeScript/Jest

Thanks for the advice @gr2m . I'm going for that approach. Do you mind the migration is applied by baby steps?

Here is a [PR](https://github.com/octokit/webhooks.js/pull/122) for migrating to `src/` folder and apply a first pika configuration to work from the final migration to TypeScript.

Is it ok to land this and iterate from here under this PR?

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+	"presets": ["@babel/preset-env"]
+}
```
oscard0m, line 3: **[opinion ❓]** Do you agree on this preset as solution to transpile ES Modules when running `pika`?

test/typescript-validate.ts
```diff
@@ -1,4 +1,5 @@
-import * as Webhooks from "..";
+import { Webhooks } from "..";
+import { EventNames } from "../lib/generated/event-names";
```
oscard0m: **[opinion❓]** Is a good approach to rely on `../lib/generated` or should we think on a structure of export { ... } inside ./index.d.ts to expose `EventNames` and `EventPayloads` from there?

.babelrc
wolfy1339: In `@pika/plugin-build-node`, the option to use custom babelrc files is disabled.This wouldn't have an effect
https://github.com/pikapkg/builders/blob/master/packages/plugin-build-node/src/index.ts#L47

test/typescript-validate.ts
wolfy1339: I think it could be a good idea to export them from the `index.d.ts`

oscard0m:
@gr2m @wolfy1339 Starting the migration of the tests to point to `pkg/...` is giving me problems. I [drafted a PR](https://github.com/octokit/webhooks.js/pull/125) with my issue, I hope we can get some ideas on how to bypass this.

Thanks!

.babelrc
oscard0m: > In `@pika/plugin-build-node`, the option to use custom babelrc files is disabled.This wouldn't have an effect
> [pikapkg/builders:packages/plugin-build-node/src/index.ts@`master`#L47](https://github.com/pikapkg/builders/blob/master/packages/plugin-build-node/src/index.ts#L47) [](/pikapkg/builders/blob/HEAD@{2020-04-16T21:43:21Z}/packages/plugin-build-node/src/index.ts#L47)

Then... I would say to *not* migrate to ES Modules on this PR so we are less aggressive on these changes and in a new PR/Issue it can be discussed how to tackle this (if it's interesting to cover). What do you think? @wolfy1339

wolfy1339: Convert all the files to TypeScript and keep using the same `import`/`export` style as you are already doing.

We could make our own plugin in the meantime until pika releases their own ESM builder, best to leave that for another PR though.

oscard0m: > Convert all the files to TypeScript and keep using the same `import`/`export` style as you are already doing.
> 
> We could make our own plugin in the meantime until pika releases their own ESM builder, best to leave that for another PR though.

Sounds good to me :)

oscard0m:
I think the PR is ready for a review closer to production:
* I still need to solve the issue pointing to `dist-src`, I hope I can figure it out with @gr2m 
* Any other suggestion or feedback regarding the migration to typescript or any other change from this PR

I hope we are close on merging this :)

index.d.ts
```diff
@@ -1,4731 +1,8 @@
 // DO NOT EDIT THIS FILE DIRECTLY
 // make edits in scripts/generate-types.js
 
-import http = require("http");
+import { EventNames } from "./lib/generated/event-names";
+import { EventPayloads } from "./lib/generated/event-payloads";
+import { Webhooks } from "./lib/generated/api";
```
gr2m: Instead of importing and then exporting, you can just do the following

```suggestion
export { EventNames } from "./lib/generated/event-names";
export { EventPayloads } from "./lib/generated/event-payloads";
export { Webhooks } from "./lib/generated/api";
```

package.json
```diff
@@ -5,7 +5,7 @@
     "access": "public"
   },
   "description": "GitHub webhook events toolset for Node.js",
-  "main": "src/index.js",
+  "main": "src/index.ts",
```
gr2m: you can remove the "main" key, `pika` sets it in `pkg/package.json` automatcially

src/event-handler/receive.ts
```diff
@@ -1,20 +1,19 @@
...
-    errorHandlers.forEach((handler) => wrapErrorHandler(handler, event));
+    errorHandlers.forEach((handler: Function) =>
+      wrapErrorHandler(handler, event)
+    );
 
     return Promise.reject(event);
   }
 
-  if (!event || !event.name) {
+  if (!event || !event.name || Array.isArray(event.name)) {
```
gr2m: why the additional check ` || Array.isArray(event.name)`?

src/middleware/README.md
```diff
@@ -3,8 +3,8 @@
 If you only need the middleware with access to the `.sign()`, `.verify()` or the receiver’s `.receive()` method, you can use the webhooks middleware directly
 
 ```js
-const Middleware = require('@octokit/webhooks/middleware')
-const middleware = new Middleware({
+import { createMiddleware } from '@octokit/webhooks/middleware'
```
gr2m: This should be 

```ts
import { createMiddleware } from '@octokit/webhooks'
```

With the migration to typescript / ESM exports, I'd export everything at the root package.

gr2m: Also if at all possible, I'd like to avoid any breaking changes. Either we make the previous `require('@octokit/webhooks/middleware')` work and log a deprecation message, or we make the change in the current version, where we make the new export work and start logging deperecation messages for the current usage

src/middleware/get-missing-headers.ts
```diff
@@ -1,4 +1,4 @@
-module.exports = getMissingHeaders;
+import { Request } from "express";
```
gr2m: Why importing Request from "express"? This is not an express middleware, but a Node.js middleware. The two are compatible. You can 

```
import { IncomingMessage} from "http"
```

and then do

```ts
request: typeof IncomingMessage
```

src/index.d.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+type Hooks = {
+  [key: string]: Function[];
+};
+
+export interface EventState extends EventHandlerOptions {
+  eventHandler?: any;
+  hooks: Hooks;
+}
```
gr2m: In the other Octokit repos I have a `src/types.ts` file if I need to define any internal types. If need to export the types, I just export them in `src/index.ts`. I don't have any `*.d.ts` files in the other Octokit repos. I wonder if you could follow that convention here?

src/middleware/isnt-webhook.ts
```diff
@@ -1,10 +1,10 @@
...
-
 // Example webhook event request:
 // https://developer.github.com/webhooks/#example-delivery
-function isntWebhook(request, options) {
+import { Request } from "express";
+
+export function isntWebhook(request: Request, options: { path?: string }) {
   // GitHub sends all events as POST requests
-  if (request.method !== "POST") {
+  if (request.method !== "POST" || !request.url) {
```
gr2m: why the additional check?

src/scripts/generate-file.ts
```diff
@@ -0,0 +1,7 @@
+import fs from "fs";
+import prettier from "prettier";
+
+export const generateFile = (filepath: string, content: string) => {
+  const output = prettier.format(content, { filepath });
```
gr2m: I think you need to add a `parser` option, e.g. compare 

https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/53e3403d39e35b25ac3b5ae872895ab4277067cd/scripts/update-endpoints/types.js#L289-L310

gr2m (commented):
I think we should leave the tests as they are, but let's get back to that later.

I've left some comments. We also have to deprecate current usage before we break it with this PR. I think the simplest would be to create a separate PR which exports everything in `lib/index.js` the same way this PR is (or will) export all the things. The previous usage should log deprecation messages. That will go a long way to help people migrate to the new version.

I think once we export everything from the root pacakge, we won't need to do many changes in the test files in this pull request, besides importing from `../../pkg`

src/middleware/get-missing-headers.ts
oscard0m: > import { IncomingMessage} from "http"

It's what I tried on first instance but I got type issues from here:
https://github.com/octokit/webhooks.js/blob/d675137b32c5867c3fe5f1d204a8075b85cb83d1/src/middleware/get-payload.js#L4-L8

image

For the rest of places, the type it's fine. Checking the PR of those lines, seems to be relying on body property. Researching a bit the origin of `body-parse` I ended up finding express types as origin (and those are an extension from `IncomingMessage` of `http`)

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/069f404e6611098c620a265d05782db36efb4fc6/types/express/index.d.ts#L22

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/069f404e6611098c620a265d05782db36efb4fc6/types/express-serve-static-core/index.d.ts#L216

---

[❓] How should we proceed here? 
* Apply everywhere ` { IncomingMesage } from 'http'` and `{ Request } from 'express'` for `src/middleware/get-payload.ts`
* Apply everywhere ` { IncomingMesage } and remove if `request.body` since it won't be possible to happen anymore due to type restrictions
* Keep `express` types everywhere

src/middleware/isnt-webhook.ts
oscard0m: > why the additional check?

Mmm... at some point I was getting coverage issues. I tried to reproduce it now to share it but... not happening anymore so... removing this extra unnecessary check :)

src/event-handler/receive.ts
oscard0m: > why the additional check ` || Array.isArray(event.name)`?

Applying types to `WebhookEvent` it determines the format can be an array:
https://github.com/dominguezcelada/webhooks.js/blob/12b6714cb2693fa651b67462f69cc2789099a1c2/src/index.d.ts#L1-L6

because of `http.d.ts` (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/069f404e6611098c620a265d05782db36efb4fc6/types/node/v12/http.d.ts#L63)

So, to avoid type errors in the code I had to apply this extra check before continuing the execution:

image

src/index.d.ts
oscard0m: > In the other Octokit repos I have a `src/types.ts` file if I need to define any internal types. If need to export the types, I just export them in `src/index.ts`. I don't have any `*.d.ts` files in the other Octokit repos. I wonder if you could follow that convention here?

Of course! Sorry to not check the file structure for types in other repos 🙏 . Applying it!

src/scripts/generate-file.ts
oscard0m: > I think you need to add a `parser` option, e.g. compare
> 
> [octokit/plugin-rest-endpoint-methods.js:scripts/update-endpoints/types.js@`53e3403`#L289-L310](https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/53e3403d39e35b25ac3b5ae872895ab4277067cd/scripts/update-endpoints/types.js#L289-L310)

Seems that Prettier [infers it](https://prettier.io/docs/en/options.html#parser) from the file extension by default. If you want I can add it as an optional parameter on this utility function :)

src/middleware/README.md
oscard0m: > This should be
> 
> ```ts
> import { createMiddleware } from '@octokit/webhooks'
> ```
> 
> With the migration to typescript / ESM exports, I'd export everything at the root package.

Applied :)

oscard0m: > Also if at all possible, I'd like to avoid any breaking changes. Either we make the previous `require('@octokit/webhooks/middleware')` work and log a deprecation message, or we make the change in the current version, where we make the new export work and start logging deperecation messages for the current usage

I can open a new PR from `master` adding this first step of logging and changing entrey points and I rebase this branch with master once the changes are landed. Do you like the idea?

oscard0m:
> I think we should leave the tests as they are, but let's get back to that later.

👌 

> I've left some comments.

Thanks for your time and for the comments :)

> We also have to deprecate current usage before we break it with this PR. I think the simplest would be to create a separate PR which exports everything in `lib/index.js` the same way this PR is (or will) export all the things. The previous usage should log deprecation messages. That will go a long way to help people migrate to the new version.

Sure! Let's go for it!

> I think once we export everything from the root pacakge, we won't need to do many changes in the test files in this pull request, besides importing from `../../pkg`

👌

src/scripts/scripts.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module "@gimenete/type-writer";
+declare module "table-builder";
```
oscard0m: //TODO: Create proper modules for third party dependencies

oscard0m:
> > We also have to deprecate current usage before we break it with this PR. I think the simplest would be to create a separate PR which exports everything in `lib/index.js` the same way this PR is (or will) export all the things. The previous usage should log deprecation messages. That will go a long way to help people migrate to the new version.
> 
> Sure! Let's go for it!

https://github.com/octokit/webhooks.js/pull/126 @gr2m

gr2m:
I only realized after merging #126, a few things:

1. We should get rid of the `lib/` folder altogether
2. The README.md files in the `src/` folder need to be updated

I'd do that in a separate PR before continuing work on this PR

oscard0m:
> I only realized after merging #126, a few things:
> 
> 1. We should get rid of the `lib/` folder altogether
> 2. The README.md files in the `src/` folder need to be updated
> 
> I'd do that in a separate PR before continuing work on this PR

Drafted a PR with some questions regarding these changes: https://github.com/octokit/webhooks.js/pull/127

oscard0m:
@gr2m @wolfy1339 I just pushed some new changes on this branch:
- Fix conflicts with latest `master` changes
- Move src/scripts/*.ts to scripts/*.ts

Is there any change which should be included on this release pending? I'm looking forward to land on `master` with this :)

src/event-handler/receive.ts
gr2m: This is a breaking change of the API. `.on()` accepts both a string and an array of strings. We have to find a way to handle that correctly ... I know it's going to be more tricky with the type definitions, but we collaborate on that

.gitignore
```diff
@@ -3,4 +3,7 @@ coverage
 node_modules
 
 # Pika builds
-pkg/
\ No newline at end of file
+pkg/
+
+# TypeScript generated compiled scripts
+scripts/*.js
```
gr2m: Would you mind to leave the `scripts/` files written in JavaScript please? Having them in TypeScript is causing unnecessary maintenance overhead / complexity.

package.json
```diff
@@ -29,17 +31,22 @@
     "simple-mock": "^0.8.0",
     "table-builder": "^2.1.1",
     "tap": "^14.0.0",
+    "tslib": "^1.11.1",
```
gr2m: could you explain what you do with `tslib`? I've never used that before

src/webhook-names.ts
```diff
@@ -179,5 +179,5 @@
   "team.removed_from_repository",
   "team_add",
   "watch",
-  "watch.started"
-]
+  "watch.started",
+];
```
gr2m: could you move this file to `src/generated/webhook-names.ts`?

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+		"removeComments": true,
+		"resolveJsonModule": true,
+		"strict": true,
+		"target": "es2020"
+	},
+	"include": [
+		"src/**/*",
+		"scripts/**/*"
+	]
+}
```
gr2m: I use the a simpler config across all my other packages, e.g. https://github.com/octokit/core.js/blob/master/tsconfig.json

Could we use the same one here?

gr2m (commented):
I didn't yet get to play with your code changes, but I have some feedback from reviewing your changes, could you look into these

This pull-request introduces the `lib/` folder again, could you change to that to `src/` and replace the `lib/generated/*.d.ts` files with `src/generated/*.ts`?

.babelrc
oscard0m: > Do we still need this now that the source code is written in TypeScript?

If we don't use it the transpilation of `import / export` is not done.

Options that came to my mind:
* keep .babelrc and PR to pika support for import/export or create our own plugin as @wolfy1339 pointed
* transpile it without pika for the moment

### Example pika/dist-src/index.js

<details>
 <summary>*with .babelrc*</summary>
<code>
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.createWebhooksApi = createWebhooksApi;
Object.defineProperty(exports, "createEventHandler", {
  enumerable: true,
  get: function get() {
    return _index.createEventHandler;
  }
});
Object.defineProperty(exports, "createMiddleware", {
  enumerable: true,
  get: function get() {
    return _index2.createMiddleware;
  }
});
Object.defineProperty(exports, "sign", {
  enumerable: true,
  get: function get() {
    return _index3.sign;
  }
});
Object.defineProperty(exports, "verify", {
  enumerable: true,
  get: function get() {
    return _index4.verify;
  }
});

var _index = require("./event-handler/index.js");

var _index2 = require("./middleware/index.js");

var _middleware = require("./middleware/middleware.js");

var _index3 = require("./sign/index.js");

var _index4 = require("./verify/index.js");

var _verifyAndReceive = require("./middleware/verify-and-receive.js");

function createWebhooksApi(options) {
    throw new Error("options.secret required");
  }

  var state = {
    eventHandler: (0, _index.createEventHandler)(options),
    path: options.path || "/",
    secret: options.secret,
    hooks: {}
  };
  return {
    sign: _index3.sign.bind(null, options.secret),
    verify: _index4.verify.bind(null, options.secret),
    on: state.eventHandler.on,
    removeListener: state.eventHandler.removeListener,
    receive: state.eventHandler.receive,
    middleware: _middleware.middleware.bind(null, state),
    verifyAndReceive: _verifyAndReceive.verifyAndReceive.bind(null, state)
  };
}
</code>
</details>

<details>
 <summary>*without .babelrc*</summary>
<code>
import { createEventHandler } from "./event-handler/index.js";
import { createMiddleware } from "./middleware/index.js";
import { middleware } from "./middleware/middleware.js";
import { sign } from "./sign/index.js";
import { verify } from "./verify/index.js";
import { verifyAndReceive } from "./middleware/verify-and-receive.js";

function createWebhooksApi(options) {
    throw new Error("options.secret required");
  }

  const state = {
    eventHandler: createEventHandler(options),
    path: options.path || "/",
    secret: options.secret,
    hooks: {}
  };
  return {
    sign: sign.bind(null, options.secret),
    verify: verify.bind(null, options.secret),
    on: state.eventHandler.on,
    removeListener: state.eventHandler.removeListener,
    receive: state.eventHandler.receive,
    middleware: middleware.bind(null, state),
    verifyAndReceive: verifyAndReceive.bind(null, state)
  };
}

export { createEventHandler, createMiddleware, createWebhooksApi, sign, verify };
```
</code>
</details>

File: src/middleware/get-missing-headers.ts (hunk 12 above)
Comment (oscard0m) [reply src/middleware/get-missing-headers.ts]: ping @gr2m :)

File: src/scripts/generate-file.ts (hunk 15 above)
Comment (oscard0m) [reply src/scripts/generate-file.ts]: Resolving comment @gr2m ;)

File: .gitignore (hunk 17 above)
Comment (oscard0m) [reply .gitignore]: > Would you mind to leave the `scripts/` files written in JavaScript please? Having them in TypeScript is causing unnecessary maintenance overhead / complexity.

Yes! No problem! It will simplify a lot the comments we have in place right now and... if we came up in the future a way to have it in typescript in a simple way we are always on time for that migration :)

File: src/scripts/scripts.d.ts (hunk 16 above)
Comment (oscard0m) [reply src/scripts/scripts.d.ts]: Not an issue anymore after this agreement: https://github.com/octokit/webhooks.js/pull/113#discussion_r416975529

File: scripts/scripts.d.ts (hunk 5 above)
Comment (oscard0m) [reply scripts/scripts.d.ts]: Not an issue anymore after this agreement:
https://github.com/octokit/webhooks.js/pull/113#discussion_r416975529

File: tsconfig.json (hunk 20 above)
Comment (oscard0m) [reply tsconfig.json]: > I use the a simpler config across all my other packages, e.g. [octokit/core.js:tsconfig.json@`master`](https://github.com/octokit/core.js/blob/master/tsconfig.json) [](/octokit/core.js/blob/HEAD@{2020-04-28T23:19:47Z}/tsconfig.json)
> 
> Could we use the same one here?

Yep sure! The only one which was mandatory was the one for resolving JSON but, after restoring JavaScript on `/scripts/*` no necessary anymore.

File: package.json (hunk 18 above)
Comment (oscard0m) [reply package.json]: > could you explain what you do with `tslib`? I've never used that before

I was playing with `--importHelpers` TypeScript flag (you know, trying to discover a bit of TS), and forgot to remove this dependency after playing with it.

**Comment by oscard0m** [comment]
> I didn't yet get to play with your code changes, but I have some feedback from reviewing your changes, could you look into these
> 
> This pull-request introduces the `lib/` folder again, could you change to that to `src/` and replace the `lib/generated/*.d.ts` files with `src/generated/*.ts`?

Yes, sorry, you are right. Applied changes to move everything to `src/generated/*`

File: src/event-handler/receive.ts (hunk 10 above)
Comment (oscard0m) [reply src/event-handler/receive.ts]: > This is a breaking change of the API. `.on()` accepts both a string and an array of strings. We have to find a way to handle that correctly ... I know it's going to be more tricky with the type definitions, but we collaborate on that

Sure! I thought it was not on octokit's end this type, and we only had to cover single string. Let me try removing this check and sort out how to manage this double type of entry :)

Comment (oscard0m) [reply src/event-handler/receive.ts]: Pushed changes for that

File: .babelrc (hunk 6 above)
Comment (gr2m) [reply .babelrc:3]: I'm sorry I'm not following ... it's okay if the `pkg/dist-src/*` files use import/export. That's what I have in the other packages, too. For example: https://unpkg.com/browse/@octokit/rest@17.7.0/dist-src/index.js

The `dist-src` folder is not what is used by dependants, they either use `dist-web/` or `dist-node/`

File: test/unit/event-handler-receive-test.js [hunk 21]
```diff
@@ -1,10 +1,9 @@
-const test = require("tap").test;
-
-const receive = require("../../src/event-handler/receive");
+import { test } from "tap";
+import { receiverHandle as receive } from "../../pkg/dist-src/event-handler/receive";
```
gr2m, line 2: oh is the reason we need the `babelrc` because we want to test internal functionality like this? 

And once we rewrite the tests to TypeScript, we can get rid of babelrc again? I wonder if in this case it'd be simpler to rewrite the tests to TypeScript at the same time ...

oscard0m: > oh is the reason we need the `babelrc` because we want to test internal functionality like this?
> 
> And once we rewrite the tests to TypeScript, we can get rid of babelrc again? I wonder if in this case it'd be simpler to rewrite the tests to TypeScript at the same time ...

As we commented [here](https://github.com/octokit/webhooks.js/pull/113#issuecomment-612578679) it would imply migrating from tap to jest + TypeScript. Don't you think is better to do it in 2 steps to not introduce more potential friction on these breaking changes? What do you think? @gr2m

.babelrc
oscard0m: > I'm sorry I'm not following ... it's okay if the `pkg/dist-src/*` files use import/export. That's what I have in the other packages, too. For example: [unpkg.com/browse/@octokit/rest@17.7.0/dist-src/index.js](https://unpkg.com/browse/@octokit/rest@17.7.0/dist-src/index.js)
> 
> The `dist-src` folder is not what is used by dependants, they either use `dist-web/` or `dist-node/`

It's because of what you pointed [here](https://github.com/octokit/webhooks.js/pull/113#discussion_r420313771).

test/unit/event-handler-receive-test.js
gr2m: Got it, thanks for reminding me. Okay, let's use babelrc then until we migrate to Jest + TypeScript 👍 

I'll test your changes locally now

index.d.ts
```diff
@@ -1,5004 +1,6 @@
 // DO NOT EDIT THIS FILE DIRECTLY
 // make edits in scripts/generate-types.js
```
gr2m, line 2: I think these lines can now be removed, as the file is no longer generated?

gr2m:
Is it possible that somewhere along the way we lost the TypeScript intellisense for `@octokit/webhooks`? When I check out v7.2.0, I get the right typings for e.g. the `.on` method:

image

Trying the same in v7.3.0 or your PR, I don't get any typings. Am I doing anything wrong? 

I don't mean to put any blame on you, you are doing great work here. I wish there was a way to test that the typings don't break that way ...

Also it looks like we introduce a breaking change with this PR, because the default export is no longer working?

```ts
// test.ts in root folder of project
import Webhooks from "./pkg";

const events = new Webhooks({
  secret: "secret",
});

events.on("");
```

oscard0m:
> Is it possible that somewhere along the way we lost the TypeScript intellisense for `@octokit/webhooks`? When I check out v7.2.0, I get the right typings for e.g. the `.on` method:
> 
> image
> 
> Trying the same in v7.3.0 or your PR, I don't get any typings. Am I doing anything wrong?

Hey @gr2m, trying 7.5.0 and it's working to me:
image

also with 7.3.0:
image

My PR (https://github.pika.dev/octokit/webhooks.js/pr/113) only works with a named import, I thought it was the intention to deprecate the default export here. It's true that with my branch Intellisense is not autocompleting:

image

* Should `new Webhooks()` with default export still be usable? (i thought we wanted everything to be named exports, actually I changed the `new Webhooks()` for the method in the tests)
* I'm going to check why Intellisense stopped working on my branch. In your case with 7.3.0, could you describe how are you trying it? is that VSCode?

> I don't mean to put any blame on you, you are doing great work here. I wish there was a way to test that the typings don't break that way ...

I know, I know... I see you are not blaming. It's fine 😄 👍 
Let me find out what was the issue and when we find it we can figure out a way to cover this case in the future.

> Also it looks like we introduce a breaking change with this PR, because the default export is no longer working?
> 
> ```ts
> // test.ts in root folder of project
> import Webhooks from "./pkg";
> 
> const events = new Webhooks({
>   secret: "secret",
> });
> 
> events.on("");
> ```

Yep, commented before

oscard0m:
> > Is it possible that somewhere along the way we lost the TypeScript intellisense for `@octokit/webhooks`?

@gr2m 

Playing a bit more what happens, to me at least, but also with older versions, is that Intellisense messes up when trying to solve `index.js` and `index.d.ts`. If you <kbd>CDM + click</kbd> on the import `@octokit/webhooks` it `Goes to Implementation` -> `index.d.ts`. Maybe it's because of [this](https://github.com/microsoft/vscode-docs/blob/master/docs/editor/intellisense.md#why-am-i-not-seeing-method-and-variable-suggestions)?

Checking on the Internet it seems to be a known issue:
* https://github.com/microsoft/TypeScript/issues/6209#issuecomment-321545425
* https://github.com/microsoft/TypeScript/issues/14479
* https://github.com/microsoft/vscode/issues/30604

Is this the issue you are experiencing maybe?

(on my PR branch, `index.d.ts` is not at root level of the project anymore so I guess this is the reason Intellisense does not work anymore)

I would like to get some feedback here:
* Should I restore the default export on `index.js`?
* Are these issues I pointed the issue you are experimenting with Intellisense?

---

Now I'm playing with `["@pika/plugin-build-types"]` to try to expose the types but I'm having problems with  it since it's not compiling the imports: https://github.com/octokit/webhooks.js/pull/132

Waiting for your feedback before continuing.

gr2m:
I tried to install the build package by pika: (https://github.com/octokit/webhooks.js/pull/113/checks?check_run_id=647456821):

```
npm install https://github.pika.dev/octokit/webhooks.js/pr/113
```

with this code

```js
const { createEventHandler } = require("@octokit/webhooks");
const webhooks = createEventHandler();
```

And it throws the error 

```
Cannot find module './event-handler/index.js'
```

🤔 

When I install `@octokit/webhooks@7.5.0`, then I don't get the intellisense in VS code with this code:

```js
const { createWebhooksApi } = require("@octokit/webhooks");
const webhooks = createWebhooksApi({
  secret: "funky",
});

webhooks.on("");
```

But it does with with this code

```js
const WebHooks = require("@octokit/webhooks");
const webhooks = new WebHooks({
  secret: "funky",
});

webhooks.on("");
```

Any idea?

oscard0m:
> Any idea?

@gr2m I just opened a PR to "fix" Intellisense on `master`: https://github.com/octokit/webhooks.js/pull/134

Once we agree on the fix on that PR and make sure `master` is stable I will check what's the issue in this branch.

Thanks

scripts/generate-types.js
```diff
@@ -1,88 +1,128 @@
...
+    [`${typeName}.repository`]: "PayloadRepository",
+    // This prevents a naming colision between the payload of a `installation_repositories` event
+    // and the `repositories` attribute of a `installation` event
+    "WebhookPayloadInstallation.repositories":
+      "WebhookPayloadInstallation_Repositories",
+  },
+});
+
+const generateEventType = (event, typeName) => `
+  public on (
```
ankeetmaini: We can use a Generic type to achieve this in one go. Greater flexibility to compose too. #141

ankeetmaini: I can port the approach here if it seems okay. :)

gr2m:
Hey @dominguezcelada what's the status with this PR?

oscard0m:
> Hey @dominguezcelada what's the status with this PR?

I've been disconnected from these PR lately. Let me catch up and post an update!

oscard0m:
> Hey @dominguezcelada what's the status with this PR?

Sorry for the delay on this PR @gr2m . I've not been quite available these days but... I'm back 💪 !!

Checking last comments what is missing now is:
* Fix tests failing (I'm going to take a look right now)
* Have type declaration shiny and working
   * Integrate latest changes on `master` regarding `index.d.ts` (I guess the output of this PR(https://github.com/octokit/webhooks.js/pull/153) and @ankeetmaini latest changes(https://github.com/octokit/webhooks.js/pull/141)
   * Validate Intellisense will work with these changes (reference: https://github.com/octokit/webhooks.js/pull/113#issuecomment-627609117)

---

Regarding Intellisense in `master`, a part from this pending issue (https://github.com/octokit/webhooks.js/pull/153), are we fine? I think it's important to have all those things sorted out before moving forward with this.

oscard0m:
- [x] Tests fixed
- [x] Adopted approach from master. @ankeetmaini if you have time and you have interest on it, could you please take a look on the changes I just applied? I'm totally open for suggestions and bugs :)
- [ ] Pending to run pre-released package using these changes but with the coverage we have right now on `typescript-validate.ts` looks promising. As soon as I test the package I will re-request a review

❓ To use ENUMs is not possible. Is not the same a union of strings than a controlled set of strings under an ENUM. I guess with the types we are putting in place is enough to avoid typos from users, the original purpose of this PR :P

@gr2m

ankeetmaini:
Will check this today!

gr2m:
Thanks for the update @dominguezcelada, thanks for your help @ankeetmaini!

src/event-handler/on.ts
```diff
@@ -1,8 +1,11 @@
-module.exports = receiverOn;
+import { webhookNames } from "../generated/webhook-names.js";
+import { EventState } from "../types";
 
-const webhookNames = require("../webhook-names.json");
-
-function receiverOn(state, webhookNameOrNames, handler) {
+export function receiverOn(
+  state: EventState,
+  webhookNameOrNames: string | string[],
```
ankeetmaini: shouldn't this be a generic type as done in the type definition file?

src/event-handler/receive.ts
```diff
@@ -1,15 +1,32 @@
-"use strict";
+import { wrapErrorHandler } from "./wrap-error-handler";
+import { WebhookEvent, EventState } from "../types";
+
+function getHooks(
+  state: EventState,
+  eventPayloadAction: string,
+  eventName: string | string[]
```
ankeetmaini: this too I thing we can use AllEventTypes

src/generated/api.d.ts
```diff
@@ -0,0 +1,152 @@
+import http = require("http");
```
ankeetmaini: since this is auto generated, can this not be in src control and get generated as part of build process?

ankeetmaini (commented):
The changes look awesome! Asked 2 small questions. Keeping generated types in code or directly let them get generated as part of build and referenced by a types field in package.json.

src/generated/api.d.ts
oscard0m: > since this is auto generated, can this not be in src control and get generated as part of build process?

It was done but during the development we ended up moving it out (https://github.com/octokit/webhooks.js/pull/113/commits/f9480b417b649de3f433b8177e0bc24b9467dbac) of `src/` folder.

Context here: https://github.com/octokit/webhooks.js/pull/127#issuecomment-620301310

---

Even though I would like to hear more about what's the idea you have in mind. Maybe has sense! @gr2m pay attention here 🍿 

Whatever we decide here. Would you agree on discussing it on a different issue and PR? This way we would not add more noise / breaking changes to this 😇

src/event-handler/on.ts
oscard0m: > shouldn't this be a generic type as done in the type definition file?

You are totally right: https://github.com/octokit/webhooks.js/pull/113/commits/a521da0d23143c3045b6aec5205a24d4da0d6977

src/event-handler/receive.ts
oscard0m: > this too I thing we can use AllEventTypes

Same here. You are totally right: https://github.com/octokit/webhooks.js/pull/113/commits/a521da0d23143c3045b6aec5205a24d4da0d6977

oscard0m:
> * [ ]  Pending to run pre-released package using these changes but with the coverage we have right now on `typescript-validate.ts` looks promising. As soon as I test the package I will re-request a review

Type generation with Pika seems to not be working fine. I'm taking a look into it but if you know what could be the issue let me know! 😅

Should be pointing to `dist-src/...` instead of `src/...`
image

image

src/generated/api.d.ts
gr2m: > can this not be in src control

It's not a comparable to a built artefact like JS files in project written in TypeScript. This file is part of the actual source, it is automated based on https://github.com/octokit/webhooks via automated pull requests. In these pull requests we want to be able to review the differences in the source files that we publish to npm

gr2m:
the types should be exported when you import / require directly from `./pkg`. `pkg/package.json` should have all the keys in order to import the right version.

I think the challenge is that the tests are written in ESM now. I think it would be easier to keep them in CJS. If this becomes too much of a headache, I'd just go ahead and rewrite the tests in TypeScript, too, that should resolve all import issues

oscard0m:
> the types should be exported when you import / require directly from `./pkg`. `pkg/package.json` should have all the keys in order to import the right version.
> 
> I think the challenge is that the tests are written in ESM now. I think it would be easier to keep them in CJS. If this becomes too much of a headache, I'd just go ahead and rewrite the tests in TypeScript, too, that should resolve all import issues

Even if I apply the path `@octokit/webhooks`, the imports do not work because pika is not generating all the files:
image
🆚 
image

```typpescript
// DO NOT EDIT THIS FILE DIRECTLY
// make edits in scripts/generate-types.js

export { EventNames } from "@octokit/webhooks/generated/event-names";
export { EventPayloads } from "@octokit/webhooks/generated/event-payloads";
export {
  Webhooks,
  createWebhooksApi,
  createEventHandler,
  createMiddleware,
  sign,
  verify,
} from "@octokit/webhooks/generated/api";
```

Before evaluating how to make the tests works we need to figure out how to tackle this. Do you know what could be the issue here? Do I debug pika's plugin? @gr2m

gr2m:
Let me look into it tomorrow

oscard0m:
> Let me look into it tomorrow

Thanks for your help and your patience once again ;)

gr2m:
I've investigated a bit but went down a rabbit whole. A few observations:

- we will need `@pika/plugin-ts-standard-pkg`, not `@pika/plugin-standard-pkg` for the pika build pipeline.
- the imports cannot point to a folder, they need to point to a file, e.g. `import { createEventHandler } from "./event-handler/index";` instead of `import { createEventHandler } from "./event-handler";`
- I'd just use requires in the tests instead of import.  Only require `../../pkg`. E.g. do `const { Webhooks } = require("../../pkg");`. You can ignore the unit tests and 100% test coverage for now, we can re-enable them once we rewrite the test files to typescript, too.
- Replace the `.d.ts` files with `.ts` files
- Remove `/index.d.ts` and the types setting in `package.json`. Pika creates its own package.json in `pkg/package.json` and sets the types key accordingly.

you can see my work-in-progress changes here:
https://github.com/octokit/webhooks.js/commit/3fa5ec4955f6f3905ddeb52e561e4a9052d97b32

I hope that helps

gr2m:
Hey @dominguezcelada where do you stand with the pull request right now?

oscard0m:
> Hey @dominguezcelada where do you stand with the pull request right now?

- [x] Applied @pika/plugin-ts-standard-pkg (https://github.com/octokit/webhooks.js/pull/113/commits/fa849cd5c26695f320acaac4a48759f4d4fa213f)
- [x] Applied /index to paths. (https://github.com/octokit/webhooks.js/pull/113/commits/256a9b9b85c7b5358e02d01b959a70ec988b542e)
`e.g. import { createEventHandler } from "./event-handler/index"; instead of import { createEventHandler } from "./event-handler";` 
- [x] Adopt path `../../pkg.` for tests. (https://github.com/octokit/webhooks.js/pull/113/commits/d45ffefbc7a821c509040ea18e9a2ef137e5112d)
- [x] Replace the .d.ts files with .ts files (https://github.com/octokit/webhooks.js/pull/113/commits/cd6069ced15582dceb81540c6f8eadce7cd8968e)
- [x] Remove /index.d.ts and the types setting in package.json. Pika creates its own package.json in pkg/package.json and sets the types key accordingly. (https://github.com/octokit/webhooks.js/pull/113/commits/c4639175847d4988498515e0ead8ae4263381714) (https://github.com/octokit/webhooks.js/pull/113/commits/b2488f092dbdf14289fd04b8689b25358e8ec3c2)

----
## TODO
- [x] Running `npm run build` broken. After last changes, typing mismatch between `Request | Response` and `IncomingMessage | ServerResponse`. It was discussed in the past (https://github.com/octokit/webhooks.js/pull/113#discussion_r411004564)
- [ ] To fix test-suite to latest changes and get all of them passing ✅ 
- [ ] https://github.com/octokit/webhooks.js/pull/113#discussion_r453990152
- [ ] To test pika pre-published package to validate all the changes
- [x] Apply missing changes from https://github.com/octokit/webhooks.js/pull/161

src/middleware/get-payload.ts
```diff
@@ -1,19 +1,19 @@
...
-function getPayload(request) {
+export function getPayload(request: IncomingMessage): Promise<Payload> {
   // If request.body already exists we can stop here
   // See https://github.com/octokit/webhooks.js/pull/23
-  if (request.body) {
-    return Promise.resolve(request.body);
-  }
+  // if (request.body) {
+  //   return Promise.resolve(request.body);
+  // }
```
oscard0m: @gr2m with latest typing changes I had to disable this lines because `request.body` is not an existing property.

Do you have more context on what was the issue in this scenario? https://github.com/octokit/webhooks.js/pull/23

gr2m: I think in serverless environments such as AWS lambda, `request.body` is already present. And if it is, we can directly return it, instead of trying to retrieve it again, which will fail

src/middleware/isnt-webhook.ts
```diff
@@ -4,14 +4,14 @@ import { IncomingMessage } from "http"
 
 export function isntWebhook(request: IncomingMessage, options: { path?: string }) {
   // GitHub sends all events as POST requests
-  if (request.method !== "POST") {
+  if (request.method !== "POST" || !request.url) {
```
gr2m: I don't think that the `!request.url` check belongs here, given the comment. We should only check for the method here. Check for the URL below

src/middleware/isnt-webhook.ts
```diff
@@ -4,14 +4,14 @@ import { IncomingMessage } from "http"
...
     return true;
   }
 
   // We must match the configured path to allow custom POST routes which include
   // the webhook route. For example if the webhook route is / then it would be
   // impossible to define a `POST /my/custom/app` route as the `POST /`.
-  if (request.url?.split("?")[0] !== options.path) {
+  if (request.url.split("?")[0] !== options.path) {
```
gr2m: If the `?` makes problems, we could do 

```suggestion
  if (typeof request.url !== 'string' || request.url.split("?")[0] !== options.path) {
```

src/middleware/get-payload.ts
oscard0m: > I think in serverless environments such as AWS lambda, `request.body` is already present. And if it is, we can directly return it, instead of trying to retrieve it again, which will fail

I see. Do you have any idea of how we should give support to this and at the same time be `@types/http` compliant? 

Checking which I think is the official [aws-lambda](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/aws-lambda) type definition apparently is not extending from `@types/http.IncomingMessage`

@gr2m

oscard0m: What about 
```typescript
export function getPayload(request: IncomingMessage & { body?: Payload }): Promise<Payload> {
```

too hacky?

gr2m:
Hey @dominguezcelada is there anything I can help with to get this over the finish line? I'd love the TypeScript transition to finish soon, it would help me with https://github.com/probot/probot/pull/1250

oscard0m:
> probot/probot#1250

Hey @gr2m ! I'm struggling a bit on last mile on this. The remaining steps to solve are:
- [ ] [This discussion](https://github.com/octokit/webhooks.js/pull/113#discussion_r457886534)
- [ ] With latest changes server-test is failing Not sure if the issue is the usage of ESModules or a different thing. Here I would appreciate some help. Probably you would find it super fast what's the issue:

image


Once all this is stable, to test back a generated pika package so we double check all the types are right and we would be ready to go 💪

oscard0m:
> > [probot/probot#1250](https://github.com/probot/probot/pull/1250)
> 
> Hey @gr2m ! I'm struggling a bit on last mile on this. The remaining steps to solve are:
> 
> * [x]  [This discussion](https://github.com/octokit/webhooks.js/pull/113#discussion_r457886534)
> * [x]  With latest changes server-test is failing Not sure if the issue is the usage of ESModules or a different thing. Here I would appreciate some help. Probably you would find it super fast what's the issue:
> 
> image
> 
> Once all this is stable, to test back a generated pika package so we double check all the types are right and we would be ready to go 💪

---

Also, if there is something else I could test before doing a release regarding your probot mentioned changes I would be more than happy to help there!

gr2m:
>  With latest changes server-test is failing Not sure if the issue is the usage of ESModules or a different thing. Here I would appreciate some help. Probably you would find it super fast what's the issue

Sure, I'll have a look into that

oscard0m:
> > With latest changes server-test is failing Not sure if the issue is the usage of ESModules or a different thing. Here I would appreciate some help. Probably you would find it super fast what's the issue
> 
> Sure, I'll have a look into that

Seems to be tap's coverage for this file: 
image

src/middleware/get-payload.ts
gr2m: This change is causing the error with the server test you see right now.

How about we do this?
```suggestion
  // @ts-ignore
  if (request.body) return Promise.resolve(request.body);
```

I think it's more explicit then altering the `request` type

gr2m:
Turns out the two remaining points are related and should both be resolved with my suggestion here:
https://github.com/octokit/webhooks.js/pull/113#discussion_r458352131

oscard0m:
> Turns out the two remaining points are related and should both be resolved with my suggestion here:
> [#113 (comment)](https://github.com/octokit/webhooks.js/pull/113#discussion_r458352131)

Seems to be failing the tap --coverage part:
image

Getting deeper but I would say is something with ESModules tap (and node+nyc inside). Is it working in your local? (Playing with this now: https://github.com/tapjs/node-tap/issues/503 https://github.com/tapjs/node-tap/issues/604). Not sure if related.

If you run --no-coverage it works.

gr2m:
I'm looking into that right now.

It might be related to the fact that you rewrote the tests to ES Modules, maybe tap/nyc cannot handle that correctly

oscard0m:
> I'm looking into that right now.
> 
> It might be related to the fact that you rewrote the tests to ES Modules, maybe tap/nyc cannot handle that correctly

More interesting refs: https://github.com/tapjs/node-tap/blob/19181ac81310ae611203fe07a42aeb31b96085b7/docs/src/content/docs/using-with/index.md

gr2m:
How about we turn of coverage for now, get this shipped, then rewrite tests into TypeScript and Jest like the other `@octokit/*` packages, at which point we'd have coverage again, too

oscard0m:
> How about we turn of coverage for now, get this shipped, then rewrite tests into TypeScript and Jest like the other `@octokit/*` packages, at which point we'd have coverage again, too

If it sounds fine to you, sounds good to me too :)

package.json
```diff
@@ -36,10 +35,10 @@
     "build": "pika build",
     "coverage": "tap --coverage-report=html && open coverage/lcov-report/index.html",
     "generate-known-webhooks": "node scripts/update-known-events.js",
-    "lint": "prettier --check 'src/**/*.{js,json}' 'test/**/*' 'scripts/**/*' index.d.ts README.md package.json",
-    "lint:fix": "prettier --write 'src/**/*.{js,json}' 'test/**/*' 'scripts/**/*' index.d.ts README.md package.json",
-    "pretest": "npm run -s lint",
-    "test": "tap --100 --coverage 'test/**/*-test.js'",
+    "lint": "prettier --check 'src/**/*.{ts,json}' 'scripts/**/*' 'test/**/*' README.md package.json",
+    "lint:fix": "prettier --write 'src/**/*.{ts,json}' 'scripts/**/*' 'test/**/*' README.md package.json",
+    "pretest": "npm run -s lint; npm run build -s",
```
gr2m:
```suggestion
    "pretest": "npm run -s lint && npm run build -s",
```

otherwise the script continues, even when `npm run -s lint` fails

gr2m:
`npx tap test/integration/server-test.js` still fails for me locally with some weird error. But if it passes here on CI and for you, let's just ignore it until we rewrote the tests to TypeScript, too, it might just resolve itself

oscard0m:
All checks passing ✅ 
Trying pika release in a dummy repo now

:')

oscard0m:
@gr2m I just did a quick test with IntelliSense and seems to work.
@ankeetmaini if you are free to double-check this pika package would be awesome (`npm install https://github.pika.dev/octokit/webhooks.js/pr/113`)

As soon as we merge this I will create issues for:
- https://github.com/octokit/webhooks.js/pull/113#issuecomment-662096492
- https://github.com/octokit/webhooks.js/pull/113#issuecomment-662088209

---

I think we are ready for final review and go 🚀 

Should this be a  beta release just in case?

gr2m:
I'm not sure if a beta release is worth it. If people want to test it out, installing from the Pika CDN is easy enough.

Let met test against Probot v9 and the upcoming v10

gr2m:
I've found these changes

- `Webhooks.WebhookEvent` is now `EventPayloads.WebhookEvent`
- `Webhooks.WebhookPayload*` are now `EventPayloads.WebhookPayload`

We should make sure to document them in the release notes.

I do wonder if it would be possible to avoid that change? I think an exported class `Webhooks` can be merged with a TypeScript namespace with the same name, so we could keep exporting the `WebhookEvent` as `Webhooks` namespace type? What do you think?

Other than that things look good with Probot 👍🏼

gr2m:
The intellisense is still not working for me:
image

I think that the reason is that in `src/index.ts`, we set the event handler type to just `Function`

```ts
  public on: (
    event: EventNames.AllEventTypes | EventNames.AllEventTypes[],
    handler: Function
  ) => void;
```

The API type seems to be defined in both `src/index.ts` and in `src/generated/api.ts`. I think this should be consolidated?

gr2m:
Hey @dominguezcelada  I'm very sorry to put pressure on this, but this pull request becomes a blocker for https://github.com/probot/probot/pull/1274. I can finish it up if you are busy? I can wait another day or two. Just let me know

oscard0m:
> Hey @dominguezcelada I'm very sorry to put pressure on this, but this pull request becomes a blocker for [probot/probot#1274](https://github.com/probot/probot/pull/1274). I can finish it up if you are busy? I can wait another day or two. Just let me know

Jumping on it right now! I tried yesterday to fix your first comment but I struggled a bit. Let me work on it right now and I'll ping you in case I have issues. Sounds reasonable to you? @gr2m

gr2m:
sounds good thanks

oscard0m:
> The intellisense is still not working for me:
> image
> 
> I think that the reason is that in `src/index.ts`, we set the event handler type to just `Function`
> 
> ```ts
>   public on: (
>     event: EventNames.AllEventTypes | EventNames.AllEventTypes[],
>     handler: Function
>   ) => void;
> ```
> 
> The API type seems to be defined in both `src/index.ts` and in `src/generated/api.ts`. I think this should be consolidated?

Focusing in this bug now. As you pointed, this PR has a kind of duplication of code between `index.ts` and `generated/api.ts` when trying to define `class Webhooks` as a regular Javascript class and declare the type at the same time. Do you have clear how it should be exposed `class Webhooks` from (`/generated/api.ts`) so we can re-use that definition and expose it on `index.ts`? Trying it on my own in the meantime @gr2m

gr2m:
sorry I don't know out of hand. I gotta take care of kids now, I'll be back tomorrow morning pacific time

oscard0m:
> sorry I don't know out of hand. I gotta take care of kids now, I'll be back tomorrow morning pacific time

No problem :)

oscard0m:
> > The intellisense is still not working for me:
> > image
> > I think that the reason is that in `src/index.ts`, we set the event handler type to just `Function`
> > ```ts
> >   public on: (
> >     event: EventNames.AllEventTypes | EventNames.AllEventTypes[],
> >     handler: Function
> >   ) => void;
> > ```
> > 
> > 
> > The API type seems to be defined in both `src/index.ts` and in `src/generated/api.ts`. I think this should be consolidated?
> 
> Focusing in this bug now. As you pointed, this PR has a kind of duplication of code between `index.ts` and `generated/api.ts` when trying to define `class Webhooks` as a regular Javascript class and declare the type at the same time. Do you have clear how it should be exposed `class Webhooks` from (`/generated/api.ts`) so we can re-use that definition and expose it on `index.ts`? Trying it on my own in the meantime @gr2m

I'm having problems trying to solve this issue. Basically I want to avoid duplicated code between the actual JS `class Webhooks` and the declaration type but, when trying it, I lose the reference due to `pika` getting rid of unused imports etc

(Example with the changes proposed: https://github.com/dominguezcelada/webhooks.js/pull/36)

Output:
----
image

-----

Tomorrow morning I will keep trying it but I would need some help to solve this in the proper way if possible 🙏

oscard0m:
> I've found these changes
> 
> * `Webhooks.WebhookEvent` is now `EventPayloads.WebhookEvent`
> * `Webhooks.WebhookPayload*` are now `EventPayloads.WebhookPayload`
> 
> We should make sure to document them in the release notes.
> 
> I do wonder if it would be possible to avoid that change? I think an exported class `Webhooks` can be merged with a TypeScript namespace with the same name, so we could keep exporting the `WebhookEvent` as `Webhooks` namespace type? What do you think?
> 
> Other than that things look good with Probot 👍🏼

I will need some help here too @gr2m (sorry for my inefficiency here solving these issues). I've been trying to play and change a bit trying to follow [this approach](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation). But I've not been able to make it "compile". A PR from this PR solving the issue or explaining in more detail how we should tackle this merge of module declarations would be awesome :)

Tomorrow morning `GMT+2` while I'm working I will try to get some slots to tackle these issues. We are in touch!

oscard0m:
> dominguezcelada#36

@gr2m working back on this issue. Let me know when you are online and available! :)

oscard0m:
> > [dominguezcelada#36](https://github.com/dominguezcelada/webhooks.js/pull/36)
> 
> @gr2m working back on this issue. Let me know when you are online and available! :)

In the meantime, I don't know if it's too 💩 solution but, hardcoding (copy-pasting) from `./genrated/api.ts` the signatures seems to make work the Intellisense. https://github.com/dominguezcelada/webhooks.js/pull/37

I keep checking how to make it in a way we reuse the declaration in the meantime

gr2m:
hey sorry I had an incident I had to take care of, I'm back now. Would you like to pair on this?

src/generated/api.ts
```diff
@@ -123,8 +123,12 @@ export declare class Webhooks {
     event: T | T[],
     callback: (event: GetWebhookPayloadTypeFromEvent<T>) => Promise<void> | void
   ): void;
-  public sign(data: any): string;
-  public verify(eventPayload: any, signature: string): boolean;
+  public sign(secret: string, payload: string | object): string;
+  public verify(
+    secret?: string,
+    eventPayload?: object,
+    signature?: string | string[]
```
gr2m: hmm this one is odd, why would signature be an array of strings?

src/generated/event-payloads.ts
```diff
@@ -862,10 +862,6 @@ export declare module EventPayloads {
     open_issues: number;
     watchers: number;
     default_branch: string;
-    allow_squash_merge?: boolean;
-    allow_merge_commit?: boolean;
-    allow_rebase_merge?: boolean;
-    delete_branch_on_merge?: boolean;
```
gr2m: I think these have been added recently, I think `@octokit/webhook-definitions` might not have been up-to-date when you run it?

oscard0m:
> hey sorry I had an incident I had to take care of, I'm back now. 

Don't be sorry @gr2m , I hope everything is ok

> Would you like to pair on this?

If you are free yes! 😄

gr2m:
https://us04web.zoom.us/j/9569811272?pwd=xIFJcA-m04E

src/generated/event-payloads.ts
oscard0m: Solved on [`c2b71c3` (#113)](https://github.com/octokit/webhooks.js/pull/113/commits/c2b71c3bbbc99ad5aa512248cad14681eca103ce)

gr2m:
I'm looking into this right now: https://github.com/octokit/webhooks.js/pull/113#issuecomment-663297034

> I've found these changes
> 
> - `Webhooks.WebhookEvent` is now `EventPayloads.WebhookEvent`
> - `Webhooks.WebhookPayload*` are now `EventPayloads.WebhookPayload`
> 
> We should make sure to document them in the release notes.
> 
> I do wonder if it would be possible to avoid that change? I think an exported class `Webhooks` can be merged with a TypeScript namespace with the same name, so we could keep exporting the `WebhookEvent` as `Webhooks` namespace type? What do you think?

oscard0m:
Merged the duplicated-webhook-class-declaration solution + ESLint fix. All tests passing. Checking in a dummy octokit project to validate Intellisense 😄 💪 :

image

image

image

image

gr2m:
awesome thank you! I'll need a little time to look into replacing the `EventPayloads.*` export with `Webhooks.*` instead. I should have something by your morning tomorrow. Thanks for all your help on this!

.github/workflows/test.yml
```diff
@@ -10,11 +10,18 @@ name: Test
...
   test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node:
+          - 10
+          - 12
+          - 14
+    name: Node ${{ matrix.node }} 
     steps:
```
oscard0m, line 20: ❤️ ❤️ ❤️ ❤️

gr2m:
looks like the `.flat()` call is not working in Node 10, but we do need to continue to support Node 10:
https://github.com/octokit/webhooks.js/runs/929624325#step:5:51

oscard0m:
> awesome thank you! I'll need a little time to look into replacing the `EventPayloads.*` export with `Webhooks.*` instead. I should have something by your morning tomorrow. Thanks for all your help on this!

Thanks for your time and your patience 🤟 !

tsconfig.json
```diff
@@ -1,10 +1,4 @@
 {
-	"compilerOptions": {
-		"esModuleInterop": true,
-		"module": "esnext",
-		"moduleResolution": "node",
-		"strict": true,
-		"target": "es2020"
-	},
+	"extends": "@tsconfig/node10/tsconfig.json",
```
oscard0m, line 2: nice 👍

gr2m:
I'm giving up on [this](https://github.com/octokit/webhooks.js/pull/113#issuecomment-663297034)

> I've found these changes
> 
> - `Webhooks.WebhookEvent` is now `EventPayloads.WebhookEvent`
> - `Webhooks.WebhookPayload*` are now `EventPayloads.WebhookPayload`
> 
> We should make sure to document them in the release notes.
> 
> I do wonder if it would be possible to avoid that change? I think an exported class `Webhooks` can be merged with a TypeScript namespace with the same name, so we could keep exporting the `WebhookEvent` as `Webhooks` namespace type? What do you think?

I've tried a few things but only hit road blocks. I do think though that we should export `WebhookEvent` separately, not as a type on EventPayloads

gr2m:
I think we should rename `EventNames.AllEventTypes` to `EventNames.All`. And then create `EventNames.AllWithoutActions` or similar, which would only include `check_run`, `issues` etc, but not `check_run.created`, `issues.opened` etc.

Then we can use that type to set the `WebhookEvent.name` property, which was set to `EventNames.AllEventTypes`, but that's incorrect. I set it to just `string` for now

src/event-handler/receive.ts
```diff
@@ -1,11 +1,11 @@
 import { wrapErrorHandler } from "./wrap-error-handler";
 import { WebhookEvent, EventState } from "../types";
-import { EventNames } from "../generated/event-names";
+// import { EventNames } from "../generated/event-names";
 
 function getHooks(
   state: EventState,
   eventPayloadAction: string,
-  eventName: EventNames.AllEventTypes
+  eventName: string // EventNames.AllEventTypes is not correct, only the event names are supported here, not the <event name>.<action> combinations
```
gr2m: see my comment here: https://github.com/octokit/webhooks.js/pull/113#issuecomment-666808970

src/generated/event-payloads.ts
```diff
@@ -4568,13 +4568,4 @@ export declare module EventPayloads {
...
     installation?: WebhookPayloadCheckRunInstallation;
   };
-
-  interface WebhookEvent<T> {
-    id: string;
-    name: string;
-    payload: T;
-    protocol?: "http" | "https";
-    host?: string;
-    url?: string;
```
gr2m, line 4588: no idea where these properties came from. My guess would be Probot. But we definitely don't use these

src/index.ts
```diff
@@ -4,17 +4,13 @@ import { middleware } from "./middleware/middleware";
...
-  WebhookEvent as WebhookEventOptions,
-} from "./types";
+import { EventHandlerOptions, EventState, WebhookEvent } from "./types";
 import { EventNames } from "./generated/types";
 import { GetWebhookPayloadTypeFromEvent } from "./generated/api";
 import { IncomingMessage, ServerResponse } from "http";
 
 class Webhooks {
-  public sign: (secret: string, payload: string | object) => string;
+  public sign: (payload: string | object) => string;
```
gr2m, line 13: this was actually incorrect, the `secret` is set in the internal state of the `webhooks` instance. You only need to set the `secret` parameter when you use the standalone `sign` method

src/middleware/middleware.ts
```diff
@@ -40,14 +40,14 @@ export function middleware(
     });
   }
 
-  const eventName = request.headers["x-github-event"] as WebhookEvent["name"];
-  const signature = request.headers["x-hub-signature"];
-  const id = request.headers["x-github-delivery"];
+  const eventName = request.headers["x-github-event"] as string;
+  const signature = request.headers["x-hub-signature"] as string;
+  const id = request.headers["x-github-delivery"] as string;
```
gr2m, line 46: headers are always strings for us, but Typescript things they can also be `string[]` which makes no sense to me

src/types.ts
```diff
@@ -1,19 +1,16 @@
-import { EventNames } from "./generated/types";
-export interface WebhookEvent {
-  id: string | string[] | undefined;
-  name: EventNames.AllEventTypes;
-  payload: Payload;
-  signature: string | string[] | undefined;
+// import { EventNames } from "./generated/types";
+export interface WebhookEvent<T> {
+  id: string;
+  name: string; // EventNames.AllEventTypes is not correct, only the event names are supported here, not the <event name>.<action> combinations
```
gr2m: see my comment here: https://github.com/octokit/webhooks.js/pull/113#issuecomment-666808970

gr2m:
I'm working on this now

> I think we should rename `EventNames.AllEventTypes` to `EventNames.All`. And then create `EventNames.AllWithoutActions` or similar, which would only include `check_run`, `issues` etc, but not `check_run.created`, `issues.opened` etc.
> 
> Then we can use that type to set the `WebhookEvent.name` property, which was set to `EventNames.AllEventTypes`, but that's incorrect. I set it to just `string` for now

src/generated/api.ts
```diff
@@ -112,30 +103,3 @@ export type GetWebhookPayloadTypeFromEvent<T> = T extends EventNames.ErrorEvent
...
-  public removeListener<T extends EventNames.AllEventTypes>(
-    event: T | T[],
-    callback: (event: GetWebhookPayloadTypeFromEvent<T>) => Promise<void> | void
-  ): void;
-  public middleware(
-    request: IncomingMessage,
-    response: ServerResponse,
-    next?: (err?: any) => void
-  ): void | Promise<void>;
-}
```
gr2m: @dominguezcelada it looks like this was not used internally at all?

gr2m:
I think I'm all done from my side.

The last thing I'm not happy about is that I don't see a way how we can export a `Webhook` namespace that merges with the `Webhook` class, and that has all the payload types imported from the `src/generated` files. I think it could work if we wrote the types directly into `src/index.ts`, but that's not elegant either.

I'm also not super happy about the naming of `EventNames.All` and `EventNames.StringNames`. But unless anyone has a better idea we'll have to live with that :)

oscard0m:
> I think I'm all done from my side.
> 
> The last thing I'm not happy about is that I don't see a way how we can export a `Webhook` namespace that merges with the `Webhook` class, and that has all the payload types imported from the `src/generated` files. I think it could work if we wrote the types directly into `src/index.ts`, but that's not elegant either.

Since we want to keep the generated part under `/generated/*` folder I think we should invest efforts on finding a solution to this which keeps the generation a part and it's consumed as a file...

> I'm also not super happy about the naming of `EventNames.All` and `EventNames.StringNames`. But unless anyone has a better idea we'll have to live with that :)

Not sure neither but... do you like `EventNames.StringNames`?

gr2m:
how about adding a TypeScript section to the README where we can document the exported types

- `WebhookEvent<T>`
- `EventPayloads`
- `EventNames`

And talking about this, I wonder if we can make the type parameter optional in `WebhookEvent<T>`? `WebhookEvent` would be the same as `WebhookEvent<any>`. I think that would make things easier, what do you think?

We can also do the change in a follow up PR, as `WebhookEvent<any>` will still work

gr2m (dismissed):
good to go 👍🏼

oscard0m:
> how about adding a TypeScript section to the README where we can document the exported types
> 
> * `WebhookEvent<T>`
> * `EventPayloads`
> * `EventNames`

Sure, I can draft a README.md entry for this. :)

> And talking about this, I wonder if we can make the type parameter optional in `WebhookEvent<T>`? `WebhookEvent` would be the same as `WebhookEvent<any>`. I think that would make things easier, what do you think?
> 
> We can also do the change in a follow up PR, as `WebhookEvent<any>` will still work

I don't remember which was the reason of this <any> but if it would be  solved with WebhookEvent<T>`? as optional to me has sense.

gr2m:
I've made the type parameter optional for `WebhookEvent` via https://github.com/octokit/webhooks.js/pull/113/commits/3e23a9073d975bf26034b180d79a717f3770ba8e

I like it much better

gr2m (approved):

PR #115, opened 2020-03-18
🐛 fix(type-writer): fix warning Prettier when running types generator

oscard0m:
## 📝 Summary

To upgrade `@gimenete/type-writer` to the latest version

## ⛱ Motivation and Context

Fix Prettier `warning` when generating types:
```bash
 { parser: "babylon" } is deprecated; we now treat it as { parser: "babel" }.
```

## 📊 How Has This Been Tested?

1. `npm run generate-types` using `@gimenete/type-writer v0.1.3` to see the warning
2. upgrade to latest @gimenete/type-writer v0.1.5"
3. validate the warning is not happening anymore

### Before
image

### After
image

## 📚 References:

PR fixing the issue on type-writer: https://github.com/gimenete/type-writer/pull/4
Prettier parser deprecation references:
* https://prettier.io/docs/en/options.html#parser
* https://prettier.io/blog/2019/01/20/1.16.0.html#rename-babylon-parser-to-babel-5647-by-wuweiweiwu

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -153,13 +153,13 @@
       }
     },
     "@gimenete/type-writer": {
-      "version": "0.1.3",
-      "resolved": "https://registry.npmjs.org/@gimenete/type-writer/-/type-writer-0.1.3.tgz",
-      "integrity": "sha512-vhpvVfM/fYqb1aAnkgOvtDKoOgU3ZYIvDnKSDAFSoBvallmGURMlHOE0/VG/gqunUZVXGCFBGHxI8swjBh+sIA==",
+      "version": "0.1.5",
+      "resolved": "https://registry.npmjs.org/@gimenete/type-writer/-/type-writer-0.1.5.tgz",
+      "integrity": "sha512-ZeVljCBixwIU4MG3Kqy/3Dc7zfIMCj9kXyCQ6hiutwFqZkTWggrhqNSEc+NfN2ztLjUG+ssdFe7PUcCIXv67aA==",
       "dev": true,
       "requires": {
         "camelcase": "^5.0.0",
-        "prettier": "^1.13.7"
+        "prettier": "^1.19.1"
       }
     },
     "@nodelib/fs.scandir": {
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "debug": "^4.0.0"
   },
   "devDependencies": {
-    "@gimenete/type-writer": "^0.1.3",
+    "@gimenete/type-writer": "^0.1.5",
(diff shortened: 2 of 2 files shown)


oscard0m:
**[❓ Question ]** 
I'm not sure if here we should leave `dependabot` to be in charge of this or, since it was a thir-party pr opened for this intentionally, we should explicitly open it (as I did) so this way appears on Release Notes and so.

---

Depending on the answer I will open a PR for [here](https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/3b5d16f838f01f7e2d1843928b03e2580538f3f6/package.json) too

gr2m (approved):
Nice, thank you!

gr2m:
to answer your question: dependabot / greenkeeper will only send updates for dependencies that are out of range. I'd appreciate pull requests for the other repositories that depend on `@gimenete/type-writer`

oscard0m:
> to answer your question: dependabot / greenkeeper will only send updates for dependencies that are out of range. I'd appreciate pull requests for the other repositories that depend on `@gimenete/type-writer`

@gr2m 

Interesting. 2 questions come to my mind here:
* Why (or when) are you using greenkeeper an dependabot? Are not both for the same purpose? (upgrading dependencies automatically?)
* What do you mean by dependencies out of range? For instance, `"@gimenete/type-writer": "^0.1.3"`, with `^` notation is containing `v0.1.5`. Is not `dependabot` / `greenkeeper` in charge of updating it automatically?

Thanks! :)

gr2m:
I'm using Greenkeeper right now.

Dependabot is now built into GitHub, it is used to update dependencies in case of security vulnerabilities. If you want to use Dependabot for all dependency updates, you can do it by installing this app: https://github.com/apps/dependabot-preview. I'll probably do that for Octokit soon, as Greenkeeper was aquired by Snyk and will be shut down

oscard0m:
> I'm using Greenkeeper right now.
> 
> 
> 
> Dependabot is now built into GitHub, it is used to update dependencies in case of security vulnerabilities. If you want to use Dependabot for all dependency updates, you can do it by installing this app: https://github.com/apps/dependabot-preview. I'll probably do that for Octokit soon, as Greenkeeper was aquired by Snyk and will be shut down

@gr2m 

Didn’t notice dependabot commits are only for vulnerabilities, hehe. Didn’t know Greenkeeper is shutting down on June 2020! But... what about snyk solution? Seems they plan to do something Open Source right? Also I’ve been working with Renovate bot, just in case you want to evaluate this solution :)

Thanks for your clarifications and help once again!

PR #125, opened 2020-04-16
🚨 test(use-pika): test pointing to built version of code source

oscard0m:
## 📝 Summary

- Pointing to `pkg/dist-src` fails because `../../lib/......` is not possible to resolve from `/pkg/dist-src/` path.

image

---

- Moving `./lib/` to `./src/lib` would be an option but I don't think it's a good practice, right? Pika warns in console about it:

image

---

❓ Any ideas here on how to solve this use case?

Thanks!


gr2m:
How about we export everything from `src/index.js`? So instead of

```js
const EventHandler = require('@octokit/webhooks/event-handler')
```

We'd do 

```js
const { EventHandler } = require('@octokit/webhooks')
```

Then you should be able to do 

```js
const EventHandler = require("../../pkg/");
```

You might need to add a script to package.json to make sure `pkg/` is updated before running tests

```json
  "pretest": "npm run build -s"
```

oscard0m:
> How about we export everything from `src/index.js`? So instead of
> 
> ```js
> const EventHandler = require('@octokit/webhooks/event-handler')
> ```
> 
> We'd do
> 
> ```js
> const { EventHandler } = require('@octokit/webhooks')
> ```
> 
> Then you should be able to do
> 
> ```js
> const EventHandler = require("../../pkg/");
> ```

I think this is a good approach! PRint it! :)

> You might need to add a script to package.json to make sure `pkg/` is updated before running tests
> 
> ```json
>   "pretest": "npm run build -s"
> ```

Yep, you are right. Adding it on the PR too

Thanks!

oscard0m:
@gr2m 

It's a subdependency the one consuming `../../lib`:
https://github.com/octokit/webhooks.js/blob/d675137b32c5867c3fe5f1d204a8075b85cb83d1/src/event-handler/on.js#L3

Since the one exposed on `index.js` would be `EventHandler` which is importing `/on.js` we end up having the same issue https://github.com/octokit/webhooks.js/pull/125/commits/085dc1dc52571b9c0f91cf397b05c784678a55e5:

image

An option could be to import `./lib` from index.js and pass it to EventHandler as parameter of the constructor, but I don't like to conditionate the implementation due to a project structure issue. What do you think? Any ideas?

---
`pkg/dist-src/index.js`

```javascript
const createEventHandler = require("./event-handler");

const middleware = require("./middleware/middleware");

const sign = require("./sign");

const verify = require("./verify");

const verifyAndReceive = require("./middleware/verify-and-receive");

function createWebhooksApi(options) {
    throw new Error("options.secret required");
  }

  const state = {
    eventHandler: createEventHandler(options),
    path: options.path || "/",
    secret: options.secret
  };
  return {
    sign: sign.bind(null, options.secret),
    verify: verify.bind(null, options.secret),
    on: state.eventHandler.on,
    removeListener: state.eventHandler.removeListener,
    receive: state.eventHandler.receive,
    middleware: middleware.bind(null, state),
    verifyAndReceive: verifyAndReceive.bind(null, state)
  };
}

module.exports = {
  createWebhooksApi,
  middleware,
  sign,
  verify,
  verifyAndReceive
};
```

gr2m:
You'll need to do a few things.

1. `*.json` files are not allowed, they cannot be imported with `import ... from "..."`. So instead of creatind `webhook-names.json` we need to create `webhook-names.js` which exports the array
2. Move `lib/webhook-names.json` to `src/webhook-names.js`

I think you also have to write the `src/` files using ES Modules, not common JS. `pika` won't be able to handle the `require` statements if I recall correctly. Sorry I totally forgot, because I usually use TypeScript directly

oscard0m:
> You'll need to do a few things.
> 
> 1. `*.json` files are not allowed, they cannot be imported with `import ... from "..."`. So instead of creatind `webhook-names.json` we need to create `webhook-names.js` which exports the array
> 2. Move `lib/webhook-names.json` to `src/webhook-names.js`
> 
> I think you also have to write the `src/` files using ES Modules, not common JS. `pika` won't be able to handle the `require` statements if I recall correctly. Sorry I totally forgot, because I usually use TypeScript directly

@gr2m 

Applied changes! Let me know what do you think about:
- coverage-map to bypass coverage issue with `tap`
- to keep ./lib/webhook-names.json and ./src/webhook-names.js for retrocompatibility (just in case somebody is consumming `./lib/webhook-names.json`)
- any other feedback or consideration I missed :)

gr2m:
I wouldn't worry about coverage while transitioning. You don't change any logic. I'd remove it and add it back once it's done rewriting into typescript

gr2m:
> to keep ./lib/webhook-names.json and ./src/webhook-names.js for retrocompatibility (just in case somebody is consumming ./lib/webhook-names.json)

if someone used `require("@octokit/webhooks/lib/webhook-names.json")` they used an internal API, due to change at any time. It will stop working after the migration to TypeScript anyway

test/integration/event-handler-test.js
```diff
@@ -1,13 +1,13 @@
 const test = require("tap").test;
 
-const EventHandler = require("../../src/event-handler");
+const createEventHandler = require("../../pkg/dist-src/event-handler");
```
gr2m, line 3: don't use the `dist-src` folder in `pkg`, instead use the export from `require("../../pkg")`.

oscard0m: I think I'm not following you here. When I run `pika build` I get this folder structure generated so it's not possible to apply the path you suggest:

image

In any case, since it's a temporary path until the TypeScript migration is finished, is it important to not use `dist-src`? 

Thanks @gr2m !

gr2m: `pkg/package.json` points to `dist-node/index.js`

oscard0m: > `pkg/package.json` points to `dist-node/index.js`

Sorry but even it's there I'm not able to make it work. I added all the code https://github.com/octokit/webhooks.js/pull/113 alrerady. Since it's not merged yet, maybe pika's package is still not available? Not sure what I'm missing.

@gr2m

gr2m:
I'll have another look at https://github.com/octokit/webhooks.js/pull/113. No worries, we will figure this out :)

PR #127, opened 2020-04-27
refactor: remove lib/ folder

oscard0m:
## 🏗️  Summary

- Move `/lib` files to `src/`
- Generate webhook-names in `src/`
- Change READMEs with new require paths for `middleware`, `event-handler`, `sign`, `verify`
- Move `scripts/` outside of `src/`

## ⛱ Motivation and Context
https://github.com/octokit/webhooks.js/pull/113#issuecomment-620274103


oscard0m:
## ❓  Questions

* What should we get rid of exactly from `lib/`? `/lib/index.js?
* Which path should we use for `index.js` then? Or should we start working with pika built package?

@gr2m

gr2m:
> What should we get rid of exactly from `lib/`? `/lib/index.js?

the entire folder. 

Move logic from `lib/index.js` into `src/index.js` and move `lib/webhook-names.json` to `src/webhook-names.json`. Update `update-known-events.js` accordingly.

> Which path should we use for `index.js` then? Or should we start working with pika built package?

No don't use pika yet, merge the code from lib/index.js into src/index.js, and update package.json's "main" field to point to src/index.js

---

one more thing: please move `src/scripts` back to `scripts`. It's not part of the source code, and I'd prefer to keep that code in JavaScript, like in the other `octokit/*` repositories. It's only used for build scripts

oscard0m:
> > What should we get rid of exactly from `lib/`? `/lib/index.js?
> 
> the entire folder.
> 
> Move logic from `lib/index.js` into `src/index.js` and move `lib/webhook-names.json` to `src/webhook-names.json`. Update `update-known-events.js` accordingly.
> 

Done ✅ 

> > Which path should we use for `index.js` then? Or should we start working with pika built package?
> 
> No don't use pika yet, merge the code from lib/index.js into src/index.js, and update package.json's "main" field to point to src/index.js

Done ✅ 

Apparently we forgot to change `package.json -> main`:
https://github.com/octokit/webhooks.js/blob/7054d1af91d5428ba61d521c3a9d5d823009d107/package.json#L8
 
> one more thing: please move `src/scripts` back to `scripts`. It's not part of the source code, and I'd prefer to keep that code in JavaScript, like in the other `octokit/*` repositories. It's only used for build scripts

Done ✅ 

@gr2m

gr2m (approved):
Great work 👍

PR #134, opened 2020-05-12
Fix intellisense broken for named import

### LINKED ISSUE #133: Intellisense not working when using named import of webhooks
When using default import it works:
image

But when using new named import it does not work:
image

More details [here](https://github.com/octokit/webhooks.js/pull/113#issuecomment-624343436)

oscard0m:
## 🗒️ Summary
- `index.d.ts` migration to `named export` + `export default`
   - Intellisense support*
- Changed signature of Webhooks when using named import:
   - From:  `import { createWebhooksApi } ...` -> `createWebhooksApi(options)`
   - To: `import { Webhooks }` -> `new Webhooks(options)`
- Publish types with `pika` via `@pika/plugin-build-types

Fixes #134 

### Intellisense Support*
|   | default import  | named import  |
|---|---|---|
| Require  | 🔴 | ✅ |
| ES Modules  | ✅ | ✅ |

#### `const Webhooks = require("@octokit/webhooks");` 🔴 
image

Intellisense will not work here. To make it work it needs to use the old `exports = Webhooks` in `index.d.ts`. The problem of using this one is that you can't export anything else, so not compatible with named exports. 

I did not came up with a better solution: https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

#### `const { Webhooks } = require("@octokit/webhooks");` ✅ 
image

#### `import Webhooks from '@octokit/webhooks'` ✅ 
image

#### `import { Webhooks } from "@octokit/webhooks";` ✅ 
image


scripts/generate-types.js
```diff
@@ -20,8 +26,7 @@ webhooks.forEach(({ name, actions, examples }) => {
       [`${typeName}.repository`]: "PayloadRepository",
       // This prevents a naming colision between the payload of a `installation_repositories` event
       // and the `repositories` attribute of a `installation` event
-      "WebhookPayloadInstallation.repositories":
-        "WebhookPayloadInstallation_Repositories",
```
wolfy1339: Please revert these changes

test/integration/deprecations-test.js
```diff
@@ -3,8 +3,10 @@ const { mock } = require("simple-mock");
 
 test("@octokit/webhooks", (t) => {
   const emitWarningMock = mock(process, "emitWarning");
-  const createWebhooksApi = require("../../src");
```
wolfy1339, line 6: Please revert these changes

wolfy1339 (dismissed):
Please revert the changes made to the styling of the JS files

test/integration/deprecations-test.js
wolfy1339: It wasn't meant for the new exporting format. Just the style/formatting of the file itself

oscard0m: > It wasn't meant for the new exporting format. Just the style/formatting of the file itself

Ah ok. But still not clear what should I restore. Running `lint` I don't get any error or warning.

wolfy1339: It fixes any issues automatically. It's fixed in your later commits

gr2m (changes requested):
I agree that exporting `{ Webhooks }` instead of `{ createWebhooksApi }` is the better API. When we rewrite it to TypeScript, we should make it a proper class.

But that change is also a breaking change, unfortunately. We need to export `{ createWebhooksApi, Webhooks }` and log a deprecation message for `createWebhooksApi`. I know it's a pain in the 🍑, but that's our policy: every API must be deprecated before removing them in the next breaking version

oscard0m:
> I agree that exporting `{ Webhooks }` instead of `{ createWebhooksApi }` is the better API. When we rewrite it to TypeScript, we should make it a proper class.

👍 

> But that change is also a breaking change, unfortunately. We need to export `{ createWebhooksApi, Webhooks }` and log a deprecation message for `createWebhooksApi`. I know it's a pain in the 🍑, but that's our policy: every API must be deprecated before removing them in the next breaking version

👍 No problem ;)

README.md
```diff
@@ -61,18 +61,23 @@ source.onmessage = (event) => {
...
+    - [webhooks.verifyAndReceive()](#webhooksverifyandreceive)
+    - [webhooks.receive()](#webhooksreceive)
+    - [webhooks.on()](#webhookson)
+    - [webhooks.removeListener()](#webhooksremovelistener)
+    - [webhooks.middleware()](#webhooksmiddleware)
+    - [Webhook events](#webhook-events)
+    - [Special events](#special-events)
+      - [`*` wildcard event](#wildcard-event)
+      - [`error` event](#error-event)
+  - [License](#license)
```
gr2m: Not sure what happened here? BUt could you please revert the change?

gr2m (changes requested):

gr2m:
Other than the README toc, this PR is good to go :)

README.md
oscard0m: > Not sure what happened here? BUt could you please revert the change?

It was a VS Code Markdown extension, I did not even notice! Fixed ;)

gr2m (approved):
🚀

PR #161, opened 2020-07-12
fix(types): Update "sign" and "verify" types in generate-types.js

jakemwood:
At some point, `sign` and `verify` were changed to require the `secret` parameter, but the accompanying TypeScript types were not updated. This PR updates those hard-coded function types, and running `npm run generate-types` seems to make some other updates to `index.d.ts` as well that I have left in this PR.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,31 +12,6 @@ type Options = {
 };
 
 declare namespace Webhooks {
-  type WebhookPayloadPingPayloadHookConfig = {
-    content_type: string;
-    insecure_ssl: string;
-    url: string;
-  };
-  type WebhookPayloadPingPayloadHook = {
-    type: string;
-    id: number;
-    name: string;
-    active: boolean;
-    events: Array<string>;
-    config: WebhookPayloadPingPayloadHookConfig;
-    updated_at: string;
-    created_at: string;
-    app_id: number;
-  };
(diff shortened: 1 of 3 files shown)


gr2m:
This might be fixed via #113 @dominguezcelada

oscard0m:
> This might be fixed via #113 @dominguezcelada

On current status of this PR the fix is not applied:

https://github.com/dominguezcelada/webhooks.js/blob/9b60c38b5bf278abbb2b96c40547169c0a2e962c/scripts/generate-types.js#L89-L90

But I agree with @jakemwood we should update generated-types to be compliant with the signatures:

https://github.com/dominguezcelada/webhooks.js/blob/9b60c38b5bf278abbb2b96c40547169c0a2e962c/src/sign/index.ts#L3

https://github.com/dominguezcelada/webhooks.js/blob/9b60c38b5bf278abbb2b96c40547169c0a2e962c/src/verify/index.ts#L5-L10

---

Maybe we should move forward with this PR and adopt these changes to https://github.com/octokit/webhooks.js/pull/113 . What do you think @gr2m 

P.S.: Thanks for reporting @jakemwood :)

gr2m:
@dominguezcelada if it works for you, let's do that 👍

gr2m (approved):

PR #168, opened 2020-07-29
feat: "workflow_dispatch" event and new actions for "pull_request", "release", and "sponsorship"

heuels:
Hey 🙂 

Following octokit/webhooks#134 and octokit/webhooks#135 here is an update for types and events.

Also, I think it would be nice if on each update of `octokit/webhooks` events and types regenerated automatically using GitHub Actions. Is it something that you would be interested in? I could create a PR for that as well.

-----
[View rendered README.md](https://github.com/heuels/webhooks.js/blob/workflow_dispatch/README.md)


gr2m (approved):
Looks good, thank you!

gr2m:
These should have been updated automatically via https://github.com/octokit/webhooks.js/blob/master/.github/workflows/update.yml and these PRs: #167 #168 but that didn't work for some reason, I'll investigate

PR #183, opened 2020-08-04
feat: set common error types from Octokit request errors

### LINKED ISSUE #180: [TypeScript] `error` argument in `.on("error")` callback should permit for additional properties
This is currently not possible

```js
webhooks.on("error", (error) => {
  const { status, headers, request } = error;
});
```

Also the `error` type is merged with the returnType of the `transform` option, which was introduced in #177. 

I think we should remove the special events `"error"` and `"*"`, and instead introduce additional methods, to make clear that these are special events, such as

- `webhooks.onError()`
- `webhooks.onAny()`

Any thoughts?

### LINKED ISSUE #182: Make the error message more helpful
When an error occurs in an event handler, we set the error message to just "Webhook handler error"

https://github.com/octokit/webhooks.js/blob/89dca081f7e3d46020ec1d89fc2b1366d7579df5/src/event-handler/receive.ts#L76

We should instead set it to the first error message. And if there are more than 1 error, we should append the message with sth like (+2 more errors)

We could also consider using https://github.com/sindresorhus/aggregate-error#readme, it's a great package, we use it in semantic-release

gr2m:
fixes #180
fixes #182

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,18 +24,18 @@
       }
     },
     "@babel/core": {
-      "version": "7.10.5",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.10.5.tgz",
-      "integrity": "sha512-O34LQooYVDXPl7QWCdW9p4NR+QlzOr7xShPPJz8GsuCU3/8ua/wqTr7gmnxXv+WBESiGU/G5s16i6tUvHkNb+w==",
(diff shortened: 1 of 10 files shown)


gr2m (commented):
tested against latest probot beta (https://github.com/probot/probot/pull/1274), this should be good to go. Happy to do follow up PRs if there is any post-merge feedback

PR #190, opened 2020-08-06
feat: reply after 9s with 202 status

### LINKED ISSUE #185: Make sure to respond within 10s in `.middleware`
> GitHub expects that integrations respond within 10 seconds of receiving the webhook payload. If your service takes longer than that to complete, then GitHub terminates the connection and the payload is lost.

https://docs.github.com/en/rest/guides/best-practices-for-integrators#favor-asynchronous-work-over-synchronous

We currently await the handling of all event handlers.

I think what we should do instead is to start a timer for 9s. If it runs out before the event handlers are done, respond with a [`202` status](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202)

The change would need to be implemented here:
https://github.com/octokit/webhooks.js/blob/aee3e94ae9450bee88c78eb073696a176913a542/src/middleware/middleware.ts#L48-L67

gr2m:
We can test this with https://github.com/sinonjs/fake-timers, without needing to wait for 10s in our tests.

If anynone wants to finish this today be my guest, if not I'll finish this up tomorrow

fixes #185

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2496,23 +2496,29 @@
       "dev": true
     },
     "compare-func": {
-      "version": "1.3.4",
-      "resolved": "https://registry.npmjs.org/compare-func/-/compare-func-1.3.4.tgz",
-      "integrity": "sha512-sq2sWtrqKPkEXAC8tEJA1+BqAH9GbFkGBtUOqrUX57VSfwp8xyktctk+uLoRy5eccTdxzDcVIztlYDpKs3Jv1Q==",
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/compare-func/-/compare-func-2.0.0.tgz",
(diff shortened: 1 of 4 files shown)


oscard0m:
This PR fixes #185 right?

gr2m:
> This PR fixes #185 right?

Yes 👍🏼 

> if not I'll finish this up tomorrow

I think it makes sense to finish up the rewrite of the tests to Jest & TypeScript first :)

gr2m:
I'll have to finish this on Friday latest. It might create some merge conflicts with your work on rewriting tests to TypeScript I'm afraid, but I need to finish up Probot v10 and this will become a blocker by then

oscard0m:
> I'll have to finish this on Friday latest. It might create some merge conflicts with your work on rewriting tests to TypeScript I'm afraid, but I need to finish up Probot v10 and this will become a blocker by then

From my side it's fine. I will be with no laptop for the next to weeks. I would ask to @wolfy1339

PR #192, opened 2020-08-07
fix(typescript): Change options to match the ones from other octokit repos

wolfy1339:
Pika complains of the values, but that is for another day

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,9 +1,10 @@
 {
-  "extends": "@tsconfig/node10/tsconfig.json",
   "compilerOptions": {
-    // required for aggregate-error
     "esModuleInterop": true,
-    "moduleResolution": "node"
+    "module": "esnext",
+    "moduleResolution": "node",
+    "strict": true,
+    "target": "es2018"
   },
   "include": ["src/**/*"]
 }


tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
-  "extends": "@tsconfig/node10/tsconfig.json",
   "compilerOptions": {
-    // required for aggregate-error
     "esModuleInterop": true,
-    "moduleResolution": "node"
+    "module": "esnext",
+    "moduleResolution": "node",
+    "strict": true,
+    "target": "es2018"
```
gr2m, line 7: I was thinking of migrating all the Octokit repos to use `"extends": "@tsconfig/node10/tsconfig.json` but didn't investigate if that's actually a good idea, given that these target both Node and Browsers. So maybe just keeping them in sync with the settings that work is the better idea

gr2m (approved):

tsconfig.json
oscard0m: Sorry, I arrive late to the discussion here. Maybe I can open a different issue for this. What if we keep extending the @tsconfig/node10 and overwrite the prop/s giving problems with pika instead? Checking the preset seems like most of the props added are the same ones [the preset uses](https://github.com/tsconfig/bases/blob/08bc28bc677531473bc6a7928ce981f506fb5dce/bases/node10.json#L5-L13).

I guess the ones giving problems are these ones?

```json
    "module": "commonjs 🆚 esnext", 
    "skipLibCheck": true 🆚 false,
    "forceConsistentCasingInFileNames": true 🆚 false
```

What do you think @gr2m and @wolfy1339

wolfy1339:
Pika also now recommends setting it to ES2020 as the build target instead of ES2018 (which was recommended when we first started using pika).

The `forceConsistentCasingInFileNames` seems like a good option to enable across the organization.

As for `skipLibCheck`, I'm not sure for the reasoning as to why it's set to false

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

gr2m:
I think I run into trouble with the `ES2020` setting and Node 10 which we still support.

> forceConsistentCasingInFileNames

This was never a problem as far as I can remember. It's helpful in theory and a hard-to-debug problem when you run into it, but I also like simple config files

> skipLibCheck

We do not set this right now do we?

oscard0m:
> I think I run into trouble with the `ES2020` setting and Node 10 which we still support.

I would suggest to keep the target for retrocompatibility to Node 10 and we can open an issue a part to discuss this in the future. 

> > forceConsistentCasingInFileNames
> 
> This was never a problem as far as I can remember. It's helpful in theory and a hard-to-debug problem when you run into it, but I also like simple config files

Same here, until we don't face the problem let's keep a minimal `tsconfig`.

> > skipLibCheck
> 
> We do not set this right now do we?

When extending `@typescript/node10` config [we were using `"skipLibCheck": true`](https://github.com/tsconfig/bases/blob/08bc28bc677531473bc6a7928ce981f506fb5dce/bases/node10.json#L12) and when not extending it [the default value is `"skipLibCheck": false`](url) so here... which one should we use?

wolfy1339:
That value doesn't affect the final output version. It goes through multiple steps.

First it transpires it 1 to 1, then makes a version for node, then a version optimised for th browser.

According to `@pika/plugin-build-node`'s docs, it compiles a version for Node 6+.

oscard0m:
> That value doesn't affect the final output version. It goes through multiple steps. First it transpires it 1 to 1, then makes a version for node, then a version optimised for th browser. According to `@pika/plugin-build-node`'s docs, it compiles a version for Node 6+.

Checking the code, seems to be pointing [to Node 8 by default](https://github.com/pikapkg/builders/blob/0d5cf19f5735e85c7286368771844918134b1262/packages/plugin-build-node/src/index.ts#L13)

Reading the README.md it is supposed to point to Node LTS, so I guess [we are safe](https://nodejs.org/en/about/releases/) here. Since now we have Node10 + Node12 + Node14 checks, are these PR checks safe enough to assure it's safe to go? @gr2m @wolfy1339 https://github.com/octokit/webhooks.js/pull/200

P.S.: @wolfy1339 where did you find the Node 6+?

PR #188, opened 2020-08-06
chore(tests): Convert to typescript and use Jest

wolfy1339:
Fixes #187

PR #193, opened 2020-08-07
fix(typescript): Remove file extension from import

wolfy1339:
Found this little bug when doing #188


oscard0m (approved):

gr2m:
I’d have considered this a refactor, not a fix. But no harm done :)

PR #199, opened 2020-08-08
WIP feat: set `error.rawPayload` on "signature does not match event payload and secret" errors

gr2m:
see https://github.com/probot/probot/pull/1274#issuecomment-670962968


gr2m:
do not merge, I will install it directly into the latest probot beta for debugging, using `https://github.pika.dev/octokit/webhooks.js/pr/199`

PR #201, opened 2020-08-09
feat(middleware): use console-log-level instead of debug package

### LINKED ISSUE #186: Replace "debug" with "log" constructor option
We don't use the `debug` package in any other Octokit module. Instead we use a `console`-compatible `log` API, and users can pass their own `log` object, see https://github.com/octokit/core.js#logging

That's what we should do for `@octokit/webhooks`, too.

I would consider removing `debug` a breaking change, but we can introduce the `log` option earlier

oscard0m:
## 📝 Summary

- To use `console-log-level` package to run logs (for consistency for the rest of the repos of Octokit)

## ⛱ Motivation and Context

Fix #186

### DIFF
--- a/src/middleware/middleware.ts
+++ b/src/middleware/middleware.ts
@@ -4,11 +4,15 @@ import { getPayload } from "./get-payload";
 import { verifyAndReceive } from "./verify-and-receive";
 import { debug } from "debug";
 import { IncomingMessage, ServerResponse } from "http";
-import { State, OctokitError, WebhookEventHandlerError } from "../types";
+import { State, WebhookEventHandlerError } from "../types";
 import { EventNames } from "../generated/event-names";
-import AggregateError from "aggregate-error";
+import consoleLogLevel from "console-log-level";
 
 const debugWebhooks = debug("webhooks:receiver");
+const logWebhooks = consoleLogLevel({
+  prefix: "webhooks:receiver",
+  level: "info",
+});
 
 export function middleware(
   state: State,
@@ -27,6 +31,7 @@ export function middleware(
     }
 
     debugWebhooks(`ignored: ${request.method} ${request.url}`);
+    logWebhooks.info(`ignored: ${request.method} ${request.url}`);
     response.statusCode = 404;
     response.end("Not found");
     return;
@@ -47,6 +52,7 @@ export function middleware(
   const id = request.headers["x-github-delivery"] as string;
 
   debugWebhooks(`${eventName} event received (id: ${id})`);
+  logWebhooks.info(`${eventName} event received (id: ${id})`);
 
   return getPayload(request)
     .then((payload: any) => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1573,6 +1573,12 @@
         "@types/node": "*"
       }
     },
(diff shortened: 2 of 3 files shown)


src/middleware/middleware.ts
```diff
@@ -4,11 +4,15 @@ import { getPayload } from "./get-payload";
 import { verifyAndReceive } from "./verify-and-receive";
 import { debug } from "debug";
```
oscard0m, line 5: Keeping `debug` package for retrocompatibility and creating issue to remove it in a different breaking change

src/middleware/middleware.ts
```diff
@@ -4,11 +4,15 @@ import { getPayload } from "./get-payload";
...
 import { IncomingMessage, ServerResponse } from "http";
-import { State, OctokitError, WebhookEventHandlerError } from "../types";
+import { State, WebhookEventHandlerError } from "../types";
 import { EventNames } from "../generated/event-names";
-import AggregateError from "aggregate-error";
```
oscard0m, line 9: Removed unused imports

src/middleware/middleware.ts
```diff
@@ -4,11 +4,15 @@ import { getPayload } from "./get-payload";
...
+import consoleLogLevel from "console-log-level";
 
 const debugWebhooks = debug("webhooks:receiver");
+const logWebhooks = consoleLogLevel({
+  prefix: "webhooks:receiver",
+  level: "info",
+});
```
oscard0m, line 15: Not sure if this is the approach you were looking for @gr2m , when you have time I would like to hear your feedback

gr2m (changes requested):
We don't have `console-log-level` as a dependency in any `@octokit/*` package. Instead we have a `{ log }` option that by default is just set to `console`. You can override `log` with your own, console-compatible object, such as what `console-log-level` returns.

So basically in the `Webhooks` constructor, you would introduce a new `log` option. If it's not set, set it to `console`. Do the same for the other modules that are exported. Keep the `log` on the internal state, then do `state.log.info` etc in the code

Does that all make sense?

oscard0m:
@jablko Does https://github.com/octokit/webhooks.js/pull/457 fix #186? I would like to close this PR if so.

src/middleware/middleware.ts
gr2m: I would not use it, it increases the bundle size, and I want `@octokit/webhooks` to work well in contraint environments such as browsers and Cloudflare Workers in future.

We can recommend to use it to users as we do in `@octokit/core`: https://github.com/octokit/core.js#logging

oscard0m: Does this issue satisfy your request? If so, can we close this PR and associated issue? https://github.com/octokit/webhooks.js/issues/476

gr2m:
> @jablko Does #457 fix #186? I would like to close this PR if so.

We still use `debug`, removing it is a breaking change. So I would leave issue open, or create a follow up issue as a reminder.

What I sometimes do is to use create a milestone for the next breaking change, then assign all relevant issues to that milestone so we don't forget anything. Ideally we would have removed `debug` as part of v8, but I forgot about it, sorry

PR #202, opened 2020-08-13
fix: set encoding before reading payload for signature verification

nogs:
Attempting to fix the signature mismatch errors seen by @gr2m in https://github.com/probot/probot/pull/1274.

When changed from javascript to typescript it looks like the body parsing of the IncomingMessage changed to a string and this may be causing encoding issues when generating the signatures to confirm authenticity of the webhook.

### DIFF
--- a/src/middleware/get-payload.ts
+++ b/src/middleware/get-payload.ts
@@ -12,6 +12,7 @@ export function getPayload(request: IncomingMessage): Promise<any> {
   return new Promise((resolve, reject) => {
     let data = "";
 
+    request.setEncoding("utf8");
     request.on("error", (error) => reject(new AggregateError([error])));
     request.on("data", (chunk) => (data += chunk));
     request.on("end", () => {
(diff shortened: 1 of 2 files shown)


gr2m:
Any chance we could add a test for that, so we don't break it again in future?

nogs:
Will have a look at test, had to fix a couple because of the extra call to setEncoding.  Just wanted to get something out here while I think about it.  

Again, going to be fun to generate some "bad" data I think. I will try and get some test data that breaks it if I can (but as you know it wasn't really consistent behaviour).

nogs:
@gr2m Not sure the best way to test this....

I've used tcpdump to capture network packets this afternoon to see if GitHub are sending a different message and, assuming the issue body is the bit causing the problems, the byte data coming across from GitHub is the same between the mismatch delivery and the working delivery, IE, nodejs is receiving the same data but not consistently handling it...

Not sure how to test something that we can't reproduce.  I'm not sure there is a way to provide a "bad" string from the mock for `getPayload` to fail on in that part, especially when I'm not sure what byte sequences are causing problems.

gr2m:
all right, I tried to create a test to reproduce a problem without the `request.setEncoding("utf8");` line, but no luck. I'll merge this in as you tested it (https://github.com/probot/probot/pull/1274#issuecomment-673375327). I'll make a follow up issue, maybe someone else will have an idea on how to cover this case with a test, thanks again!

PR #234, opened 2020-08-25
test(sign-test): restore 100% coverage for sign-test

oscard0m:


src/sign/index.ts
```diff
@@ -1,6 +1,6 @@
 import { createHmac } from "crypto";
 
-export function sign(secret: string, payload: string | object): string {
+export function sign(secret?: string, payload?: string | object): string {
```
gr2m, line 3: can you revert this change? Both these arguments really are required, and we don't want to suggest otherwise.

PR #238, opened 2020-08-25
test(server-test): refactor mock setTimeout using Jest's utility

### LINKED ISSUE #217: replace "@sinonjs/fake-timers" with Jest timers
follow up for https://github.com/octokit/webhooks.js/pull/188

I've introduced `@sinonjs/fake-timers` in https://github.com/octokit/webhooks.js/pull/190, before the migration to Jest. As Jest has its own solution for mocking time, we should use that instead of `@sinonjs/fake-timers`

oscard0m:
## 📝 Summary

Use Jest's fake timeout utility instead of `sinonjs/fake-timer` one

## ⛱ Motivation and Context

Fix #217


gr2m (approved):
Looks great, thanks!

PR #247, opened 2020-08-30
Add types for `merged_by` property

sorenlouv:


gr2m (commented):
We cannot update that file manually, it's generated from https://github.com/octokit/webhooks. You can follow and collaborate on https://github.com/octokit/webhooks/issues/136 to for better event payload types

PR #258, opened 2020-09-12
Refactor the types and DRY it out

### LINKED ISSUE #181: [maintainance] Unreachable code?
I'm fairly sure that `Array.isArray(eventName)` in this code can never be true

https://github.com/octokit/webhooks.js/blob/89dca081f7e3d46020ec1d89fc2b1366d7579df5/src/event-handler/receive.ts#L12-L16

My guess is that this line is a relic from a previous version. We do cover this line, but only through unit tests.

The `.receive()` method as exported here:

https://github.com/octokit/webhooks.js/blob/89dca081f7e3d46020ec1d89fc2b1366d7579df5/src/event-handler/index.ts#L18

will always be an event object, which name property can never be an array, as these unit tests suggest

https://github.com/octokit/webhooks.js/blob/89dca081f7e3d46020ec1d89fc2b1366d7579df5/test/unit/event-handler-receive-test.js#L30-L47

Or am i missing something?

### LINKED ISSUE #250: [Proposal] [TypeScript] prefer interface to conditional types
```ts
interface PayloadMap {
  'check_run': WebhookPayloadCheckRun
  'check_run.completed': WebhookPayloadCheckRun
  ...
}

type AllEvents = keyof PayloadMap | 'error' | '*'

export declare type GetWebhookPayloadTypeFromEvent<E = AllEvents, T = WebhookEvent> =
  E extends 'error' ? WebhookEventHandlerError :
  E extends '*' ? any :
  WebhookEvent<PayloadMap[E]> & T
```

This will not only help reduce dist size but also optimize type check performance.

ankeetmaini:
Fixes #250 
Fixes #181


@types/aggregate-error/index.d.ts
```diff
@@ -0,0 +1,53 @@
...
+	//=> [Error: foo]
+	//=> [Error: bar]
+	//=> [Error: baz]
+	```
+	*/
+    constructor(errors: ReadonlyArray<T | { [key: string]: any } | string>);
+
+    [Symbol.iterator](): IterableIterator<T>;
+  }
+}
```
gr2m, line 53: could you move this change into a separate PR?

src/middleware/verify-and-receive.ts
```diff
@@ -7,15 +7,14 @@ import {
 
 export function verifyAndReceive<T extends All>(
   state: State<T>,
-  event: EventTypesPayload[T] & { signature: string }
+  event: EventTypesPayload[T] & { signature?: string }
```
gr2m, line 10: why making signature optional here? I don't think it makes sense to call `.verifyAndReceive()` but not have a signature set?

@types/aggregate-error/index.d.ts
gr2m: I think we should also move it into `src/@types`

gr2m: On 2nd thought, it looks like the whole PR is mostly focused on refactoring the usage of AggregateError and not on what is described in https://github.com/octokit/webhooks.js/issues/250 or  #181. Maybe update the title and body of the pull request, and let's focus on that first?

ankeetmaini: Regarding the changes mentioned in the issues (#250) I've removed the huge conditional expression in favour of an object mapping.

I was going through the codebase and at some point found that everything was `any`. So I'm trying to find the best way to model the types.

src/middleware/verify-and-receive.ts
ankeetmaini: I agree, I was just trying out. This is still a WIP. Will remove this :)

ankeetmaini:
Closing this in favour of #292. I don't know what I was trying to do here. LOL

PR #270, opened 2020-09-18
fix(typescript): workaround for trouble with aggregate-error type exports (#245)

### LINKED ISSUE #245: [TypeScript] v7.11.2 has a hard requirement on specific TS configs

gr2m:
resolves #245
resolves #270

 thanks @andrewbranch /cc @dominguezcelada

### DIFF
--- a/src/middleware/middleware.ts
+++ b/src/middleware/middleware.ts
@@ -7,9 +7,6 @@ import { IncomingMessage, ServerResponse } from "http";
 import { State, WebhookError, WebhookEventHandlerError } from "../types";
 import { EventNames } from "../generated/event-names";
 
-// @ts-ignore to address #245
-import AggregateError from "aggregate-error";
-
 const debugWebhooks = debug("webhooks:receiver");
 
 export function middleware(
--- a/src/middleware/middleware.ts
+++ b/src/middleware/middleware.ts
@@ -6,6 +6,8 @@ import { debug } from "debug";
 import { IncomingMessage, ServerResponse } from "http";
 import { State, WebhookError, WebhookEventHandlerError } from "../types";
 import { EventNames } from "../generated/event-names";
+
+// @ts-ignore to address #245
 import AggregateError from "aggregate-error";
 
 const debugWebhooks = debug("webhooks:receiver");
--- a/src/types.ts
+++ b/src/types.ts
@@ -47,7 +47,8 @@ export interface WebhookEventHandlerError extends AggregateError<WebhookError> {
   errors: WebhookError[];
 }
 
-declare class AggregateError<T extends Error = Error> extends Error
+declare class AggregateError<T extends Error = Error>
+  extends Error
   implements Iterable<T> {
   readonly name: "AggregateError";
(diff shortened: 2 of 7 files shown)


src/middleware/middleware.ts
```diff
@@ -6,6 +6,8 @@ import { debug } from "debug";
 import { IncomingMessage, ServerResponse } from "http";
 import { State, WebhookError, WebhookEventHandlerError } from "../types";
 import { EventNames } from "../generated/event-names";
+
+// @ts-ignore to address #245
 import AggregateError from "aggregate-error";
```
andrewbranch: Is this import actually used in this file?

andrewbranch (dismissed):
You’ll want to try a pika build and check that the dist output looks like you expect, but I think this should work.

gr2m:
> You’ll want to try a pika build and check that the dist output looks like you expect, but I think this should work.

I use `pika-ci` to generate previews of the to-be-published assets. It can be downloaded here:
https://github.pika.dev/octokit/webhooks.js/pr/270. It all looks good as far as I can tell?

src/middleware/middleware.ts
gr2m: huh, good catch :)

gr2m: resolved via 034ae6e

gr2m:
@dominguezcelada are you happy with these changes?

@robcresswell can you please check if that resolves https://github.com/octokit/webhooks.js/issues/245 for you? To test this PR, you can `npm install https://github.pika.dev/octokit/webhooks.js/pr/270`

andrewbranch:
Tested by throwing that bundle into my node_modules. Looks good 🌟

oscard0m:
Tested in a local repo I was playing with. Running `tsc` succeeds now.

I created a new issue to make sure we do a followup on this (since the existing one will be closed when we merge this): https://github.com/octokit/webhooks.js/issues/271

src/types.ts
```diff
@@ -47,3 +46,12 @@ export interface WebhookEventHandlerError extends AggregateError<WebhookError> {
    */
   errors: WebhookError[];
 }
+
+declare class AggregateError<T extends Error = Error>
```
oscard0m, line 50: Even though is a temporary fix, we should mention and give credit to the original repo? https://github.com/octokit/webhooks.js/pull/248#issuecomment-695092969 @gr2m

oscard0m (commented):
Looks good to me. Only check small comment

src/types.ts
gr2m: oopsies sorry I missed your comment. You are right. I'll send a follow up pR

PR #283, opened 2020-09-30
test(ts-validate): event prop in WebhookError is a deprecated prop

oscard0m:
## 📝 Summary

To remove `event` property from `WebhookError` in test

## ⛱ Motivation and Context

Property `event` in `WebhookError` is deprecated


gr2m:
I think in general we should continue to test APIs that are deprecated until we remove them in a breaking release, in order to prevent regressions for users who didn't migrate to the new APIs. I like to create a separate `deprecations.test.ts` file to keep all tests for deprecated APIs in a single place. Maybe in this case we could add a new section in `test/typescript-validate.ts` where we put all the APIs that are deprecated?

gr2m (approved):
Looks good 👍🏼

PR #284, opened 2020-09-30
refactor(generate-types): add missing DO-NOT-EDIT comment to remaining generated type files

oscard0m:
## 📝 Summary

Add missing `//DO NOT EDIT` comment to remaining generated type files


scripts/generate-types.js
```diff
@@ -62,6 +62,8 @@ webhooks.forEach(({ name, actions, examples }) => {
 conditionalType.push("never");
 
 const getWebhookPayloadTypeFromEvent = `
+${doNotEditThisFileDisclaimer}
```
gr2m, line 65: I think `doNotEditThisFileDisclaimer` needs to be defined in this file for it to work?

scripts/update-known-events.js
```diff
@@ -3,6 +3,9 @@ const Table = require("table-builder");
 const WEBOOOKS = require("@octokit/webhooks-definitions/index.json");
 const { generateFile } = require("./generate-file");
 
+const doNotEditThisFileDisclaimer = `
+// DO NOT EDIT THIS FILE DIRECTLY
```
gr2m: I think I would slightly rephrase it
```suggestion
// THIS FILE IS GENERATED - DO NOT EDIT DIRECTLY
```

gr2m (commented):
can you generated the `src/generated/*` files so they get updated as part of this PR?

oscard0m:
> can you generated the `src/generated/*` files so they get updated as part of this PR?

sure!

scripts/generate-types.js
oscard0m: > I think `doNotEditThisFileDisclaimer` needs to be defined in this file for it to work?

It was already defined. Only missing on some generated files:

https://github.com/octokit/webhooks.js/blob/96c33db3e5a47d0a221afe7d8eca32c6dd58144a/scripts/generate-types.js#L9-L11

gr2m (approved):
Great!

PR #286, opened 2020-10-01
refactor(types.ts): allow handler functions to return any

### LINKED ISSUE #178: [TypeScript] do not restrict return/resolve type of `.on()` callback
I don't see a reason why we should restrict what the `.on(event, callback)` callback function returns. People use early returns or want to return promises in non-async callback functions.

This code currently emits TypeScript errors

```js
webhooks.on("issues", () => {
  return 1;
});
webhooks.on("issues", () => {
  return Promise.resolve(1);
});
webhooks.on("issues", async () => {
  return "foo";
});
```

I don't think it should. We should allow for any return type, both synchronous and asynchronous

copperwall:
Hey @gr2m 👋 , thought I'd give a typescript PR a shot.

Currently handler functions that are passed to `Webhooks.on` can only
return `void` or `Promise<void>`. The return values from these functions
are only used to wait for all handlers for a given webhook event to
finish before the promise returned from `Webhooks.receive` resolves (or
rejects if any of the handler functions error out).

This updates the type signature to allow handler functions to return
anything.

Since this type signature was used in both `Webhooks.on` and
`Webhooks.removeListener`, this PR adds named type called
`HandlerFunction<E, T>` in `types.ts` and adds test cases to
`typescript-validate.ts` for the more lax types.

Fixes #178

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,9 +4,14 @@ import { middleware } from "./middleware/middleware";
 import { sign } from "./sign/index";
 import { verify } from "./verify/index";
 import { verifyAndReceive } from "./middleware/verify-and-receive";
-import { Options, State, WebhookEvent, WebhookError } from "./types";
+import {
+  Options,
+  State,
+  WebhookEvent,
+  WebhookError,
+  HandlerFunction,
+} from "./types";
 import { EventNames } from "./generated/event-names";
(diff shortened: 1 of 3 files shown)


gr2m (approved):

gr2m:
Thanks Chris!

PR #290, opened 2020-10-01
build(tsconfig): use @octokit/tsconfig preset in local tsconfig.json

### LINKED ISSUE #289: To create `octokit/tsconfig` and consume it from local tsconfig.json of this repository
_Originally posted by @gr2m in https://github.com/octokit/webhooks.js/pull/288#issuecomment-702370784_

oscard0m:
## 📝 Summary

Replace local `tsconfig.json` configuration by the one inherited from `@octokit/tsconfig`

## ⛱ Motivation and Context

Adopt `@octokit/tsconfig` in all repos using TS
Fix #289

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   },
   "devDependencies": {
     "@gimenete/type-writer": "^0.1.5",
-    "@octokit/tsconfig": "^1.0.0",
+    "@octokit/tsconfig": "^1.0.1",
     "@octokit/webhooks-definitions": "3.13.2",
     "@pika/pack": "^0.5.0",
     "@pika/plugin-build-node": "^0.9.2",
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   },
   "devDependencies": {
     "@gimenete/type-writer": "^0.1.5",
+    "@octokit/tsconfig": "^1.0.0",
     "@octokit/webhooks-definitions": "3.13.2",
     "@pika/pack": "^0.5.0",
     "@pika/plugin-build-node": "^0.9.2",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,12 +1,4 @@
 {
-  "compilerOptions": {
-    "module": "esnext",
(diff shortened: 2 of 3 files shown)


tsconfig.json
```diff
@@ -1,12 +1,4 @@
...
-    "module": "esnext",
-    "moduleResolution": "node",
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
-    "strict": true,
-    "target": "es2020",
-    "resolveJsonModule": true
-  },
+  "extends": "@octokit/tsconfig",
   "include": ["src/**/*"]
```
oscard0m, line 3: We still need to use `include` here because `include` from the extends is resolved from the relative path where is extended from:

https://www.typescriptlang.org/tsconfig#extends

Do you know a way to make it work at "preset" level? If not, I guess we will need to remove that entry in `@octokit/tsconfig` right?

gr2m: Yes, it seems like it. There is an open discussion about it here: https://github.com/microsoft/TypeScript/issues/25430

Let's remove the `"include"` key from `@octokit/tsconfig`. Do you want to send a PR?

oscard0m: > Yes, it seems like it. There is an open discussion about it here: [microsoft/TypeScript#25430](https://github.com/microsoft/TypeScript/issues/25430)
> 
> Let's remove the `"include"` key from `@octokit/tsconfig`. Do you want to send a PR?

Done

package.json
```diff
@@ -16,6 +16,7 @@
   },
   "devDependencies": {
     "@gimenete/type-writer": "^0.1.5",
+    "@octokit/tsconfig": "^1.0.0",
```
gr2m: Update the version once it's out, then this is good to go

```suggestion
    "@octokit/tsconfig": "^1.0.1",
```

gr2m: you will have to update the lock file, too

gr2m (approved):

PR #291, opened 2020-10-02
refactor: remove unreachable code in receive

### LINKED ISSUE #181: [maintainance] Unreachable code?

ankeetmaini:
Fixes #181

### DIFF
--- a/src/event-handler/receive.ts
+++ b/src/event-handler/receive.ts
@@ -17,12 +17,7 @@ function getHooks(
 ): Function[] {
   const hooks = [state.hooks[`${eventName}.${eventPayloadAction}`]];
 
-  if (Array.isArray(eventName)) {
-    eventName.forEach((name) => hooks.push(state.hooks[name]));
-  } else {
-    hooks.push(state.hooks[eventName]);
-  }
-
+  hooks.push(state.hooks[eventName]);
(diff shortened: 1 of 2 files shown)


ankeetmaini:
I've split the changes for #181 in this PR, I'll break #258 into another fixing types. I've overdone changes there.

_P.S: I'm not participating in Hacktoberfest, so this is not about that :)_

gr2m (approved):
Thanks 👍🏼

PR #292, opened 2020-10-03
fix: typescript autocomplete and types for public api

### LINKED ISSUE #250: [Proposal] [TypeScript] prefer interface to conditional types

### LINKED ISSUE #279: event.payload is typed as any for all known events
Reported at https://github.com/probot/probot/issues/1366 by @bcoe

Example

```ts
import { Webhooks } from "@octokit/webhooks";

const events = new Webhooks({
  secret: "secret",
});

events.on("pull_request.labeled", (event) => {
  event.payload
});
```

VS Code screenshot 

image

ankeetmaini:
It also uses the approach defined in #250 to use interface types instead of a big conditional.

Fixes #250 
Fixes #279

> single event
Screenshot 2020-10-03 at 8 42 44 PM

> multiple events
Screenshot 2020-10-03 at 8 44 57 PM

> event names
Screenshot 2020-10-03 at 8 43 09 PM


src/generated/event-names.ts
```diff
@@ -1,381 +1,6 @@
...
-    | SponsorshipEvent
-    | StarEvent
-    | StatusEvent
-    | TeamEvent
-    | TeamAddEvent
-    | WatchEvent
-    | WorkflowDispatchEvent
-    | WorkflowRunEvent;
+  type All = keyof EventTypesPayload;
 }
```
gr2m: can we get rid of this file altogether?

src/index.ts
```diff
@@ -11,19 +11,19 @@ import {
   WebhookError,
   HandlerFunction,
 } from "./types";
-import { EventNames } from "./generated/event-names";
 import { IncomingMessage, ServerResponse } from "http";
+import { EventNames } from "./generated/event-names";
 
-class Webhooks<T extends WebhookEvent = WebhookEvent> {
+class Webhooks<T extends WebhookEvent = WebhookEvent, U = {}> {
```
gr2m, line 18: What exactly is the U here?

gr2m (commented):
thanks for getting this started!

gr2m (commented):
Thanks for getting this started!

src/generated/event-names.ts
ankeetmaini: Done! I only did not remove it as it was exported from entrypoint. Will this be a breaking change for the users?

src/index.ts
ankeetmaini: U represents the additional properties that the transform function returns apart from the event payload object. Used here https://github.com/ankeetmaini/webhooks.js/blob/4e8d65b891f4adaa1426eb505a92b321c307bb2f/test/typescript-validate.ts#L124

gr2m: I think before the type was inferred automatically from the return type of the transform method, is that still the case?

ankeetmaini: I'm not sure how that can be achieved. I can try and move the `Option<T>` param up so that it can be passed down to `on` handler and then try and extract the return type of `Option<T>['transform']` but since the `transform` function would return an intersection type like `WebhookEvent<any> & {some: 'other', type: 'attributes'}` I can't use a conditional value to extract the second part. Mostly because intersection types are not distributed like union ones.

Also if I take the entire value, then intellisense goes out the window as `WebhookEvent<any>` takes precedence.

ankeetmaini: This can probably be solved by changing the `Webhook<T>` signature, right now T is the actual payload, if we're to map that as the event  name, then we can also theoretically get the value returned automatically. But this would be a big change like I was trying to attempt in #258 and I was all over the place :P

gr2m: Okay, let's leave it as is, we can do a follow up PR if we come up with something clever :) 

Can you please add a comment explaining the `U` type parameter?

ankeetmaini: I've added the comment! I'll try and see if I can improve the types further in a follow up PR :D

gr2m (approved):
Changes look good to me, and I confirmed locally that https://github.com/octokit/webhooks.js/issues/279 is resolved by this PR

@dominguezcelada @wolfy1339 can one of you have another look?

wolfy1339:
Seems to work fine for me

wolfy1339 (approved):

oscard0m:
> Changes look good to me, and I confirmed locally that #279 is resolved by this PR
> 
> @dominguezcelada @wolfy1339 can one of you have another look?

Sorry for the delay on following up this. Even it's already merged. Looks good to me 👍🏽 

@ankeetmaini amazing effort and job here. Congrats! :)

PR #321, opened 2020-10-13
feature: export `WebhooksEvents` from @octokit/webhooks.js

### LINKED ISSUE #302: Export `Events` types from `@octokit/webhooks`
follow up to https://github.com/probot/probot/pull/1377#discussion_r501845369

We need access to the Events Types map defined in https://github.com/octokit/webhooks.js/blob/89aa7f76872ea63222670db3dc39299840be6805/src/generated/get-webhook-payload-type-from-event.ts#L7-L353

I think we should call the exported Interface `Events`, or maybe `WebhooksEvents`, because it also includes the special `*` and `error` events?

/cc  @ankeetmaini

ankeetmaini:
Fixes #302 

I went with `WebhooksEvents` as mentioned in your original [comment](https://github.com/octokit/webhooks.js/issues/302#issuecomment-705711304) as we'd have `Events` in future too. Let me know if this is okay, then I'll create a PR on the `probot` repo importing the same name instead of a deep internal file.

Also `WebhooksEvents` or `WebhookEvents`? :)


gr2m (commented):
Thanks!

> Also `WebhooksEvents` or `WebhookEvents`? :)

`WebhookEvents` events, please. Could you make the change? Otherwise it looks good

oscard0m (approved):

wolfy1339 (approved):

PR #322, opened 2020-10-13
feat(sign): support sha256 algorithm

### LINKED ISSUE #320: support SHA-256 signature verification headers from Webhooks
follow up to https://github.com/probot/probot/issues/1379 /cc @gwestersf 

> **Is your feature request related to a problem? Please describe.**
> 
> GitHub will soon begin sending SHA-256 signature headers with Webhooks, when configured with a secret. The header will look like:
> 
> `X-Hub-Signature-256: sha-256=kljsdfjkldasjfkaejkfwekuweur3298r2398r89fdsuf3`
> 
> **Describe the solution you'd like**
>
> Probot should use the SHA-256 header instead of the SHA-1 header for verification. Note that both headers will be sent, and there is no deprecation date on SHA-1 support at this time.

I would start out by creating a test. 

Here is where we currently calculate the signature using sha-1: https://github.com/octokit/webhooks.js/blob/716ee636fc73231b64659cca5bc257439d664ed3/src/sign/index.ts#L11

The signatures have `sha1=` / `sha256=` prefixes respectively, they also have different lenghts. We shouldn't need to introduce any new APIs to differentiate the two. We should be able to implement both with full backwards compatibility

oscard0m:
## 📝 Summary

`sign()` supports 
* `sha256` signature
* keeps retrocompatibility with `sha1` signature
 
## ⛱ Motivation and Context

Fix #320

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,38 +149,10 @@ Returns the `webhooks` API.
 ### webhooks.sign()
 
 ```js
-webhooks.sign(secret, eventPayload);
-webhooks.sign({ secret, algorithm }, eventPayload);
+webhooks.sign(eventPayload);
 ```
 
 <table width="100%">
-  <tr>
-    <td>
-      <code>
-        secret
-      </code>
-      <em>
-        (string)
-      </em>
-    </td>
-    <td>
-      <strong>Required.</strong>
-      secret to generated the signature
-    </td>
-  </tr>
-  <tr>
-    <td>
-      <code>
-        algorithm
-      </code>
-      <em>
-        (string)
-      </em>
-    </td>
-    <td>
-      Algorithm to calculate signature. Can be set to `sha1` or `sha256`. Defaults to `sha1`.
-    </td>
-  </tr>
   <tr>
     <td>
       <code>
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ Returns the `webhooks` API.
 
 ```js
 webhooks.sign(secret, eventPayload);
+webhooks.sign({ secret, algorithm }, eventPayload);
 ```
 
 <table width="100%">
@@ -170,44 +171,14 @@ webhooks.sign(secret, eventPayload);
   <tr>
     <td>
       <code>
-        eventPayload
+        algorithm
       </code>
       <em>
-        (Object)
+        (string)
       </em>
     </td>
     <td>
-      <strong>Required.</strong>
-      Webhook request payload as received from GitHub
-    </td>
-  </tr>
-</table>
-
-or
-
-```js
-webhooks.sign({
-  secret: <my_awesome_secret>,
-  algorithm: <"sha1" | "sha256">
-}, eventPayload);
-```
-
-<table width="100%">
-  <tr>
-    <td>
-      <code>
-        options
-      </code>
-      <em>
-        (Object)
-      </em>
-    </td>
-    <td>
-      <strong>Required.</strong> secret to generated the signature
-      algorithm: secret to generated the signature
-    </td>
-    <td>
-      algorithm to be applied when encrypting (sha1 by default)
+      Algorithm to calculate signature. Can be set to `sha1` or `sha256`. Defaults to `sha1`.
     </td>
   </tr>
   <tr>
--- a/README.md
+++ b/README.md
@@ -149,10 +149,67 @@ Returns the `webhooks` API.
 ### webhooks.sign()
 
 ```js
-webhooks.sign(eventPayload);
+webhooks.sign(secret, eventPayload);
 ```
 
 <table width="100%">
+  <tr>
+    <td>
+      <code>
+        secret
+      </code>
+      <em>
+        (string)
+      </em>
+    </td>
+    <td>
+      <strong>Required.</strong>
+      secret to generated the signature
+    </td>
+  </tr>
+  <tr>
+    <td>
+      <code>
+        eventPayload
+      </code>
+      <em>
+        (Object)
+      </em>
+    </td>
+    <td>
+      <strong>Required.</strong>
+      Webhook request payload as received from GitHub
+    </td>
+  </tr>
+</table>
+
+or
+
+```js
+webhooks.sign({
+  secret: <my_awesome_secret>,
+  algorithm: <"sha1" | "sha256">
+}, eventPayload);
+```
+
+<table width="100%">
+  <tr>
+    <td>
+      <code>
+        options
+      </code>
+      <em>
+        (Object)
+      </em>
+    </td>
+    <td>
+      <strong>Required.</strong> secret to generated the signature
+      algorithm: secret to generated the signature
+    </td>
+    <td>
+      algorithm to be applied when encrypting (sha1 by default)
+    </td>
+  </tr>
   <tr>
     <td>
       <code>
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -14,16 +14,14 @@ export function sign(
   options: SignOptions | string,
   payload: string | object
 ): string {
-  let secret;
-  let algorithm;
+  const { secret, algorithm } =
+    typeof options === "string"
+      ? { secret: options, algorithm: Algorithm.SHA1 }
+      : {
+          secret: options.secret,
+          algorithm: options.algorithm || Algorithm.SHA1,
+        };
 
-  if (typeof options === "string") {
-    secret = options;
-    algorithm = Algorithm.SHA1;
-  } else {
-    secret = options.secret;
-    algorithm = options.algorithm || Algorithm.SHA1;
-  }
   // @ts-ignore throw friendly error message when required options are missing
   if (!secret || !payload) {
     throw new TypeError("[@octokit/webhooks] secret & payload required");
@@ -31,7 +29,7 @@ export function sign(
 
   if (!Object.values(Algorithm).includes(algorithm as Algorithm)) {
     throw new TypeError(
-      "[@octokit/webhooks] algorithm should be 'sha1' or 'sha256'"
+      `[@octokit/webhooks] Algorithm ${algorithm} is not supported. Must be  'sha1' or 'sha256'`
     );
   }
 
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -7,7 +7,7 @@ export enum Algorithm {
 
 type SignOptions = {
   secret: string;
-  algorithm?: Algorithm;
+  algorithm?: string;
 };
 
 export function sign(
@@ -27,6 +27,10 @@ export function sign(
   // @ts-ignore throw friendly error message when required options are missing
   if (!secret || !payload) {
     throw new TypeError("secret & payload required");
+  }
+
+  if (!Object.values(Algorithm).includes(algorithm as Algorithm)) {
+    throw new TypeError("algorithm should be 'sha1' or 'sha256'");
   }
 
   payload =
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -1,6 +1,29 @@
 import { createHmac } from "crypto";
 
-export function sign(secret: string, payload: string | object): string {
+export enum Algorithm {
+  SHA1 = "sha1",
+  SHA256 = "sha256",
+}
+
+type SignOptions = {
+  secret: string;
+  algorithm?: Algorithm;
+};
+
+export function sign(
+  options: SignOptions | string,
+  payload: string | object
+): string {
+  let secret;
+  let algorithm;
+
+  if (typeof options === "string") {
+    secret = options;
+    algorithm = Algorithm.SHA1;
+  } else {
+    secret = options.secret;
+    algorithm = options.algorithm ?? Algorithm.SHA1;
+  }
   // @ts-ignore throw friendly error message when required options are missing
   if (!secret || !payload) {
     throw new TypeError("secret & payload required");
@@ -8,7 +31,9 @@ export function sign(secret: string, payload: string | object): string {
 
   payload =
     typeof payload === "string" ? payload : toNormalizedJsonString(payload);
-  return "sha1=" + createHmac("sha1", secret).update(payload).digest("hex");
+  return `${algorithm}=${createHmac(algorithm, secret)
+    .update(payload)
+    .digest("hex")}`;
 }
 
 function toNormalizedJsonString(payload: object) {
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -26,11 +26,11 @@ export function sign(
   }
   // @ts-ignore throw friendly error message when required options are missing
   if (!secret || !payload) {
-    throw new TypeError("secret & payload required");
+    throw new TypeError("[@octokit/webhooks] secret & payload required");
   }
 
   if (!Object.values(Algorithm).includes(algorithm as Algorithm)) {
-    throw new TypeError("algorithm should be 'sha1' or 'sha256'");
+    throw new TypeError("[@octokit/webhooks] algorithm should be 'sha1' or 'sha256'");
   }
 
   payload =
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -22,7 +22,7 @@ export function sign(
     algorithm = Algorithm.SHA1;
   } else {
     secret = options.secret;
-    algorithm = options.algorithm ?? Algorithm.SHA1;
+    algorithm = options.algorithm || Algorithm.SHA1;
   }
   // @ts-ignore throw friendly error message when required options are missing
   if (!secret || !payload) {
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -30,7 +30,9 @@ export function sign(
   }
 
   if (!Object.values(Algorithm).includes(algorithm as Algorithm)) {
-    throw new TypeError("[@octokit/webhooks] algorithm should be 'sha1' or 'sha256'");
+    throw new TypeError(
+      "[@octokit/webhooks] algorithm should be 'sha1' or 'sha256'"
+    );
   }
 
   payload =
--- a/src/verify/index.ts
+++ b/src/verify/index.ts
@@ -3,7 +3,7 @@ import { Buffer } from "buffer";
 import { sign } from "../sign/index";
 
 const getAlgorithm = (signature: string) => {
-  return signature.startsWith("sha256=")? "sha256" : "sha1";
+  return signature.startsWith("sha256=") ? "sha256" : "sha1";
 };
 
 export function verify(
--- a/src/verify/index.ts
+++ b/src/verify/index.ts
@@ -2,6 +2,10 @@ import { timingSafeEqual } from "crypto";
 import { Buffer } from "buffer";
 import { sign } from "../sign/index";
 
+const getAlgorithm = (signature: string) => {
+  return signature.indexOf("sha256=") === 0 ? "sha256" : "sha1";
+};
+
 export function verify(
   secret?: string,
   eventPayload?: object,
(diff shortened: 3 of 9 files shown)


test/integration/sign-test.ts
```diff
@@ -20,12 +20,22 @@ test("sign(secret) without eventPayload throws", () => {
...
+});
+
+test("sign(secret, eventPayload) with eventPayload as string returns expected sha1 signature", () => {
   const signature = sign(secret, JSON.stringify(eventPayload));
   expect(signature).toBe("sha1=d03207e4b030cf234e3447bac4d93add4c6643d8");
 });
+
+test("sign(secret, eventPayload) with eventPayload as string returns expected sha256 signature", () => {
+  const signature = sign(secret, JSON.stringify(eventPayload));
+  expect(signature).toBe("sha256=kljsdfjkldasjfkaejkfwekuweur3298r2398r89fdsuf3");
```
oscard0m: In order to support this I only came up with a modification with the signature with an optional third parameter with Webhook header's, which will be the one determining if is going to use `sha1` or `sha256`. Would that be the approach to follow or you had something different in mind?

```typescript
function sign(secret: string, payload: string | object, headers?: string): string {
  // @ts-ignore throw friendly error message when required options are missing
  if (!secret || !payload) {
    throw new TypeError("secret & payload required");
  }

  payload =
    typeof payload === "string" ? payload : toNormalizedJsonString(payload);
  
  const algorithm = containsSha256Header(headers)? "sha256" : "sha1"
  return `${algorithm}=${createHmac(algorithm, secret).update(payload).digest("hex")}`;
}
```

@gr2m

gr2m: I see why the additional parameter is needed. I think passing the `headers` doesn't feel right here, the `sign` method does not need to be aware of all headers, it only needs to know which algorithm to use. 

How about this:

1. By default `sign(secret, payload)` returns a signature based on the `sha1` algorithm, so we don't introduce a breaking change. We  can change the default with the next breaking version.
2. In order to get a a signature using the `sha-256` algorithm, we can either introduce a third argument `algorithm?: string`, or we allow the first argument to be an object with two keys: `sign({ secret, algorithm }, payload)`. I think I'd prefer the latter. 

What do you think @dominguezcelada @wolfy1339?

wolfy1339: Option 2 would be okay

oscard0m: I prefer option 2, so we make options object more extensible instead of a long list of arguments for the method. I go for it

src/sign/index.ts
```diff
@@ -1,14 +1,43 @@
...
+    algorithm = options.algorithm ?? Algorithm.SHA1;
+  }
   // @ts-ignore throw friendly error message when required options are missing
   if (!secret || !payload) {
     throw new TypeError("secret & payload required");
   }
 
+  if (!Object.values(Algorithm).includes(algorithm as Algorithm)) {
+    throw new TypeError("algorithm should be 'sha1' or 'sha256'");
+  }
```
oscard0m: Should be this something we control from `@octokit/webhooks.js` or we should pass whatever string the user enters and `createHmac()` should buble up the error?

---

In case we keep this `enum` approach, TypeScript question, is there a more elegant way to control internally if a `string` is one of the allowed values under an `enum` (but keep using `string` instead of `enum` as signature of the method)?

src/sign/index.ts
```diff
@@ -1,14 +1,43 @@
...
+  let secret;
+  let algorithm;
+
+  if (typeof options === "string") {
+    secret = options;
+    algorithm = Algorithm.SHA1;
+  } else {
+    secret = options.secret;
+    algorithm = options.algorithm ?? Algorithm.SHA1;
+  }
```
oscard0m: I see this a bit messy, if you come up with a cleaner way to write this, please let me know :)

test/typescript-validate.ts
```diff
@@ -149,3 +151,5 @@ export function webhookErrorTestDeprecated(error: WebhookError) {
   console.log(event);
   console.log(request);
 }
+
+sign("randomSecret", {});
```
oscard0m: Should we move this to deprecated part of typescript-validate? or not yet?

src/sign/index.ts
gr2m: I'd prefer to control it. GitHub only supports `sha1` and `sha256`, it makes no sense for us to support any other algorithm, and could be the source of a hard to debug problem. Throwing an error will make it obvious.

As a side note, I started to prefix error and log messages with the package name
```suggestion
  if (!Object.values(Algorithm).includes(algorithm as Algorithm)) {
    throw new TypeError("[@octokit/webhooks] algorithm should be 'sha1' or 'sha256'");
  }
```

We should do that in with all errors / log messages, but in a separate PR

test/typescript-validate.ts
gr2m: no I wouldn't consider it deprecated, both passing a string or an object are valid, I don't expect we will deprecate either anytime soon

src/sign/index.ts
oscard0m: Created issue associated to this comment ;)

oscard0m: > In case we keep this enum approach, TypeScript question, is there a more elegant way to control internally if a string is one of the allowed values under an enum (but keep using string instead of enum as signature of the method)?

@wolfy1339 TS question to the rescue! Any good idea on how to control this in a less hacky way? 😇

src/sign/index.ts
oscard0m: > I see this a bit messy, if you come up with a cleaner way to write this, please let me know :)

@wolfy1339 TS question to the rescue! Any good idea on how to control this in a less hacky way? 😇

src/sign/index.ts
```diff
@@ -1,14 +1,45 @@
...
+  let secret;
+  let algorithm;
+
+  if (typeof options === "string") {
+    secret = options;
+    algorithm = Algorithm.SHA1;
+  } else {
+    secret = options.secret;
+    algorithm = options.algorithm || Algorithm.SHA1;
+  }
```
gr2m: I think I'd prefer

```suggestion
  const {secret, algorithm} =
    typeof options === 'string'
      ? {secret: options, algorithm: Algorithm.SHA1}
      : {secret: options.secret, algorith: options.algorithm || Algorithm.SHA1};
```

src/sign/index.ts
```diff
@@ -1,14 +1,45 @@
...
   // @ts-ignore throw friendly error message when required options are missing
   if (!secret || !payload) {
-    throw new TypeError("secret & payload required");
+    throw new TypeError("[@octokit/webhooks] secret & payload required");
+  }
+
+  if (!Object.values(Algorithm).includes(algorithm as Algorithm)) {
+    throw new TypeError(
+      "[@octokit/webhooks] algorithm should be 'sha1' or 'sha256'"
```
gr2m:
```suggestion
      `[@octokit/webhooks] Algorithm ${algorithm} is not supported. Must be  'sha1' or 'sha256'`
```

test/integration/sign-test.ts
```diff
@@ -20,12 +20,71 @@ test("sign(secret) without eventPayload throws", () => {
   expect(() => sign.bind(null, secret)()).toThrow();
 });
 
-test("sign(secret, eventPayload) with eventPayload as object returns expected signature", () => {
-  const signature = sign(secret, eventPayload);
-  expect(signature).toBe("sha1=d03207e4b030cf234e3447bac4d93add4c6643d8");
+test("sign({secret, algorithm}) with an invalit algorithm throws", () => {
```
gr2m:
```suggestion
test("sign({secret, algorithm}) with invalid algorithm throws", () => {
```

test/integration/sign-test.ts
```diff
@@ -20,12 +20,71 @@ test("sign(secret) without eventPayload throws", () => {
...
+      const signature = sign(secret, JSON.stringify(eventPayload));
+      expect(signature).toBe("sha1=d03207e4b030cf234e3447bac4d93add4c6643d8");
+    });
+
+    test("sign({secret}, eventPayload)", () => {
+      const signature = sign({ secret }, JSON.stringify(eventPayload));
+      expect(signature).toBe("sha1=d03207e4b030cf234e3447bac4d93add4c6643d8");
+    });
+
+    test("sign({secret, sha1}, eventPayload)", () => {
```
gr2m:
```suggestion
    test("sign({secret, algorithm: 'sha1'}, eventPayload)", () => {
```

test/integration/sign-test.ts
```diff
@@ -20,12 +20,71 @@ test("sign(secret) without eventPayload throws", () => {
...
+      const signature = sign(secret, eventPayload);
+      expect(signature).toBe("sha1=d03207e4b030cf234e3447bac4d93add4c6643d8");
+    });
+
+    test("sign({secret}, eventPayload)", () => {
+      const signature = sign({ secret }, eventPayload);
+      expect(signature).toBe("sha1=d03207e4b030cf234e3447bac4d93add4c6643d8");
+    });
+
+    test("sign({secret, sha1}, eventPayload)", () => {
```
gr2m:
```suggestion
    test("sign({secret, algorithm: 'sha1'}, eventPayload)", () => {
```

test/integration/sign-test.ts
```diff
@@ -20,12 +20,71 @@ test("sign(secret) without eventPayload throws", () => {
...
+      const signature = sign(
+        { secret, algorithm: "sha1" },
+        JSON.stringify(eventPayload)
+      );
+      expect(signature).toBe("sha1=d03207e4b030cf234e3447bac4d93add4c6643d8");
+    });
+  });
+
+  describe("resturns expected sha256 signature", () => {
```
gr2m:
```suggestion
  describe("returns expected sha256 signature", () => {
```

test/integration/sign-test.ts
```diff
@@ -20,12 +20,71 @@ test("sign(secret) without eventPayload throws", () => {
...
+    test("sign({secret, algorithm}, eventPayload)", () => {
```
gr2m:
```suggestion
    test("sign({secret, algorithm: 'sha256'}, eventPayload)", () => {
```

gr2m (commented):
looks good overall, just a few comments

README.md
```diff
@@ -149,10 +149,67 @@ Returns the `webhooks` API.
...
+        (Object)
+      </em>
+    </td>
+    <td>
+      <strong>Required.</strong> secret to generated the signature
+      algorithm: secret to generated the signature
+    </td>
+    <td>
+      algorithm to be applied when encrypting (sha1 by default)
+    </td>
```
oscard0m: @gr2m What do you think?

gr2m: I'd list the two call signatures on top and then have a single table, like this:

<details>

````md
```js
webhooks.sign(secret, eventPayload);
webhooks.sign({ secret, algorithm }, eventPayload);
```

<table width="100%">
  <tr>
    <td>
      <code>
        secret
      </code>
      <em>
        (string)
      </em>
    </td>
    <td>
      <strong>Required.</strong>
      secret to generated the signature
    </td>
  </tr>
  <tr>
    <td>
      <code>
        algorithm
      </code>
      <em>
        (string)
      </em>
    </td>
    <td>
      Algorithm to calculate signature. Can be set to `sha1` or `sha256`. Defaults to `sha1`.
    </td>
  </tr>
  <tr>
    <td>
      <code>
        eventPayload
      </code>
      <em>
        (Object)
      </em>
    </td>
    <td>
      <strong>Required.</strong>
      Webhook request payload as received from GitHub
    </td>
  </tr>
</table>
````

</details>

**Review by gr2m — COMMENTED** [review]
I'm not sure if you are still working on the PR. What I think we should add is a test that receives a webhook request with both headers and then uses the new 256 signature for verification

I think the change will need to happen here: https://github.com/octokit/webhooks.js/blob/6ff2cd81c9dd86cfc11333ef92d56705e208b9fb/src/verify/index.ts#L15

you can check the prefix of `signature` to derive the algorithm used to calculate the signature, and then pass the algoritm to the sign method. And you will have to adapt the code around https://github.com/octokit/webhooks.js/blob/6ff2cd81c9dd86cfc11333ef92d56705e208b9fb/src/middleware/middleware.ts#L45 where we retrieve the signature from the request header

File: README.md [hunk 13]
```diff
@@ -149,10 +149,38 @@ Returns the `webhooks` API.
 ### webhooks.sign()
 
 ```js
-webhooks.sign(eventPayload);
+webhooks.sign(secret, eventPayload);
+webhooks.sign({ secret, algorithm }, eventPayload);
```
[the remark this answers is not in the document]
oscard0m: > Sorry I only now realized that this change is incorrect. When you use the `sign` method standalone, than you can pass the `secret` or the `{ secret, algorithm }` option. But on a `webhooks` instance, the pass the first parameter implicitly in the constructor:
> 
> https://github.com/octokit/webhooks.js/blob/6ff2cd81c9dd86cfc11333ef92d56705e208b9fb/src/index.ts#L55
> 
> So I think you have to revert the changes here, but you should update the docs at [`6ff2cd8`/src/sign](https://github.com/octokit/webhooks.js/tree/6ff2cd81c9dd86cfc11333ef92d56705e208b9fb/src/sign) instead
> 
> Sorry I should have caught that earlier

No problem! My fault to not pay more attention. Let me revert changes + add them in sign's README.

src/verify/index.ts
```diff
@@ -2,6 +2,10 @@ import { timingSafeEqual } from "crypto";
 import { Buffer } from "buffer";
 import { sign } from "../sign/index";
 
+const getAlgorithm = (signature: string) => {
+  return signature.indexOf("sha256=") === 0 ? "sha256" : "sha1";
+};
```
oscard0m: @gr2m is it fine the check at this stage the algorithm to be applied or you want to be decided before and change the signature of verify with an `algorithm?` parameter?

oscard0m (commented):
Let me know if the test coverage I added would be enough or you would add more stuff / cases

src/verify/index.ts
```diff
@@ -2,6 +2,10 @@ import { timingSafeEqual } from "crypto";
 import { Buffer } from "buffer";
 import { sign } from "../sign/index";
 
+const getAlgorithm = (signature: string) => {
+  return signature.indexOf("sha256=") === 0 ? "sha256" : "sha1";
```
gr2m:
```suggestion
  return signature.startsWith("sha256=")? "sha256" : "sha1";
```

src/verify/index.ts
gr2m: Your implementation works good to me 👍🏼

gr2m (commented):
> Let me know if the test coverage I added would be enough or you would add more stuff / cases

Can you add a test case to `test/integration/server-test.ts` with the new `X-Hub-Signature-256` header set? In fact, I think we should update all existing tests to have both headers, and then have one test where the new `X-Hub-Signature-256` header is missing, as this is still the fact for the current GitHub Enterprise Server versions. If that's too much work, we could do that tests cleanup in a follow up PR, and for now just add one test where both headers are present, and make sure the `X-Hub-Signature-256`  signature is used for verification (e.g. set an invalid value for the `X-Hub-Signature` header)

test/integration/server-test.ts
```diff
@@ -6,6 +6,18 @@ import { promisify } from "util";
...
 
+const signatureHeaders = [
+  {
+    algorithm: "sha1",
+    signature: "sha1=f4d795e69b5d03c139cc6ea991ad3e5762d13e2f",
+  },
+  {
+    algorithm: "sha256",
+    signature:
+      "sha256=4864d2759938a15468b5df9ade20bf161da9b4f737ea61794142f3484236bda3",
```
dominguezcelada, line 17: WIP. After adding coverage for `sha256` some tests failing. Not sure if the sha256 is incorrect or there's a bug at sign/verify stage. Tomorrow I will keep checking

test/integration/server-test.ts
```diff
@@ -6,6 +6,10 @@ import { promisify } from "util";
 import { Webhooks } from "../../src";
 import pushEventPayload from "../fixtures/push-payload.json";
 
+const signatureSha1 = "sha1=f4d795e69b5d03c139cc6ea991ad3e5762d13e2f";
+const signatureSha256 =
+  "sha256=4864d2759938a15468b5df9ade20bf161da9b4f737ea61794142f3484236bda3";
```
oscard0m: tests failing. I hope is tha sha256 I generated wrong. Tomorrow I'll keep researching

gr2m: try 

```suggestion
  "sha256=2b49327af77d51c4c23700118b11d15c81d90c3bd57dafbbe32eb50085ce67e0";
```

That made the tests pass for me. I calculated it like this:

```js
sign(
  { secret: "mysecret", algorithm: "sha256" },
  pushEventPayload
);
```

You probably had a typo in the `secret`?

oscard0m: Something wrong with the secret or payload I was passing yep.

test/integration/server-test.ts
```diff
@@ -47,6 +51,77 @@ describe("server-test", () => {
...
+
+      .then(() => {
+        return axios.post(
+          `http://localhost:${availablePort}`,
+          pushEventPayload,
+          {
+            headers: {
+              "X-GitHub-Delivery": "123e4567-e89b-12d3-a456-426655440000",
+              "X-GitHub-Event": "push",
+              "X-Hub-Signature": signatureSha1,
```
gr2m: the description says "with wrong sha1 but right sha256", but it seems that you pass in a correct signature here? Maybe replace it with some bogus value instead?
```suggestion
              "X-Hub-Signature": "ignored",
```

test/integration/verify-test.ts
```diff
@@ -4,63 +4,114 @@ const eventPayload = {
...
 
-test("verify(secret, eventPayload, signature) returns false for correct secret", () => {
-  const signatureMatches = verify("foo", eventPayload, signature);
-  expect(signatureMatches).toBe(false);
+  test("verify(secret, eventPayload, signature) returns false for correct secret", () => {
+    const signatureMatches = verify("foo", eventPayload, signatureSHA1);
+    expect(signatureMatches).toBe(false);
+  });
+
+  test("verify(secret, eventPayload, signature) returns true if eventPayload contains (#71)", () => {
```
gr2m: it looks like the test now exists twice? Did you mean to test with different algorithms? If so you need to change the test and maybe mention the algorithm in the test description? Also can you amend the description? I think it should have said
```suggestion
  test("verify(secret, eventPayload, signature) returns true if eventPayload contains special characters (#71)", () => {
```

test/integration/verify-test.ts
```diff
@@ -4,63 +4,114 @@ const eventPayload = {
...
+    expect(() => verify.bind(null, secret, eventPayload)()).toThrow();
+  });
 });
 
-test("verify(secret, eventPayload, signature) returns true for correct signature", () => {
-  const signatureMatches = verify(secret, eventPayload, signature);
-  expect(signatureMatches).toBe(true);
-});
+describe("for SHA1 signature", () => {
+  test("verify(secret, eventPayload, signature) returns true for correct signature", () => {
```
gr2m: can you mention the different algorithm in the test descriptions so you don't use the same description for two tests? E.g.
```suggestion
  test("verify(secret, eventPayload, signatureSHA1) returns true for correct signature", () => {
```

gr2m (changes requested):

test/integration/verify-test.ts
oscard0m: I forgot to change the algorithm there!🙏🏽 

For each unit test (sha1 and sha256) for special characters:
- [x] applying proper algorithm
- [x] mentioning algorithm in test description
- [x] amend `description()`
- [x] added `special characters` to the sentence

gr2m (approved):
Great work, thanks for tackling this Dominguez!

PR #324, opened 2020-10-13
feat: Add onAny and onError functions

### LINKED ISSUE #304: Create .onAny and .onError methods, deprecate "*" and "error" events
Follow up to https://github.com/octokit/webhooks.js/issues/180

I would reserve the `.on(event, handler)` method for actual events only, and define new methods for special events that are unique to `@octokit/webhooks`.

That will also address https://github.com/octokit/webhooks.js/issues/302, because we would no longer need to define `*` and `error` as part of the overall event map

wolfy1339:
Fixes #304

### DIFF
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -1,4 +1,8 @@
-import { receiverOn as on } from "./on";
+import {
+  receiverOn as on,
+  receiverOnAny as onAny,
+  receiverOnError as onError,
+} from "./on";
 import { receiverHandle as receive } from "./receive";
 import { removeListener } from "./remove-listener";
 import { Options, State } from "../types";
@@ -14,8 +18,8 @@ export function createEventHandler(options: Options<any>) {
 
   return {
     on: on.bind(null, state),
-    onAny: on.bind(null, state).bind(null, "*"),
-    onError: on.bind(null, state).bind(null, "error"),
+    onAny: onAny.bind(null, state),
+    onError: onError.bind(null, state),
     removeListener: removeListener.bind(null, state),
     receive: receive.bind(null, state),
   };
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -14,6 +14,8 @@ export function createEventHandler(options: Options<any>) {
 
   return {
     on: on.bind(null, state),
+    onAny: on.bind(null, state).bind(null, "*"),
+    onError: on.bind(null, state).bind(null, "error"),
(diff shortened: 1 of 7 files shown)


oscard0m (dismissed):
LGTM

src/event-handler/index.ts
```diff
@@ -14,6 +14,8 @@ export function createEventHandler(options: Options<any>) {
 
   return {
     on: on.bind(null, state),
+    onAny: on.bind(null, state).bind(null, "*"),
+    onError: on.bind(null, state).bind(null, "error"),
```
wolfy1339: This is probably not the best solution. But it ensures backwards compatibility and it works

test/typescript-validate.ts
```diff
@@ -69,12 +69,19 @@ export default async function () {
 
   verify("randomSecret", {}, "randomSignature");
 
+  // This is deprecated usage
```
oscard0m, line 72: 💡  there's a deprecated section at the end of this file :)

test/typescript-validate.ts
```diff
@@ -124,6 +131,7 @@ export default async function () {
     console.log(event.foo);
   });
 
+  // This is deprecated usage
```
oscard0m, line 134: https://github.com/octokit/webhooks.js/pull/324/files#r504256440

src/event-handler/index.ts
gr2m: I would want to log a deprecation message when `.on()` is called with either `*` or `error`. While this is a neat trick, it won't allow us to do that.

I'd also like to add tests for `.on("error", handler)` and `.on("*", handler)` to a new `test/integration/deprecations-test.js` file, similar to what we have in other @octokit repositories. That makes it easier to prepare new breaking releases, and having these tests prevents accidental regressions. The tests should test for the deprecation waring to be logged

gr2m: btw you can bind as many arguments as you like, so you could do this instead: 

```js
    onAny: on.bind(null, state, "*"),
    onError: on.bind(null, state, "error"),
```

wolfy1339: Done

wolfy1339:
@gr2m I can't seem to push to the branch anymore

gr2m:
@wolfy1339 sorry for that. I've invited you and @dominguezcelada to the new @octokit/js-community team, you should have write access to all JS repositories now :)

gr2m (approved):
looks great 👍🏼

oscard0m (approved):

PR #339, opened 2020-10-26
fix: prefix error messages with "[@octokit/webhooks] "

### LINKED ISSUE #325: Prefix error messages with "[@octokit/webhooks] "
To add initial `[@octokit/webhooks]` starting message on every Error of the project.

---

I'd prefer to control it. GitHub only supports `sha1` and `sha256`, it makes no sense for us to support any other algorithm, and could be the source of a hard to debug problem. Throwing an error will make it obvious.

As a side note, I started to prefix error and log messages with the package name
```suggestion
  if (!Object.values(Algorithm).includes(algorithm as Algorithm)) {
    throw new TypeError("[@octokit/webhooks] algorithm should be 'sha1' or 'sha256'");
  }
```

We should do that in with all errors / log messages, but in a separate PR

_Originally posted by @gr2m in https://github.com/octokit/webhooks.js/pull/322#r504259368_

srujandeshpande:
Fixes #325 

Prefixed all error messages with "[@octokit/webhooks] "
middleware-test.ts was modified to expect the new error.

-----
[View rendered src/event-handler/README.md](https://github.com/srujandeshpande/webhooks.js/blob/master/src/event-handler/README.md)

### DIFF
--- a/src/event-handler/README.md
+++ b/src/event-handler/README.md
@@ -31,7 +31,7 @@ const { verify } = require('@octokit/webhooks')
 const secret = 'mysecret'
 
 if (!verify(secret, request.payload, request.headers['x-hub-signature'])) {
-  throw new Error('Signature does not match event payload & secret')
+  throw new Error('[@octokit/webhooks] Signature does not match event payload & secret')
 }
 ```
 
--- a/src/event-handler/README.md
+++ b/src/event-handler/README.md
@@ -31,7 +31,7 @@ const { verify } = require('@octokit/webhooks')
 const secret = 'mysecret'
 
 if (!verify(secret, request.payload, request.headers['x-hub-signature'])) {
-  throw new Error('[@octokit/webhooks] Signature does not match event payload & secret')
+  throw new Error('Signature does not match event payload & secret')
 }
 ```
 
--- a/test/integration/event-handler-test.ts
+++ b/test/integration/event-handler-test.ts
@@ -75,7 +75,7 @@ test("events", (done) => {
       });
 
       eventHandler.on("push", () => {
-        throw new Error("oops");
+        throw new Error("[@octokit/webhooks] oops");
       });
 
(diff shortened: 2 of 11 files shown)


test/integration/event-handler-test.ts
```diff
@@ -75,7 +75,7 @@ test("events", (done) => {
       });
 
       eventHandler.on("push", () => {
-        throw new Error("oops");
+        throw new Error("[@octokit/webhooks] oops");
```
gr2m: you don't need the prefix in the test files, only in the code (`src/`). Could you change these back please, to keep the tests simple?

src/event-handler/README.md
```diff
@@ -31,7 +31,7 @@ const { verify } = require('@octokit/webhooks')
 const secret = 'mysecret'
 
 if (!verify(secret, request.payload, request.headers['x-hub-signature'])) {
-  throw new Error('Signature does not match event payload & secret')
+  throw new Error('[@octokit/webhooks] Signature does not match event payload & secret')
```
gr2m: As this is user code, and not part of `@octokit/webhooks`, we don't need to add the prefix here. It's just a usage example. Can you remove it in this place, please?

gr2m (changes requested):
Thanks for looking into this Srujan! There are a few changes I'd like you to revert. I'm sorry we should have made the instructions more clear. We only need to add the `[@octokit/webhooks] ` prefix to `*.ts` files in the `src/` folder

test/integration/event-handler-test.ts
srujandeshpande: Apologies, reverted.

src/event-handler/README.md
srujandeshpande: Done!

gr2m (approved):
Great PR, thank you! 💐

PR #369, opened 2020-12-03
fix(typescript): add types for `check_run.requested_action`

codebytere:
Bump `@octokit/webhooks-definitions` to pick up https://github.com/octokit/webhooks/pull/222 and fix missing `requested_action?: WebhookPayloadCheckRunRequestedAction;` in `WebhookPayloadCheckRun`

cc @gr2m

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,29 +13,24 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.11.0",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.11.0.tgz",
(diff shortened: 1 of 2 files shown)


gr2m (changes requested):
Thanks! I should automate this update so we don't need to wait for Dependabot to do its thing over night.

You need to run `npm run generate-types` to update the TypeScript definitions and push the change. I wanted to do the update for you but I cannot push to your fork repository, it's shown as "unknown repository"?

codebytere:
@gr2m oops yep i'll fix it, one sec!

codebytere:
Open at https://github.com/octokit/webhooks.js/pull/370

oscard0m:
> I should automate this update so we don't need to wait for Dependabot to do its thing over night.

Do you want me to create an issue for this? @gr2m :)

gr2m:
Thanks, already done via https://github.com/octokit/webhooks.js/commit/f1d9c485420fdd0e39814a2cf134cd0c2cf7cacc :)

PR #373, opened 2020-12-04
fix(typescript): mark `verify` parameters as required

G-Rath:
The parameters for `verify` are not optional - this matches the pattern used for `sign`.

### DIFF
--- a/src/middleware/verify-and-receive.ts
+++ b/src/middleware/verify-and-receive.ts
@@ -5,7 +5,12 @@ export function verifyAndReceive(
   state: State,
   event: WebhookEvent & { signature: string }
 ): any {
-  const matchesSignature = verify(state.secret, event.payload, event.signature);
+  // verify will validate that the secret is not undefined
+  const matchesSignature = verify(
+    state.secret!,
+    event.payload,
+    event.signature
+  );
 
   if (!matchesSignature) {
     const error = new Error(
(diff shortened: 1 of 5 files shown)


src/middleware/verify-and-receive.ts
```diff
@@ -5,7 +5,12 @@ export function verifyAndReceive(
   state: State,
   event: WebhookEvent & { signature: string }
 ): any {
-  const matchesSignature = verify(state.secret, event.payload, event.signature);
+  // verify will validate that the secret is not undefined
+  const matchesSignature = verify(
+    state.secret!,
```
G-Rath, line 10: I'm typically not a fan of using `!` - ideally I'd have an assertion here about the `state.secret`, but tbh this isn't the worse use of it 🤷

gr2m: We are all still figuring out the whole TypeScript thing, we got much better at it since we migrated `@octokit/webhooks`, but there is surely lots of potential of improvement :)

gr2m (approved):
Great changes, thanks!

src/middleware/verify-and-receive.ts
G-Rath: No problem - I'm a big TS fan myself, so happy to help and be pinged on things :)

Issue #271, opened 2020-09-19
Restore aggregate-error external dependency


oscard0m:
* To solve TypeScript issue when importing `aggregate-error` dependency (context in #245)
* Undo patch changes to bypass the issue (details in https://github.com/octokit/webhooks.js/pull/270)

PR #374, opened 2020-12-04
refactor(typescript): use `ts-expect-error` instead of `ts-ignore`

G-Rath:
Replaces uses of `ts-ignore` with `ts-expect-error` - these are better as they effectively invert typescripts behaviour so that it errors if there *isn't* an error (unlike `ts-ignore` which just causes errors to be ignored regardless of if there are actually errors).

This revealed a couple of `ts-ignore`s that are not needed 🎉

### DIFF
--- a/src/event-handler/receive.ts
+++ b/src/event-handler/receive.ts
@@ -20,8 +20,7 @@ function getHooks(
   hooks.push(state.hooks[eventName]);
   hooks.push(state.hooks["*"]);
 
-  // @ts-ignore
-  return [].concat(...hooks.filter(Boolean));
+  return ([] as Function[]).concat(...hooks.filter(Boolean));
 }
 
 // main handler function
--- a/src/middleware/get-payload.ts
+++ b/src/middleware/get-payload.ts
@@ -10,7 +10,7 @@ export function getPayload(
   // If request.body already exists we can stop here
   // See https://github.com/octokit/webhooks.js/pull/23
 
-  // @ts-ignore
+  // @ts-expect-error
   if (request.body) return Promise.resolve(request.body);
 
   return new Promise((resolve, reject) => {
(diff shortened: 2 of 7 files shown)


src/event-handler/receive.ts
```diff
@@ -1,4 +1,3 @@
-// @ts-ignore to address #245
```
wolfy1339: Please see #245 and #271

src/middleware/get-payload.ts
```diff
@@ -1,4 +1,3 @@
-// @ts-ignore to address #245
```
wolfy1339: Please see #245 and #271

wolfy1339 (changes requested):

src/event-handler/receive.ts
```diff
@@ -20,8 +19,7 @@ function getHooks(
   hooks.push(state.hooks[eventName]);
   hooks.push(state.hooks["*"]);
 
-  // @ts-ignore
-  return [].concat(...hooks.filter(Boolean));
+  return ([] as Function[]).concat(...hooks.filter(Boolean));
```
G-Rath, line 23: It seems that there's a disconnect between the typechecking done by `validate:ts` & `ts-jest`, meaning that `validate:ts` doesn't think this is an error but `ts-jest` does.

I've just resolved the error by casting the array, to remove the need for any ignoring of ts :)

src/event-handler/receive.ts
G-Rath: Ah so these do need to remain as `@ts-ignore` - fair enough, I'll update.

wolfy1339 (approved):

gr2m (approved):
Nice, thanks @G-Rath!

PR #375, opened 2020-12-04
chore: update `ts-jest` config

G-Rath:
Addresses this warning in CI:

> ts-jest[config] (WARN) The option `tsConfig` is deprecated and will be removed in ts-jest 27, use `tsconfig` instead

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     },
     "globals": {
       "ts-jest": {
-        "tsConfig": {
+        "tsconfig": {
           "esModuleInterop": true
         }
       }


gr2m (approved):
sweet, thank you!

wolfy1339:
If you don't mind, can you check other Octokit repositories and update with the same change

G-Rath:
@wolfy1339 all the others I looked at (which is all the "Typescript" repos via the Github UI) are just using `preset: "ts-jest"`, so should be fine :)

PR #376, opened 2020-12-05
feat(typescript): make `verify` a type-guard

G-Rath:
I think this is a good thing, and hopefully just work, but does mean the min TS version required would be 4.1.

I originally posted this in #372 but felt I might as well just PR it for its own discussion.

### DIFF
--- a/src/verify/index.ts
+++ b/src/verify/index.ts
@@ -1,6 +1,17 @@
 import { timingSafeEqual } from "crypto";
 import { Buffer } from "buffer";
+import { EventTypesPayload } from "../generated/get-webhook-payload-type-from-event";
 import { sign } from "../sign/index";
+
+type WebhookEvents = Exclude<
+  keyof EventTypesPayload,
+  `${string}.${string}` | "errors" | "*"
+>;
+
+type GithubEvent<TName extends WebhookEvents = WebhookEvents> = Omit<
+  EventTypesPayload[TName],
+  "name"
+> & { name: TName };
 
 const getAlgorithm = (signature: string) => {
   return signature.startsWith("sha256=") ? "sha256" : "sha1";
@@ -10,7 +21,7 @@ export function verify(
   secret: string,
   eventPayload: object,
   signature: string
-): boolean {
+): eventPayload is GithubEvent {
   if (!secret || !eventPayload || !signature) {
     throw new TypeError(
       "[@octokit/webhooks] secret, eventPayload & signature required"


src/verify/index.ts
```diff
@@ -1,7 +1,18 @@
 import { timingSafeEqual } from "crypto";
 import { Buffer } from "buffer";
+import { EventTypesPayload } from "../generated/get-webhook-payload-type-from-event";
 import { sign } from "../sign/index";
 
+type WebhookEvents = Exclude<
+  keyof EventTypesPayload,
+  `${string}.${string}` | "errors" | "*"
```
gr2m, line 8: woah I didn't know you can do the template literals with `string` types! Very cool

src/verify/index.ts
```diff
@@ -10,7 +21,7 @@ export function verify(
   secret: string,
   eventPayload: object,
   signature: string
-): boolean {
+): eventPayload is GithubEvent {
```
gr2m, line 24: could you explain how that helps exactly? I'm not familiar with that notion

src/verify/index.ts
G-Rath: Yeah, thats the nice hotness that landed with 4.1 - it only arrived about 2 weeks ago.

I'm actually very happy with this bit of code, because it's my first sensible reason to use it in production code, and it works like a treat 😍

src/verify/index.ts
G-Rath: No problem - type guards allow you to define functions that narrow types based on if they return true or false. We actually now also have assertion guards which are the same except the function throws (they landed a few months ago, and very cool).

So by making this a type-guard, you can do this:

```
const signature = event.headers['X-Hub-Signature'];
const { GITHUB_WEBHOOK_SECRET } = process.env;
const body = JSON.parse(event.body ?? '{}') as object;

if (verify(GITHUB_WEBHOOK_SECRET, body, signature)) {
  console.log('Got an event named', body.name, 'from GitHub');
}
```

In the above, because `verify` is a type guard, TypeScript knows that in the scope of the if statement `body` *has* to be of type `GithubEvent` when that code is running since it only runs when its `true`, and so treats it as such while within the braces of the if statement.

`Array.isArray` is an example of a built-in type guard - that function is defined as `isArray<T>(arg: T | {}): arg is T[]`.

(it's actually a bit bigger, as it has a conditional check to say `"arg is (if T is readonly ? readonly T[] : T[])"`, but I've removed that to keep things simple)

G-Rath: [Here's the code in action, vs the current method](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgUQG4FMB2MAqBPMdAZwAUBDPAGwjIBMAaODKYAMzzgF85WoIQ4AcgACEAMYwIAa2AwA9AHd0AIwAWEaUUEBuAFC6YBdHADqK9dLRYYROAF4UADzGUArrXQAeXXDhT0eBCsKBjY+ISkFNR09D5wAAYAJAhEMCyYAOacAHTJqelZ8XAAPnAAROhQfFBEZSXlAFRlugB8egZGcADisqquylbYnjgAcmQgxuiOMFi0tmZqGlKDNvam5ksrRC1rAPIgst6+K+HE5FQ0tADao+PoALqxvmWYd807AGSIcK8TAFxwW4TLjtDwuMhQYysVyYCTACCYJiVNh4ABMAAo4kR0GJITAAflgJknnB0KEYOdorQARBlAArHEwElEYAZV4wVyQglpIkZXQASgBZOslMucGAth6MD6A3J7TECNSPERDnRyggtDwNPpjP59haCDibHRzBR6MEIDw2Nx6BggkY6s1jAtVtZ7M56EE-L1ht8vgVmCIEEo6Gy1Ay5q6EHgZERwuwPzutHtcEdeGyv3Qzt4-G6sgAEv0vXpfJx9L5jab2BiXda8Sm087LSy2WQOZCvT64v7FcHQ+HI9G4LHSeTExNkw6NenM9m+AIpYXlMW4mWy0A).

You'll see that the first use of `body.name` has an error, whereas the second doesn't:

image

G-Rath:
@gr2m I've just confirmed something I suspected but only just gotten far enough to confirm: `webhooks.js` transforms the event in from GitHub to match the interface of `name, payload` etc.

That makes a lot of what we've been talking about (both on this PR & others) mostly theoretical, since the types I've been trying to consume are not correct for my use-case - ultimately what I'm looking for is types for the payloads, and an easy way to do the verification (I can do it myself, but since this repo provides both types + `verify`, that's a win).

I'd still be interested in talking about how things could be improved, and contributing to that, but it means this PR isn't right.

src/verify/index.ts
gr2m: Perfect, I got it now, makes a lot of sense. Thank you for the thorough explanation!

PR #377, opened 2020-12-06
feat(typescript): type `algorithm` property for `sign` options

G-Rath:
I've done it like this because if you do just `Algorithm` TS won't let you pass it in as just a string - you'll have to always use the `Algorithm` enum.

While this isn't a bad thing (since the enum is exported), it'd be a change for TS users - this way you can pass both the enum and plain strings without being able to pass just any old string.

Another way around this would be to make `Algorithm` a `const` enum, but then that'd break people who are currently importing and using the enum since it would no longer exist at runtime.

These changes could be made later down the line if desired, such as if a major release is done.

### DIFF
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -7,7 +7,7 @@ export enum Algorithm {
 
 type SignOptions = {
   secret: string;
-  algorithm?: string;
+  algorithm?: Algorithm | "sha1" | "sha256";
 };
 
 export function sign(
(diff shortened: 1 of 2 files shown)


gr2m:
If I understand correctly, this will improve the typing of the `sign` function? If that's the case, this is a (typescript) feature, not only a refactor.

gr2m (approved):

PR #380, opened 2020-12-07
fix: move `@pika/plugin-ts-standard-pkg` to `devDependencies`

G-Rath:
This isn't needed at runtime, so doesn't need to be pulled in downstream as a dependency :)


gr2m (approved):
Indeed, thank you!

PR #387, opened 2020-12-11
fix(typescript): `eventPayload` parameter for `verify()` can be `string`

G-Rath:
As per [this comment](https://github.com/octokit/webhooks.js/pull/372#issuecomment-739539034), `eventPayload` can be either a string or an object, but currently it only accepts an `object` type.


gr2m (approved):
Thanks!

PR #388, opened 2020-12-11
fix(typescript): `eventPayload` parameter for `Webhooks#verify` can be `string`

G-Rath:
🤦 I forgot to adjust this type too when doing #387


gr2m (approved):
not your fault, we should have tests in place to catch things like this. If only testing TypeScript wouldn't be so hard :( Thanks for the follow up fix!

PR #390, opened 2020-12-13
Improve types

G-Rath:
This is a work-in-progress - there is already an improvement that should be shippable that I'll pull out into its own PR soon (generating the possible values for `action` on payloads that have that property), but the rest is... being worked on.

This changes the type generation to create a type for each *action* of an event, with all the correctness that implies:

  - action-level events are now typed to the specific payload for that action
    - i.e `check_run.created` has a payload of `EventPayloads.WebhookPayloadCheckRunCreatedAction`
  - event-level events are now typed to the payload of the union of all possible actions for that event
  - actions that don't have examples use `{ action: <action> }` to ensure they're not `any`'d

I've used these new types in my current project, and they seem to be correct - however, there does seem to be a minor but noticeable performance difference in WebStorm that makes me less excited about shipping these. This difference only seems to be for a few features i.e autocompletion seems to be unaffected, but there is a delay of ~3 seconds in highlighting properties for these types when I put my caret on them, and it also seems to have confused some autocompletion logic in a few cases (@anstarovoyt would love to submit this on YouTrack, but not sure the best way)

This has also exploded the number of lines - the file is now 13228 LOC. This is because each action is made up of unique types, meaning theres no type-reuse. I've attempted to hack together some simplification that just replaces the types of each action with an alias to a "similar type" if it exists, which shaved off most of the new lines but has its own drawbacks:

   - it significantly slows the type generation process down,
   - it results in a lot of spaghetti type alises (i.e `type X = A; type A = Y; type Y = Q; ... and so on`),
   - it doesn't use them most sensible names, since it effectively alises everything to the first definition of a particular type.

(This has confirmed that the performance drop does not related to the number of lines).

Ultimately, I've settled on that the best way to do the deduplication would be for `type-writer` to support ["definitions"](https://json-schema.org/understanding-json-schema/structuring.html#id1) - this would let us drop a reference for objects that have the same structure across examples to tell `type-writer` not to generate a unique type each time.

However, the bigger blocker for this is what it's highlighted how many actions are missing examples!

```
No examples exist for code_scanning_alert.appeared_in_branch
No examples exist for code_scanning_alert.closed_by_user
No examples exist for code_scanning_alert.created
No examples exist for code_scanning_alert.fixed
No examples exist for code_scanning_alert.reopened_by_user
No examples exist for deploy_key.deleted
No examples exist for installation.new_permissions_accepted
No examples exist for installation.suspend
No examples exist for installation.unsuspend
No examples exist for issues.closed
[...]

PR #392, opened 2020-12-13
feat(typescript): type `action` property on payloads

G-Rath:
Cherry picked from #390 - this should be fine to land as its a straight-forward improvement.


gr2m (approved):
This is great, thanks a lot!

PR #397, opened 2020-12-15
feat(typescript): deduplicate `Sender` type

G-Rath:
This was a bit of a test to see how easy it would be to do some de-duplication of common types, which looks promising.

The next blocker is needing to implement a way of getting all the property paths for an object since we can't use wildcards in `namedKeyPaths` which means we'd have to manually specify all the things to replace.

This change is based on the fact that [`sender` is always on every request](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#webhook-payload-object-common-properties).

Interesting the `sender` for `WebhookPayloadMarketplacePurchase` has an `email` property instead of `node_id` - I've excluded that for now as `type-writer` doesn't seem to support anyway to do `extends` or `&` :(

### DIFF
--- a/scripts/generate-types.js
+++ b/scripts/generate-types.js
@@ -14,16 +14,24 @@ const doNotEditThisFileDisclaimer = `
 // make edits in scripts/generate-types.js`;
 const eventPayloadsVariable = "EventPayloads";
 
-const generatePayloadType = (typeName) => ({
-  rootTypeName: typeName,
-  namedKeyPaths: {
+const generatePayloadType = (typeName) => {
+  const namedKeyPaths = {
     [`${typeName}.repository`]: "PayloadRepository",
     // This prevents a naming colision between the payload of a `installation_repositories` event
     // and the `repositories` attribute of a `installation` event
     "WebhookPayloadInstallation.repositories":
       "WebhookPayloadInstallation_Repositories",
-  },
-});
+  };
+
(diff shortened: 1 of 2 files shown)


wolfy1339 (dismissed):

G-Rath:
@wolfy1339 that was bad timing on my part 😅 - I just dropped the `Webhook` from the name to match `PayloadRepository` which is also deduplicated in the same way

wolfy1339 (approved):

gr2m (approved):
Thanks! I think before we add more workarounds like this, we should create a proper schema via https://github.com/octokit/webhooks.js/pull/397/files. There are tools to translate a JSON Schema to TypeScript. We could probably reuse schemas from the official OpenAPI Spec from the activity endpoints: https://raw.githubusercontent.com/octokit/openapi/main/generated/api.github.com.json

G-Rath:
@gr2m I agree that's probably going to be the better way to go - I've had decent experience with both writing json schemas & generating types from them, so happy to help with improving the schemas once that PR is landed, and with generating good types from them.

There's a few quick and easy de-duplications at the top level that I think would be good to land but otherwise agree that the time would be better spend focused on the schema route.

That does however require the schemas to be landed - I'd be happy to help finish it off, but not the PR author so what I can do it limited 🙂

wolfy1339:
You can always send another PR for that. You could even re-use the commits from the original PR

PR #401, opened 2020-12-30
Add failing test to verify that "request.setEncoding("utf8")" is needed

### LINKED ISSUE #203: Add failing test to verify that "request.setEncoding("utf8")" is needed
This is a follow up to https://github.com/octokit/webhooks.js/pull/202.

@nogs did some outstanding research work on why we suddenly saw a rise of `signature does not match event payload and secret` errors with the new Probot v10 version: https://github.com/probot/probot/pull/1274#issuecomment-671857820

We couldn't manage to create a test that fails if the `request.setEncoding("utf8")` added in https://github.com/octokit/webhooks.js/pull/202 is missing. If anyone has an idea on how to accomplish this, we'd be very greatful.

It might be worth to investigate other webhook SDKs, such as https://github.com/stripe/stripe-node/blob/master/lib/Webhooks.js

copperwall:
Hey @gr2m 👋 

This adds assertions to the `tests/integration/middleware-test.ts` test cases to make sure that `request.setEncoding('utf8')` is called when the `getPayload` middleware is called. This modifies the RequestMock to use a jest mock for setEncoding.

I also changed the test cases in that file to use the Jest functionality where the test case returns a promise that the test runner will wait to resolve before considering the test case to be done, instead of using the `done` callback method. With the promise approach, the assertions in the middleware's then handler are guaranteed to be called before the test considers itself done, so I believe we can remove the `expect.assertions` call. Also if for some reason the promise returned by the middleware rejects, Jest will be able to handle the rejection instead of it being caught globally. I made that in a separate commit though, so if the refactoring is a little too much I can remove that from the PR too.

Here's an example of the failing tests when `setEncoding` is not called

image

Fixes #203

### DIFF
--- a/test/integration/middleware-test.ts
+++ b/test/integration/middleware-test.ts
@@ -11,6 +11,7 @@ type RequestMock = EventEmitter & {
   method: RequestMethodType;
   headers: { [key: string]: string };
   url: string;
+  setEncoding: jest.Mock<void, string[]>;
 };
 
 const headers = {
@@ -24,7 +25,7 @@ test("Invalid payload", (done) => {
     method: RequestMethodType.POST,
     headers,
     url: "/",
-    setEncoding: function () {},
+    setEncoding: jest.fn(),
   });
 
   const responseMock = {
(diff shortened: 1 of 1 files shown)


gr2m (approved):
This is great work @copperwall, thank you for taking it on! I was hoping to somehow find an example payload that would fail without setting the utf8 encoding, in order to have a full integration test, but this is better than nothing, and makes sure we don't accidentally remove it in future.

PR #403, opened 2021-01-15
refactor: define `body` property on `IncomingMessage`

G-Rath:
This arguments the type of `IncomingMessage` so that `body` is defined, meaning we don't have to use `@ts-expect-error`.

### DIFF
--- a/src/middleware/get-payload.ts
+++ b/src/middleware/get-payload.ts
@@ -4,13 +4,18 @@ import AggregateError from "aggregate-error";
 import { IncomingMessage } from "http";
 import { EventTypesPayload } from "../generated/get-webhook-payload-type-from-event";
 
+declare module "http" {
+  interface IncomingMessage {
+    body?: EventTypesPayload;
+  }
+}
+
 export function getPayload(
   request: IncomingMessage
 ): Promise<EventTypesPayload> {
   // If request.body already exists we can stop here
   // See https://github.com/octokit/webhooks.js/pull/23
 
-  // @ts-expect-error
   if (request.body) return Promise.resolve(request.body);
 
   return new Promise((resolve, reject) => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -2604,30 +2604,11 @@
         "@babel/types": "^7.3.0"
       }
     },
-    "@types/body-parser": {
-      "version": "1.19.0",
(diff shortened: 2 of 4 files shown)


wolfy1339 (dismissed):

G-Rath:
```
Error: node_modules/@types/express-serve-static-core/index.d.ts(219,18): error TS2430: Interface 'Request<P, ResBody, ReqBody, ReqQuery>' incorrectly extends interface 'IncomingMessage'.
  Types of property 'body' are incompatible.
    Type 'ReqBody' is not assignable to type 'EventTypesPayload'.
```

Well thats sad but not entirely unexpected - I'll look into this later to see if I can make TS happy, but probably not worth it tbh

G-Rath:
Except it turns out we don't  need `@types/express` 🎉

wolfy1339 (dismissed):

wolfy1339 (approved):

oscard0m (approved):

src/middleware/get-payload.ts
```diff
@@ -4,13 +4,18 @@ import AggregateError from "aggregate-error";
 import { IncomingMessage } from "http";
 import { EventTypesPayload } from "../generated/get-webhook-payload-type-from-event";
 
+declare module "http" {
+  interface IncomingMessage {
+    body?: EventTypesPayload;
+  }
+}
```
gr2m, line 11: could this become a problem if people use the `IncomingMessage` interface in an app that uses `@octokit/webhooks`? I'd be cautious to pollute global name space? Doesn't that have similar problems to pollution Object prototypes in JavaScript?

G-Rath: > Doesn't that have similar problems to pollution Object prototypes in JavaScript?

Yes and no - ultimately this is about making the types more accurate so it should be fine, but it's usually not something you want to be doing a whole lot for similar reasons to polluting object prototypes.

But because of how TypeScript loads types, it'll only "see" this if you import the file (i.e importing `getPayload`), at which point it becomes structurally accurate since it's possible for `IncomingMessage` to have an optional `body` property on it of a particular type, and so this enables setting that property on such a type.

(You do raise a good point: this is something very much to keep in mind since this is in the "library" space instead of the "app" space, and is why I closed #404 - I'm happy to revert it if it does end up causing trouble for someone)

PR #404, opened 2021-01-15
refactor: define header properties on `IncomingHttpHeaders`

G-Rath:
This arguments the type of `IncomingHttpHeaders` so that the headers are defined, meaning we don't have to cast.


G-Rath:
Actually on second thoughts I think it might be better to keep these as casts since `request` is being passed in (i.e. if this was a `RequestParser` that set these properties, it'd be more appropriate to define them like this).

~Instead, I'll replace them with defined-assertion operations.~ (except they can be arrays, so double nope)

PR #406, opened 2021-01-15
feat: use new types from `@octokit/webhooks-definitions`

### LINKED ISSUE #364: `pull_request.merged` event listener never invoked
**What happened?**

`webhooks.on('pull_request.merged', ...)` is never invoked. Instead, the listener for `pull_request.closed` is called.

**What did you expect to happen?**

I expected the listener attached to `pull_request.merged` be called when a pull request is merged.

**What the problem might be**

`merged` isn't even a valid action on the `pull_request` payload. Instead, a pull request is merged if its `action === 'closed'` and `merged === true`. 

Either a special case should be added to get `pull_request.merged` to work, or it should be removed to avoid this bug.

G-Rath:
Here it is 🎉 

There's a few things to polish off and discuss before landing this, but the bulk of it is done so making the PR now.

* we're missing some events in our schema definitions.
* I've done a big rename of the types so that they're more accurate (have exported the renamed types under their old names too to be less breaking).
* I've improved the types in places so that they're accurate
* have refactored the type generation script into TS - happy to move it back to JS, or to convert the other scripts into TS
  * I tried originally trying to build the type via types, but it's just too hard (if at all possible) - you can get close with 4.1 features, but even then it's a lot of work for TS to do meaning IDE autocompletion suffers vs us just generating a static type
* probably something else I've forgotten to mention 😅 

I'm breaking some things out into their own little PRs where possible.


wolfy1339 (dismissed):

G-Rath:
I think we're good to go - note this'll be rebased at least twice when I merge in #403 & #410, so approves won't stick until then

While I've marked everything in the old generated payloads as `deprecated`, I'd like to delete them asap.
Also, there's probably a couple of redundant type aliases in there that could be cleaned up, but shouldn't be a big deal if that happens later.

wolfy1339 (dismissed):

oscard0m (commented):
I don't feel confident enough to give an approval but, for what I read, everything looks good to me :)

src/generated/webhook-names.ts
```diff
@@ -41,8 +41,8 @@ export const webhookNames = [
   "installation.created",
   "installation.deleted",
   "installation.new_permissions_accepted",
-  "installation.suspend",
-  "installation.unsuspend",
+  "installation.suspended",
```
wolfy1339, line 44: These values are incorrect, I fixed these in octokit/webhooks#300

src/generated/webhook-names.ts
```diff
@@ -41,8 +41,8 @@ export const webhookNames = [
...
+  "installation.unsuspended",
```
wolfy1339: These values are incorrect, I fixed these in [octokit/webhooks#300](https://github.com/octokit/webhooks/pull/300)
https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#installation

src/generated/get-webhook-payload-type-from-event.ts
```diff
@@ -1,215 +1,412 @@
...
+  fork: ForkEvent;
+  github_app_authorization: GithubAppAuthorizationEvent;
+  "github_app_authorization.revoked": GithubAppAuthorizationRevokedEvent;
+  gollum: GollumEvent;
+  installation: InstallationEvent;
+  "installation.created": InstallationCreatedEvent;
+  "installation.deleted": InstallationDeletedEvent;
+  "installation.new_permissions_accepted": InstallationNewPermissionsAcceptedEvent;
+  "installation.suspended": InstallationSuspendedEvent;
+  "installation.unsuspended": InstallationUnsuspendedEvent;
```
wolfy1339: These values are incorrect, I fixed these in octokit/webhooks#300
https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#installation

wolfy1339 (changes requested):

gr2m:
> While I've marked everything in the old generated payloads as `deprecated`, I'd like to delete them asap.

Should these deprecations be part of this PR or did they already happen?

`@octokit/webhooks` has several deprecated APIs, we can cut a breaking release anytime and get rid of them all at once.

G-Rath:
> Should these deprecations be part of this PR or did they already happen?

I've done them in this PR, but can easily break them out into their own PR if you like. I figured it made sense to deprecate in the same PR that lands the replacement.

src/generated/webhook-names.ts
wolfy1339: Fixed in https://github.com/octokit/webhooks/pull/324

src/generated/get-webhook-payload-type-from-event.ts
wolfy1339: Fixed in https://github.com/octokit/webhooks/pull/324

wolfy1339 (dismissed):
Once you update `@octokit/webhooks-definitions` with the required patch it will be good to merge.
Otherwise LGTM

G-Rath:
I've pinned the version constraint for `@octokit/webhooks-definitions` to be an exact version, and updated the `update` workflow to update the constraint with the latest version.

The reason for this is that otherwise it can be updated by anything that regenerates the lockfile, which we've got renovate doing every week, but we need to make sure `generate-types` is run every time the version of this package changes which won't happen so it's PRs will always fail CI.

wolfy1339 (dismissed):

.github/workflows/update.yml
```diff
@@ -11,8 +11,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
       - run: npm ci
-      - run: npm install @octokit/webhooks-definitions@latest
-      - run: npm run update-known-events
+      - run: npm install -ED @octokit/webhooks-definitions@latest
```
gr2m: I'd prefer not to use abbreviations. They are great for typing commands in your terminal, but they are harder to read/understand to others and your future self. Can you use the full flag names instead?

package.json
```diff
@@ -12,12 +12,11 @@
   "scripts": {
     "build": "pika build",
     "coverage": "jest --coverage && open coverage/lcov-report/index.html",
-    "generate-types": "node scripts/generate-types.js",
+    "generate-types": "ts-node -T scripts/generate-types.ts",
```
gr2m: I guess `-T` is `--transpile-only`? Can you write that out please?

src/event-handler/remove-listener.ts
```diff
@@ -1,8 +1,8 @@
-import { State } from "../types";
+import { EmitterEventName, State } from "../types";
 
 export function removeListener(
   state: State,
-  webhookNameOrNames: string | string[],
+  webhookNameOrNames: EmitterEventName | EmitterEventName[],
```
gr2m, line 5: Will this work with `*` and `error`? Related: https://github.com/octokit/webhooks.js/pull/416#pullrequestreview-579913781

gr2m (commented):
My understanding is that we can merge this in as a feature release, as it does not introduce any breaking changes? If so, +1 for merging from me. Otherwise please let me know what the breaking changes are. If breaking changes are TypeScript types only we can release a feature release, I don't consider TypeScript type changes breaking changes as they don't affect production code, they only occur at build time

src/event-handler/remove-listener.ts
G-Rath: Yup! `EmitterEventName` is the `keyof EmitterEventMap`, which is of type `EmitterWebhookEventMap & { error: WebhookEventHandlerError; };`

image

gr2m (dismissed):

wolfy1339 (dismissed):

wolfy1339 (approved):

G-Rath:
Alright, let's do this! we can fix up any issues in post 😎

G-Rath:
> Error: src/event-handler/index.ts(10,17): error TS4058: Return type of exported function has or is using name 'BaseWebhookEvent' from external module "/home/runner/work/webhooks.js/webhooks.js/src/types" but cannot be named.

well, that's odd. I'll see if I can figure out what's going wrong.

G-Rath:
I know why this is breaking - will have a PR up shortly

PR #416, opened 2021-01-29
docs: add the information on how to remove the `onAny()` & `onError()` event handlers

### LINKED ISSUE #415: Add docs for removing `onAny` listener
`onAny` was added by https://github.com/octokit/webhooks.js/pull/324, but there's no documentation on how to remove the listener. In order to do so, you need to call `removeListener()` with `*` as the event name.

wolfy1339:
-----
[View rendered README.md](https://github.com/octokit/webhooks.js/blob/add-docs-remove-onAny-onError/README.md)


G-Rath (approved):

gr2m (approved):
Looks good! I'm not too happy with the API, but don't have a better idea either.  Maybe we should add methods like `.removeAnyListener(handler)` and `.removeErrorListener(handler)`?

We should add tests for using `.removeListener` with `*` and `error`, I'll make a follow up issue

PR #417, opened 2021-01-29
chore: move `jest` config into `jest.config.ts`

G-Rath:
Using `jest.config.ts` gives us a few advantages:

  * strict config checking, which shows up in your editor rather than having to run `jest`
  * reduces the size of published package a bit, since the config is no longer in the `package.json`
  * lets us use javascript features that make things safer/easier/clearer etc (i.e `/.../u.source` for regexps)

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,0 +1,25 @@
+import { Config } from "@jest/types";
+import "ts-jest";
+
+const config: Config.InitialOptions = {
+  coverageThreshold: {
+    global: {
+      statements: 100,
+      branches: 100,
+      functions: 100,
+      lines: 100,
+    },
+  },
+  globals: {
+    "ts-jest": {
+      tsconfig: {
+        esModuleInterop: true,
+      },
+    },
+  },
+  preset: "ts-jest",
+  testEnvironment: "node",
+  testRegex: /test\/.*\/.*.test.ts/u.source,
+};
+
+export default config;
--- a/package-lock.json
+++ b/package-lock.json
@@ -2511,6 +2511,12 @@
         "picomatch": "^2.0.4"
       }
     },
+    "arg": {
+      "version": "4.1.3",
+      "resolved": "https://registry.npmjs.org/arg/-/arg-4.1.3.tgz",
+      "integrity": "sha512-58S9QDqG0Xx27YwPSt9fJxivjYl432YCwfDMfZ+71RAqUrZef7LrKQZ3LHLOwCS4FLNBplP533Zx895SeOCHvA==",
+      "dev": true
+    },
     "argparse": {
       "version": "1.0.10",
       "resolved": "https://registry.npmjs.org/argparse/-/argparse-1.0.10.tgz",
@@ -3778,6 +3784,12 @@
         "parse-json": "^4.0.0"
       }
     },
+    "create-require": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/create-require/-/create-require-1.1.1.tgz",
+      "integrity": "sha512-dcKFX3jn0MpIaXjisoRvexIJVEKzaq7z2rZKxf+MSr9TkdmHmsU4m2lcLojrj/FHl8mk5VxMmYA+ftRkP/3oKQ==",
+      "dev": true
+    },
     "cross-spawn": {
       "version": "7.0.3",
       "resolved": "https://registry.npmjs.org/cross-spawn/-/cross-spawn-7.0.3.tgz",
@@ -4022,6 +4034,12 @@
       "version": "3.1.0",
       "resolved": "https://registry.npmjs.org/detect-newline/-/detect-newline-3.1.0.tgz",
       "integrity": "sha512-TLz+x/vEXm/Y7P7wn1EJFNLxYpUD4TgMosxY6fAVJUnJMbupHBOncxyWUG9OpTaH9EBD7uFI5LfEgmMOc54DsA==",
+      "dev": true
+    },
+    "diff": {
+      "version": "4.0.2",
+      "resolved": "https://registry.npmjs.org/diff/-/diff-4.0.2.tgz",
+      "integrity": "sha512-58lmxKSA4BNyLz+HHMUzlOEpg09FV+ev6ZMe3vJihgdxzgcwZ8VoEEPmALCZG9LmqfVoNMMKpttIYTVG6uDY7A==",
       "dev": true
     },
     "diff-sequences": {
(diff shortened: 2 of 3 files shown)


gr2m:
I'm not a big fan of adding dotfiles/config files to the repository root, but that point might be worth the added clutter:

> strict config checking, which shows up in your editor rather than having to run `jest`

Few questions:

- Does the config file need to be in the project root in order to work in editors?
- Can we create a central repository with the config and reference it as a base line across the @octokit repositories? Would that still work with editors?

G-Rath:
> I'm not a big fan of adding dotfiles/config files to the repository root

I'm curious to why you're not a fan of them.

I find having explicit configuration files makes it easier to know what technologies a repo is using, and less hunting around for the configuration of a tool when you need to make changes or want to know what "modes' are being used.

It also can be an advantage for tooling as it can be easier to load and find (i.e I'm one of the core maintainers for `eslint-plugin-jset` - there exists a future where our rules will look at your config and use that to adjust settings; this is a lot harder to support if configured via ` package.json` as config files can just be `require`'d and things like extending packages "just works").

It also means we're not bloating the `package.json` - while the size is not a lot in the schema of things, it does mean changes to the configuration mean changes to the published package, increasing the diff delta between versions (also not a huge deal relative to other stuff, but still the smaller the diff the easier it is to show auditors etc)

(another way to put it is that I find it easier to navigate the clutter on the file system than in `package.json`)

> Does the config file need to be in the project root in order to work in editors?

No - editors themselves don't actually the config file; the advantage is gained by it just being TypeScript so editors just apply their standard TS file handling and TS does the rest.

> Can we create a central repository with the config and reference it as a base line across the @octokit repositories?

You could by just regular `require` (iirc jest doesn't let you point to a package in the same way tools like `prettier`, `eslint`, etc do), but I wouldn't recommend it because that'll likely cause a decent amount of pain since the configs are not "one-size-fits-all" due to the nature of testing (i.e what a repo actually tests will be reasonably different per repo, so they'll all likely have a bit of uniqueness to their test config & setup)

gr2m:
> I'm curious to why you're not a fan of them.
> 
> I find having explicit configuration files makes it easier to know what technologies a repo is using, and less hunting around for the configuration of a tool when you need to make changes or want to know what "modes' are being used.

That's a valid point. I don't like adding them if I can avoid them. I try to keep new contributors in mind, having repositories with lots of configuration files feel overwhelming. This is very subjective of course.

I should add that I don't like configuration in the first place :D I'm a big zero-configuration fan. I used to configure everything to my liking, but I realized that it's counterproductive when working in teams. For example, I don't like the defaults of `prettier`, but I no longer care, the value of using the default prettier settings are just too good.

I feel like the configuration of `jest` feels like a default configuration, it's as minimal as it can be to work with our environment. But if you think that contributors are confused by some of the configuration and won't be able to find the settings, and/or the editors will pick up the configuration and provide some added intelliSense, then let's do it.

>     * strict config checking, which shows up in your editor rather than having to run `jest`

Thinking of it, can you share what the strict config checking would look like that would show up in the editor given our minimal configuration?

G-Rath:
Sure, here's an example of one that was already happening with the `testMatch` property:

image

It technically only takes `string[] | undefined`, but we were passing `null` - it didn't error because `jest` is currently using a truthy check, but if they were to switch to a form of strict checking in the future that'd break our config.

Likewise with `ts-jest`, which arguments `@jest/types` with its properties:

```
declare module '@jest/types' {
    namespace Config {
        interface ConfigGlobals {
            'ts-jest': TsJestGlobalOptions;
        }
    }
}
```

(this is why we're doing `import "ts-jest";`, to tell TS to "open" that file which means it then sees the interface merging)

If `ts-jest` changes an option that affects our config we'll be told by TypeScript (and also we get autocompletion which is nice).
This is already actually the case: `tsconfig` was renamed to `tsConfig` but not yet removed - while TS can't express the deprecation, it can tell us if/when the deprecated property is removed.

Finally another very strong advantage is that we can use comments, which can be used to explain why we've enabled a particular option or doing something a particular way.

> But if you think that contributors are confused by some of the configuration and won't be able to find the settings

I'm not strongly attached to it (despite what it might seem 😅), but do feel it makes for a better experience - it's very easy for `package.json` to become overloaded with properties for configuring everything (which is why some tooling has started to deprecate loading from `package.json`), and that by having explicit configs like this it helps contributors find their way around by making it clear the tools being used.

tbh I think its a case of which is the lesser evil, as the configs are there either in files on disk or all together in `package.json` 🤷

gr2m:
> which is why some tooling has started to deprecate loading from `package.json`

which ones? I haven't seen that happening

gr2m (dismissed):
thanks for bearing with me, you convinced me 👍🏼

G-Rath:
> which ones? I haven't seen that happening

sorry "deprecating" is a bit strong, but I've definitely seen some movement: `eslint` has (/ is in the process of) deprecated all their configs in favor of `eslint.config.js`, which includes [`package.json`](https://github.com/eslint/rfcs/tree/master/designs/2019-config-simplification#how-does-this-affect-configuration-via-packagejson) - I also want to say `husky` too. Some also recommend "not package.json" (i.e (`babel`)[https://babeljs.io/docs/en/configuration#whats-your-use-case])

> thanks for bearing with me

No problem, always happy to talk things out - you have valid points as well :)

oscard0m:
> > Can we create a central repository with the config and reference it as a base line across the @octokit repositories?
> 
> You could by just regular `require` (iirc jest doesn't let you point to a package in the same way tools like `prettier`, `eslint`, etc do), but I wouldn't recommend it because that'll likely cause a decent amount of pain since the configs are not "one-size-fits-all" due to the nature of testing (i.e what a repo actually tests will be reasonably different per repo, so they'll all likely have a bit of uniqueness to their test config & setup)

Sorry if I'm late to this discussion. What I like about centralizing the configurations (jest, tsconfig, renovate config...) is that with only one commit, you can propagate the changes to all Octokit ecosystem.

Do you want me to check existing `jest` configs and we evaluate if it could have sense to centralize it? Or maybe you have in mind a different way to maintain the configuration across repos.

@gr2m @G-Rath

oscard0m (approved):

G-Rath:
@oscard0m sorry, I rebased 😅 

> Do you want me to check existing jest configs and we evaluate if it could have sense to centralize it?

You've already quoted my thoughts on centralizing jest configs :)

A lot of it boils down to the fact that most of the time you'll always have a config file of some sort, even if it's just configuring the tool in question to use your centralized config, and so think there is a trap you can fall into similar to DRY vs WET code where you centralize too much, resulting in config that's hard to work with or effectively utilize.

(I've rewritten this a few times without being able to come up with a write-up that I think introduces confusion, so instead I'm keeping this short and sweet since I'm at work - always happy for people to investigate, but for `jest` in particular I suspect there won't be huge amount of code to centralize + having to take into account jest versions, upgrades, etc would make it not worth it)

oscard0m:
> @oscard0m sorry, I rebased 😅

No problem :)

> (I've rewritten this a few times without being able to come up with a write-up that I think introduces confusion, so instead I'm keeping this short and sweet since I'm at work - always happy for people to investigate, but for `jest` in particular I suspect there won't be huge amount of code to centralize + having to take into account jest versions, upgrades, etc would make it not worth it)

Let's go with decentralized configs and, if changing configs across repos becomes a recurrent task, we re-discuss the centralization of jest configs then :)

PR #429, opened 2021-02-02
test(ts-validate): add coverage to return type of 'createEventhHandler()'

### LINKED ISSUE #426: Improve CI to detect bad internal typing in the project
> > Is it possible (maybe in a different PR) to add test coverage to prevent this breaking change in the future?
> 
> Totally - so a quick solution would be to just do `npm run build` as part of our PR checks, to ensure building is always successful, but we can also use [`@typescript-eslint/explicit-module-boundary-types`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md?rgh-link-date=2021-02-02T19%3A35%3A00Z) to ensure all exports are completely typed.
> 
> That's something I think should be addressed in another PR.

I can create an issue for this too, and we discuss if `npm run build` as a Check Run would be the solution to go. Thumbs up if you are ok with it

_Originally posted by @oscard0m in https://github.com/octokit/webhooks.js/issues/422#issuecomment-771928734_

oscard0m:
## 📝 Summary

add coverage to return type of 'createEventhHandler()'

## ⛱ Motivation and Context

Fix #426

### DIFF
--- a/test/typescript-validate.ts
+++ b/test/typescript-validate.ts
@@ -103,7 +103,12 @@ export default async function () {
     secret: "bleh",
   });
 
-  createEventHandler({ secret: "bleh" });
+  const eventHandler = createEventHandler({ secret: "bleh" });
+  console.log(eventHandler.on);
+  console.log(eventHandler.onAny);
+  console.log(eventHandler.onError);
+  console.log(eventHandler.removeListener);
+  console.log(eventHandler.receive);
 
   createMiddleware({
     secret: "mysecret",


gr2m:
blocked by https://github.com/octokit/webhooks.js/issues/426#issuecomment-772035786 @G-Rath

wolfy1339:
Shall we close this as well since #426 was closed?

gr2m:
yeah, let's 👍🏼

Issue #431, opened 2021-02-02
event.name type/value disagreement


jablko:
I get `TS2367: This condition will always return 'false' since the types '"check_run.completed" | "code_scanning_alert.fixed"' and '"check_run"' have no overlap.` from the following since 7.22.0/#406:
```TypeScript
on(["check_run.completed", "code_scanning_alert.fixed"], (event) => {
  if (event.name === "check_run") {
    console.log(`a run was ${event.payload.action}!`);
  }
});
```

I confirmed this wasn't an error in 7.21.0.

Thanks for the work in #406 to upgrade the types! :heart:

G-Rath:
This is because you'll never get an event whose name is `check_run` - you'll only ever get `check_run.completed` or `code_scanning_alert.fixed`.

jablko:
The docs describe using the X-GitHub-Event header https://github.com/octokit/webhooks.js#user-content-webhooksverifyandreceive:~:text=name%3A%20request.headers%5B%22x%2Dgithub%2Devent%22%5D%2C

which is what `webhooks.middleware()` does https://github.com/octokit/webhooks.js/blob/56f81470e63ea9655ef4c6d4501f14a5cd24267b/src/middleware/middleware.ts#L48

and in the GitHub docs that header doesn't include the action https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#example-delivery:~:text=%3E%20X%2DGitHub%2DEvent%3A%20issues

What am I missing?

G-Rath:
> The docs describe using the X-GitHub-Event

It says that for `webhooks.receive`, so it's correct.

However the docs for `webhooks.on` are misleading as they say:

> Required. Name of the event. One of [GitHub's supported event names.](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads)

Really it should say something like "One of Githubs supported event names, or (if the event has an `action` property) the name of an event followed by its action in form of `<event>.<action>`", and link [here](https://github.com/octokit/webhooks.js#webhook-events).

jablko:
I can call `webhooks.on()` on "check_run.completed" but my handler is still getting the X-GitHub-Event header value (in this case "check_run"):
```TypeScript
const { Webhooks } = require("@octokit/webhooks");
const webhooks = new Webhooks({
  secret: "mysecret",
});
webhooks.on("check_run.completed", (event) => console.log(event.name)); // -> "check_run"
webhooks.receive({
[...]

PR #434, opened 2021-02-03
fix(types): accept string event.name for receive()

jablko:
Since #428 I get the following type error:
```TypeScript
webhooks.receive({
  id: request.headers["x-github-delivery"],
  name: request.headers["x-github-event"],
  ^^^^ TS2322: Type 'string' is not assignable to type '"status" | "delete" | "package" | "installation" | "repository" | "team" | "member" | "project" | "public" | "push" | "deployment" | "label" | "milestone" | "release" | "organization" | ... 187 more ... | "workflow_run.requested"'.
  payload: request.body,
});
```

Instead of changing `webhooks.receive()` to match `eventHandler.receive()` (as in #428), can we do the reverse and change `eventHandler.receive()` to match `webhooks.receive()` (as in this PR)?

I additionally added `| undefined` because:
1. It's checked in `.receive()` https://github.com/octokit/webhooks.js/blob/b7991135bbc10e759320094353946127f04d2079/src/event-handler/receive.ts#L46-L48
2. It avoids having to either non-null assert `request.headers["x-github-event"]!` or test it before calling `.receive()` (redundant since it's checked again by `.receive()`) when running with `--noUncheckedIndexedAccess`

### DIFF
--- a/src/event-handler/receive.ts
+++ b/src/event-handler/receive.ts
@@ -1,8 +1,6 @@
 // @ts-ignore to address #245
 import AggregateError from "aggregate-error";
-import { EmitterEventWebhookPayloadMap } from "../generated/get-webhook-payload-type-from-event";
 import {
-  EmitterEventName,
   EmitterWebhookEvent,
   State,
   WebhookError,
@@ -10,15 +8,10 @@ import {
 } from "../types";
 import { wrapErrorHandler } from "./wrap-error-handler";
 
-type EventAction = Extract<
-  EmitterEventWebhookPayloadMap[keyof EmitterEventWebhookPayloadMap],
-  { action: string }
->["action"];
-
 function getHooks(
   state: State,
-  eventPayloadAction: EventAction | null,
-  eventName: EmitterEventName
+  eventPayloadAction: unknown,
+  eventName: string
 ): Function[] {
   const hooks = [state.hooks[eventName], state.hooks["*"]];
 
@@ -30,7 +23,14 @@ function getHooks(
 }
 
 // main handler function
-export function receiverHandle(state: State, event: EmitterWebhookEvent) {
+export function receiverHandle(
+  state: State,
+  event: {
+    id: string | undefined;
+    name: string | undefined;
+    payload: unknown;
+  }
+) {
   const errorHandlers = state.hooks.error || [];
 
   if (event instanceof Error) {
@@ -54,7 +54,9 @@ export function receiverHandle(state: State, event: EmitterWebhookEvent) {
   // flatten arrays of event listeners and remove undefined values
   const hooks = getHooks(
     state,
-    "action" in event.payload ? event.payload.action : null,
+    typeof event.payload === "object" && "action" in event.payload!
+      ? (event.payload as { action: unknown }).action
+      : null,
     event.name
   );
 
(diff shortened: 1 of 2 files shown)


wolfy1339:
I disagree with you, I think you should use this code

```ts
webhooks.receive({
  id: request.headers["x-github-delivery"],
  name: request.headers["x-github-event"] as EmitterWebhookEventName,
  payload: request.body,
});
```

G-Rath:
> It's checked in .receive()

That's a runtime check targeted as JS users - allowing `undefined` defeats the benefit that you get by using TypeScript, since you'll not be told at compile-time if the value you're passing in could be `undefined`.

jablko:
@G-Rath Okay, I've removed the `| undefined` addition from this PR.

@wolfy1339 `EmitterWebhookEventName` seems like a postcondition of `.receive()` (it checks `event.name` against a list of hooks before running the matching handlers) vs. a precondition? What am I missing?

G-Rath:
> What am I missing?

Pretty much the same thing as with the `| undefined` - it's a runtime check that we do to warn people when they're passing in an event name we don't know about, but if they're using TypeScript we can tell them about this at compile-time rather than as runtime.

Again as with `| undefined`, your change would mean I'd not know if my variable is valid.

(Also note that we're *warning* not erroring for this check)

jablko:
`name: string` hides the fact that `event.name` might be bogus from the caller, whereas `name: EmitterWebhookEventName` hides that fact from the `.receive()` implementation (if it might be bogus?). Which side is responsible for confirming that `event.name` is in fact an `EmitterWebhookEventName`?

[The examples in the docs](https://github.com/octokit/webhooks.js#user-content-webhooksverifyandreceive:~:text=name%3A%20request.headers%5B%22x%2Dgithub%2Devent%22%5D%2C) had me thinking that e.g. `webhooks.on(["check_run", "code_scanning_alert"], handler)` was a convenient way to validate `request.headers["x-github-event"]`.

G-Rath:
> Which side is responsible for confirming that event.name is in fact an EmitterWebhookEventName?

Neither side is - that responsibility lies with the caller of the function (aka whoevers consuming the dependency). 

The example in the documentation is using the `x-github-event` header from a request, which Github sets accordingly and so we expect to be correct. If you're passing in values from other sources, then the responsibility falls on you to ensure they're correct.

We do some basic runtime checking for our JS users for the things that are cheap and easy to check, but we only warn on the actual event names because we don't want to be too restrictive - otherwise the library would be unusable if we were ever missing an event (i.e [a bug/typo in our schemas](https://github.com/octokit/webhooks.js/pull/406#discussion_r563401037), using a fancy new event that we'd not implemented schemas for, or even maybe custom endpoints or prototyping because you're working at Github on implementing that fancy new event).

jablko:
> The example in the documentation is using the `x-github-event` header from a request, which Github sets accordingly and so we expect to be correct. If you're passing in values from other sources, then the responsibility falls on you to ensure they're correct.

Isn't `webhooks.on(["check_run", "code_scanning_alert"], handler)` a convenient way to do that?

G-Rath:
That would require us to error instead of warn on event names, which we don't want to do because we don't want to be too restrictive.

There are pros and cons to both approaches, but this library has gone the route of warning meaning at the very least changing it to error would be a breaking change.

G-Rath:
🤦 I used the wrong issue number, sorry - I meant to close #431 with that PR.

jablko:
I see my mistake: if `webhooks.onAny()` is going to be `handler: (event: EmitterAnyEvent) => any` vs. `handler: (event: { name: string, payload: unknown }) => any` then the input needs to be correspondingly typed. Thanks for your patience and sorry for the noise!

PR #435, opened 2021-02-03
fix(typescript): resolve correct value for `BaseWebhookEvent#name`

### LINKED ISSUE #431: event.name type/value disagreement

G-Rath:
This was actually a lot easier than I expected, but it makes total sense in hindsight.

This does put our minimal TypeScript version at 4.1 or higher - personally I don't have a problem with this, especially since we consider TypeScript changes to be non-breaking, but worth mentioning.

Fixes #431

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -23,11 +23,17 @@ export type EmitterEvent = EmitterEventMap[EmitterEventName];
 
 export type EmitterAnyEvent = EmitterWebhookEventMap["*"];
 
+export type ToWebhookEvent<
+  TEmitterEvent extends string
+> = TEmitterEvent extends `${infer TWebhookEvent}.${string}`
+  ? TWebhookEvent
+  : TEmitterEvent;
+
 interface BaseWebhookEvent<
   TName extends keyof EmitterEventPayloadMap = keyof EmitterEventPayloadMap
 > {
   id: string;
-  name: TName;
+  name: ToWebhookEvent<TName>;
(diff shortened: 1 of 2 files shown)


oscard0m (dismissed):

wolfy1339 (dismissed):

wolfy1339 (approved):

gr2m (approved):
love the TypeScript tests, very cool!

oscard0m (approved):

G-Rath:
> love the TypeScript tests, very cool!

Yeah they're actually really nice - I think they're the way to go for testing typescript in other repos.

-----

Since everyones approved, I'm guessing we're happy requiring >= 4.1 for Typescript - I'll merge this in an hour or so in case anyone wants to change their mind, but otherwise will 🚀 on

gr2m:
> happy requiring >= 4.1 for Typescript

yes. And if anyone complaints, send them my way 😜

PR #438, opened 2021-02-04
feat: add `removeAllListeners` function

### LINKED ISSUE #437: Add ability to remove *all* listeners for an event
**What’s missing?**

Currently, if you use `removeListener()`, you must provide an event and a handler, there is no way to remove all listeners for an event

**Why?**

In my integration tests, all of my listeners are loaded before every test, I'd like to clear all event listeners for a specific event before the test.

**Alternatives you tried**

From looking at the source, there doesn't seem to be a way to do this. I can work around this with my setup

wolfy1339:
Resolves #437

### DIFF
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -13,7 +13,7 @@ import {
   receiverOnError as onError,
 } from "./on";
 import { receiverHandle as receive } from "./receive";
-import { removeListener, removeListeners } from "./remove-listener";
+import { removeListener, removeAllListeners } from "./remove-listener";
 
 interface EventHandler<TTransformed = unknown> {
   on<E extends EmitterEventName>(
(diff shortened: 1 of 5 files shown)


wolfy1339:
I'm not sure why the coverage dropped since I didn't touch the tests for the `on()` function

G-Rath:
I think a better name for this would be `removeAllListeners`, as that matches with the native [`EventEmitter`](https://nodejs.org/api/events.html#events_emitter_removealllisteners_eventname) API in Node.

G-Rath:
>   receive.ts             |   97.22 |      100 |    87.5 |     100 |                   

tbh that looks like a bug somewhere to me, since it doesn't report what's actually uncovered.

gr2m:
Let's discuss if we need this at all first. It never came up before, and it seems the use case here is testing, not use in production. I think the real solution here is to change the tests

PR #441, opened 2021-02-04
🚧 refactor(types): replace EmitterEventWebhookPayloadMap

jablko:
Can we eliminate `get-webhook-payload-type-from-event.ts` thanks to #435?
- Replace `EmitterEventWebhookPayloadMap[TName]` with `EventPayloadMap[TWebhookEvent] & { action: TAction }`
- Replace `keyof EmitterEventWebhookPayloadMap` with `typeof emitterEventNames[number]`


G-Rath:
I had originally laid the types out like this (or similar to this), but decided against it in favor of how it is currently as it's easier to predict the behaviour (both in typings and in performance).

scripts/generate-types.ts
```diff
@@ -196,7 +182,7 @@ const run = () => {
     '"*",',
     '"error",',
     ...properties.map(([key]) => `"${key}",`),
-    "];",
+    "] as const;",
```
G-Rath, line 185:
```suggestion
    "]",
```

This doesn't need to be `const`.

src/event-handler/receive.ts
```diff
@@ -11,8 +10,8 @@ import {
 import { wrapErrorHandler } from "./wrap-error-handler";
 
 type EventAction = Extract<
-  EmitterEventWebhookPayloadMap[keyof EmitterEventWebhookPayloadMap],
-  { action: string }
+  EmitterWebhookEvent["payload"],
+  { action: any }
```
G-Rath, line 14:
```suggestion
  { action: string }
```

G-Rath:
~I've not checked out this out locally, but I feel like it's not preserving the mapping of the event + action to the specific type for that action. We're probably not got any tests covering this as there are some subtle differences in the events.~

~I'll push one up so we can see.~

Nope it'll be fine because that's handled by the types for each payload being a union :facepalm:

scripts/generate-types.ts
jablko: Otherwise `typeof emitterEventNames[number]` is `string` vs. `"push" | "error" | "public" | "label" | "meta" | "fork" | "ping" | "repository" | "milestone" | "member" | "team" | "status" | "installation" | "project" | "deployment" | "release" | ... 189 more ... | "workflow_run.requested"`

G-Rath: sorry that's my bad - I thought you were `const`ing the array.

src/types.ts
```diff
@@ -1,11 +1,21 @@
...
 
-type EmitterEventPayloadMap = { "*": Schema } & EmitterEventWebhookPayloadMap;
+type EmitterEventPayloadMap = { "*": Schema } & EventPayloadMap;
 
 export type EmitterWebhookEventMap = {
-  [K in keyof EmitterEventPayloadMap]: BaseWebhookEvent<K>;
+  [K in Exclude<
+    typeof emitterEventNames[number],
+    "error"
+  >]: K extends `${infer TWebhookEvent}.${infer TAction}`
```
G-Rath: Can you please make this a "type function"? i.e.

```
type WithAction<...> = K extends ... // etc
```

src/types.ts
```diff
@@ -1,11 +1,21 @@
...
+} from "@octokit/webhooks-definitions/schema";
+import type { emitterEventNames } from "./generated/webhook-names";
 
-type EmitterEventPayloadMap = { "*": Schema } & EmitterEventWebhookPayloadMap;
+type EmitterEventPayloadMap = { "*": Schema } & EventPayloadMap;
 
 export type EmitterWebhookEventMap = {
-  [K in keyof EmitterEventPayloadMap]: BaseWebhookEvent<K>;
+  [K in Exclude<
+    typeof emitterEventNames[number],
```
G-Rath: I think let's make this a type alias in the generated type:

```
export type EmitterEventName = typeof emitterEventNames[number];
```

G-Rath (changes requested):
I'm going to get started on a `next` branch, which'll likely include this change so going to hold off merging it for now.

G-Rath:
Just FYI, as I mentioned in #444: this change causes *massive* performance problems in the current version of WebStorm - while TypeScript seems to be able to handle it fine, I don't think there's a huge gain to be had that makes it worth putting such a massive blocker on anyone using that IDE.

I've submitted an issue on YouTrack and pinged the team, so hopefully it'll get fixed promptly (but even once fixed we'd want to wait for the release, which'd be a few months later).

jablko:
I wonder if removing `EmitterWebhookEventMap`/`EmitterEventMap` would make a difference? When you get a chance, would you mind seeing whether the latest revision makes a difference in WebStorm, one way or the other?

G-Rath:
That seems to have fixed both this and the performance problems caused by inferring the impact of `transform` in #440 🎉 

Currently I'm on the fence about shipping this because of the huge impact if I'm wrong, but def going to keep this patch around to apply it at a later date:

<details>
<summary>
patch
</summary>

```
Index: src/types.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/types.ts b/src/types.ts
--- a/src/types.ts	(revision 560ff73ef02453fd0a28705e042e8ad3538c43ef)
+++ b/src/types.ts	(date 1612557132345)
@@ -1,25 +1,33 @@
-import type { RequestError } from "@octokit/request-error";
-import type { EmitterEventWebhookPayloadMap } from "./generated/get-webhook-payload-type-from-event";
+import { RequestError } from '@octokit/request-error';
+import type {
+  EventPayloadMap,
+  Schema,
+  WebhookEventName,
+} from "@octokit/webhooks-definitions/schema";
+import type { EmitterEventName } from "./generated/webhook-names";
+
+export { EmitterEventName };
 
-export type EmitterWebhookEventMap = {
-  [K in keyof EmitterEventWebhookPayloadMap]: BaseWebhookEvent<K>;
-};
+type EmitterEventPayloadMap = { "*": Schema } & EventPayloadMap;
 
-export type EmitterWebhookEventName = keyof EmitterWebhookEventMap;
-export type EmitterWebhookEvent = EmitterWebhookEventMap[EmitterWebhookEventName];
+export type EmitterWebhookEventName = EmitterEventName;
 
-type ToWebhookEvent<
-  TEmitterEvent extends string
-> = TEmitterEvent extends `${infer TWebhookEvent}.${string}`
-  ? TWebhookEvent
-  : TEmitterEvent;
+export type EmitterWebhookEvent<
+  TEmitterEvent = WebhookEventName
+> = TEmitterEvent extends `${infer TWebhookEvent}.${infer TAction}`
+  ? BaseWebhookEvent<Extract<TWebhookEvent, keyof EmitterEventPayloadMap>> & {
+      payload: { action: TAction };
+    }
+  : BaseWebhookEvent<Extract<TEmitterEvent, keyof EmitterEventPayloadMap>>;
 
-interface BaseWebhookEvent<
-  TName extends EmitterWebhookEventName = EmitterWebhookEventName
-> {
+type EmitterEvent<TEmitterEvent> = TEmitterEvent extends "error"
+  ? WebhookEventHandlerError
+  : EmitterWebhookEvent<TEmitterEvent>;
+
+interface BaseWebhookEvent<TName extends keyof EmitterEventPayloadMap> {
   id: string;
-  name: ToWebhookEvent<TName>;
-  payload: EmitterEventWebhookPayloadMap[TName];
+  name: TName;
+  payload: EmitterEventPayloadMap[TName];
 }
 
 export interface Options<
@@ -39,15 +47,9 @@
 // type MaybeArray<T> = T | T[];
 
 export type HandlerFunction<
-  TName extends EmitterWebhookEventName | EmitterWebhookEventName[],
+  TName extends EmitterEventName | EmitterEventName[],
   TTransformed
-> = (
-  event: EmitterWebhookEventMap[Extract<
-    EmitterWebhookEventName,
-    EnsureArray<TName>[number]
-  >] &
-    TTransformed
-) => any;
+> = (event: EmitterEvent<EnsureArray<TName>[number]> & TTransformed) => any;
 
 type Hooks = {
   [key: string]: Function[];
Index: test/typescript-validate.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/test/typescript-validate.ts b/test/typescript-validate.ts
--- a/test/typescript-validate.ts	(revision 560ff73ef02453fd0a28705e042e8ad3538c43ef)
+++ b/test/typescript-validate.ts	(date 1612557132359)
@@ -9,7 +9,7 @@
   WebhookError,
 } from "../src/index";
 import { createServer } from "http";
-import { HandlerFunction, EmitterWebhookEventName } from "../src/types";
+import { EmitterEventName, HandlerFunction } from '../src/types';
 
 // ************************************************************
 // THIS CODE IS NOT EXECUTED. IT IS FOR TYPECHECKING ONLY
@@ -23,7 +23,7 @@
   }
 };
 
-declare const on: <E extends EmitterWebhookEventName>(
+declare const on: <E extends EmitterEventName>(
   name: E | E[],
   callback: HandlerFunction<E, unknown>
 ) => void;
Index: src/generated/webhook-names.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/generated/webhook-names.ts b/src/generated/webhook-names.ts
--- a/src/generated/webhook-names.ts	(revision 560ff73ef02453fd0a28705e042e8ad3538c43ef)
+++ b/src/generated/webhook-names.ts	(date 1612556741606)
@@ -206,4 +206,6 @@
   "workflow_run",
   "workflow_run.completed",
   "workflow_run.requested",
-];
+] as const;
+
+export type EmitterEventName = typeof emitterEventNames[number];
```
</details>

In saying that, I think it probably has fixed the issue and I'm just being paranoid 😅 
I'll let things sit for a bit so that the WebStorm team has a chance to look into it, and to give myself some time to put the types through their paces a bit more to confirm, but I think this should be fine to ship in the near future! 🎉

(/cc @anstarovoyt, since it might help you pin down where the performance issue is)

oscard0m:
> I've submitted an issue on YouTrack and pinged the team, so hopefully it'll get fixed promptly (but even once fixed we'd want to wait for the release, which'd be a few months later).

Could you link please the issue on YouTrack, please? :) 🙏🏽

G-Rath:
https://youtrack.jetbrains.com/issue/WEB-49483

PR #443, opened 2021-02-04
fix(types): .onAny() event.name

jablko:
Like #435, the `.onAny()` `event.name` is e.g. "check_run", vs. "*"
https://github.com/octokit/webhooks.js/blob/aaec66e85d8e33cd5352152ed715af652be25d9d/test/integration/event-handler-test.ts#L30
https://github.com/octokit/webhooks.js/blob/aaec66e85d8e33cd5352152ed715af652be25d9d/test/integration/event-handler-test.ts#L65
https://github.com/octokit/webhooks.js/blob/aaec66e85d8e33cd5352152ed715af652be25d9d/test/integration/event-handler-test.ts#L69

Thanks for #435!

### DIFF
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -1,5 +1,4 @@
 import type {
-  EmitterAnyEvent,
   EmitterEventName,
   EmitterWebhookEvent,
   HandlerFunction,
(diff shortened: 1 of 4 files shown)


G-Rath (approved):
@gr2m I'm thinking it might be a good time to just cut a new major release, dropping the old types and the `error` & `*` events as that'd let us do some good simplification to the types and make things less confusing (at least for me 😅)

gr2m:
> @gr2m I'm thinking it might be a good time to just cut a new major release, dropping the old types and the `error` & `*` events as that'd let us do some good simplification to the types and make things less confusing (at least for me 😅)

Yes, agreed

G-Rath (changes requested):
I'm going to get started on a `next` branch, which'll likely include this change so going to hold off merging it for now.

PR #444, opened 2021-02-05
v8.0.0

### LINKED ISSUE #445: TypeScript section in Readme out of date?
I noticed that there are currently a bunch of issues open about types issues and also a few PRs in flight but afaict, none of them seem to address this.

It seems to me that the currently published version of `@octokit/webhooks` does not actually export the types specified in https://github.com/octokit/webhooks.js#typescript, those docs seem to reference some of the older types which are now partially deprecated, too.
I'd love to help with updating but not sure I properly understand how the currently exported types are to be used 🤔

G-Rath:
I've done this all in one PR as there's some overlapping changes and I think it's just easier to review as a whole. There are a few changes I'll look to pull out (specifically the test & typo fixings).

I think this is pretty much good to go, but going to make it a draft for now all the same.

Closes #439 
Closes #440
Closes #441
Closes #443
Closes #445
Closes #447

------

I've deliberately *not* applied the main change from #441 as while it's nice not to have the file, it causes *massive* performance issues in my IDE which I think are not worth the trade off since it'll occur for other WebStorm users.

There's also a performance problem when having to infer the generic for `TTransformed`, but that can be worked around by just providing the generic explicitly so so long as I don't have `typescript-validate` open I can still work on the project without issue 😅  

I have opened a ticket in YouTrack, and will hopefully be able to get that resolved.

----

Also, we could revert the type refactoring on 7.x.x so that `probot` and co wouldn't have as a sad time, and then just include that refactor in this release.

-----
[View rendered .github/ISSUE_TEMPLATE/04_feature_request.md](https://github.com/octokit/webhooks.js/blob/next/.github/ISSUE_TEMPLATE/04_feature_request.md)
[View rendered README.md](https://github.com/octokit/webhooks.js/blob/next/README.md)


G-Rath:
@anstarovoyt @prigara the performance problems are pretty massive & showstopping, and it'd be very easy for a package to ship with types that'd grind things to a halt so it'd be great to get this resolved quickly.

Happy to help anyway I can :)

wolfy1339 (dismissed):

gr2m:
just a quick note: because this is a release branch (`next`) which has special meaning to semantic-release, do not do any rebasing and force pushing. And we should merge this PR using <kdb>Rebase and merge</kdb> instead of squashing. `semantic-release` is using git notes to keep track of preview releases. If we would do any squashing or force-pushing we would loose that state

G-Rath:
Yeah, sorry I didn't know it was setup to do a pre-release (which is very cool, and didn't know you could do - might "borrow" that config for some of my other projects) 😓

Any follow up changes I'll do to this branch via PR, but it seems like it's good to go already 🤷 

Are there any other breaking changes you'd talked about doing previously that could be included?

(also @gr2m I'll let you do the merging for this one 🙂)

scripts/generate-types.ts
```diff
@@ -193,8 +194,6 @@ const run = () => {
     "// make edits in scripts/generate-types.ts",
     "",
     "export const emitterEventNames = [",
-    '"*",',
-    '"error",',
```
gr2m, line 197: so glad to be finally rid of this 👍🏼

gr2m (commented):
we can also remove this from the tests I think?

https://github.com/octokit/webhooks.js/blob/2db33e5c11b80cf6e2831afde9392e96803ee267/test/typescript-validate.ts#L199-L206

gr2m:
I wonder, now that we have explicit event handlers `.onAny` and `.onError`, we should probably throw if someone tries to listen to `"*"` or `"error"` events, as we use these strings internally to keep track of the `.onAny`/`.onError` calls. Rather an edge case and it would fail when compiling TypeScript, but might be worth to throw a useful error anyway at least for `v8.x` to help people upgrade

gr2m:
btw next time let's create a `beta` branch for pre-releases, that will create `8.0.0-beta.1`, `8.0.0-beta.2`, etc. Once we think it's ready, we can still go through a `next` branch to have more people test it before flagging it as `@latest` on npm/GitHub releases

G-Rath:
Bamn! I've refactored the types based off #441 (thanks @jablko ❤️), and updated the `typescript` section of the README.

I think this is now good to go, unless anyone has any other changes they think we could ship in this major?

The only other thing off the top of my head that I could do is renaming the types to be consistent - they're mostly already fine now, but we could rename `EmitterWebhookEvent` to `WebhookEvent` as its a bit shorter and now makes more sense? Plus a few others 🤔

src/event-handler/on.ts
```diff
@@ -29,18 +29,20 @@ export function receiverOn(
...
-    );
+  if (["*", "error"].includes(webhookNameOrNames)) {
+    const webhookName =
+      (webhookNameOrNames as string) === "*" ? "any" : webhookNameOrNames;
+
+    const message = `Using the "${webhookNameOrNames}" event with the regular Webhooks.on() function is not supported. Please use the Webhooks.on${
+      webhookName.charAt(0).toUpperCase() + webhookName.slice(1)
+    }() method instead`;
+
+    throw new Error(message);
```
wolfy1339, line 40: Let's get rid of this altogether. It isn't needed anymore

G-Rath: This was added [at the request](https://github.com/octokit/webhooks.js/pull/444#issuecomment-774256384) of @gr2m :)

lumaxis:
This will also fix #445

wolfy1339 (dismissed):

src/types.ts
```diff
@@ -1,64 +1,45 @@
...
-  TName extends keyof EmitterEventPayloadMap = keyof EmitterEventPayloadMap
-> {
+import { RequestError } from "@octokit/request-error";
+import type {
+  WebhookEventMap,
+  WebhookEventName,
+} from "@octokit/webhooks-definitions/schema";
+import type { emitterEventNames } from "./generated/webhook-names";
+
+export type EmitterWebhookEventName = typeof emitterEventNames[number];
```
jablko, line 8: What made you decide to revert the type alias in the generated type? https://github.com/octokit/webhooks.js/pull/441#discussion_r570537115

I don't have a strong opinion, just curious.

jablko:
Can we land #443, #447, and #441, which this branch incorporates, to separate the breaking from non-breaking changes?

src/types.ts
```diff
@@ -1,64 +1,45 @@
...
+export type EmitterWebhookEvent<
+  TEmitterEvent extends EmitterWebhookEventName = EmitterWebhookEventName
```
jablko, line 10:
```suggestion
  TEmitterEvent extends EmitterWebhookEventName = WebhookEventName
```
I don't know if it makes a difference, but in #441 I went with `WebhookEventName` here because `EmitterWebhookEvent<WebhookEventName>` is functionally equivalent to `EmitterWebhookEvent<EmitterWebhookEventName>` and `WebhookEventName` is smaller than `EmitterWebhookEventName` so I thought it might perform better? https://github.com/jablko/webhooks.js/compare/patch-2..octokit:next#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aL14-R10

src/types.ts
```diff
@@ -1,64 +1,45 @@
...
-> = TEmitterEvent extends `${infer TWebhookEvent}.${string}`
-  ? TWebhookEvent
-  : TEmitterEvent;
-
-interface BaseWebhookEvent<
-  TName extends keyof EmitterEventPayloadMap = keyof EmitterEventPayloadMap
-> {
+import { RequestError } from "@octokit/request-error";
+import type {
+  WebhookEventMap,
```
jablko, line 3:
```suggestion
  EventPayloadMap,
```
No strong opinion, but to my eye `EventPayloadMap` makes it clearer that this is a map from `WebhookEventName` -> payload vs. `WebhookEventName` -> event? https://github.com/jablko/webhooks.js/compare/patch-2..octokit:next#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aL3-R3

G-Rath: It's a map of webhook events - the fact that it's called "Payload" is a remnant of previous versions, and at some point will likely be removed.

src/types.ts
G-Rath: I'd prefer to use the most accurate type - if it was causing more performance problems, I'd want to know about them here since you could still run into them in userland by using that type :)

G-Rath:
> Can we land #443, #447, and #441, which this branch incorporates, to separate the breaking from non-breaking changes?

I class all these changes as breaking, because of the size of the refactors, and would just in general like to get v8 landed.
While we could land these in 7.x, if there is further bug fixing to be done we'd not be able to land those once we've landed 8.x, so might as well just go to 8.x.

I don't mind reverting the all the type refactor stuff in 7.x so that people can stay on that longer if they want?

@gr2m while everyones approved this, I'll leave the releasing in your hands since I don't want to mess it up 😅

src/types.ts
G-Rath: I just found it was nicer to have everything being imported from the same place, rather than having an extra import line 😅

jablko, line 8:
```suggestion
export type EmitterEventName = typeof emitterEventNames[number];
```
What do you think of just `EmitterEventName`, now every `EmitterEventName` is an `EmitterWebhookEventName` (`"*"` and `"error"` having been removed)?

G-Rath: [yup](https://github.com/octokit/webhooks.js/pull/444#issuecomment-774778669), but I'm not so concerned about that since its still accurate with the current name, and everyones approved this version 🤷🙂

jablko: Following the pattern of `TEmitterEvent` vs. `TWebhookEvent`: https://github.com/octokit/webhooks.js/blob/fc5277d9b315fe5faad8827c8f0716032ce25b0c/src/types.ts#L11

gr2m:
> > Can we land #443, #447, and #441, which this branch incorporates, to separate the breaking from non-breaking changes?
> 
> I class all these changes as breaking, because of the size of the refactors, and would just in general like to get v8 landed

I agree. Sorry @jablko. The upgrade to v8 shouldn't be too hard.

> @gr2m while everyones approved this, I'll leave the releasing in your hands since I don't want to mess it up 😅

I got it

gr2m:
well the automation didn't work for some reason 🤷🏼 but I've fixed it manually. Version 8.0.2 is now the latest on npm and GitHub

G-Rath:
Feel free to blame that on me 😅 next time I'll not call the branch `next`

oscard0m:
Besides the automatic release not working, how was it? All working and shinning?

PR #448, opened 2021-02-05
fix(docs): Use correct links in subfolder Readme files

lumaxis:
-----
[View rendered src/middleware/README.md](https://github.com/lumaxis/webhooks.js/blob/patch-1/src/middleware/README.md)


src/middleware/README.md
```diff
@@ -16,6 +16,6 @@ require('http').createServer(middleware).listen(3000)
 
 ## API
 
-The `middleware` API implements [`.on()`](../#webhookson) and [`.removeListener()`](../#webhooksremovelistener).
+The `middleware` API implements [`.on()`](https://github.com/octokit/webhooks.js/blob/HEAD/README.md#webhookson) and [`.removeListener()`](https://github.com/octokit/webhooks.js/blob/HEAD/README.md#webhooksremovelistener).
```
G-Rath:
```suggestion
The `middleware` API implements [`.on()`](../README.md#webhookson) and [`.removeListener()`](../README.md#webhooksremovelistener).
```

I think that should fix the issue :)

gr2m: The current links work on GitHub: https://github.com/lumaxis/webhooks.js/blob/master/src/middleware/README.md

Is there a context where they do not work for you?

lumaxis: @gr2m Yes, the links on the page you linked to do not actually work for me 🤔 
If I click either of them, I land on https://github.com/lumaxis/webhooks.js/tree/master/src#webhookson which is wrong.

G-Rath: They don't work for me - they take me to https://github.com/octokit/webhooks.js/tree/master/src#webhookson which is why I think `README.md` needs to be added, to point to the specific file (since `../` is just a directory)

lumaxis: Almost! Your exact suggestion doesn't work for me but I think `../../README.md` does 🤔

lumaxis:
I just noticed this is broken in some of the other subfolder `README.md` files too, let me clone this branch locally real quick and fix all of them.

src/middleware/README.md
G-Rath: Yup, that's my math being wrong - we're 2 directories down, not one 😅

gr2m: huh odd, I could have sworn they worked before 🤷🏼 

This shoudl work

```suggestion
The `middleware` API implements [`.on()`](../../README.md#webhookson) and [`.removeListener()`](../../README.md#webhooksremovelistener).
```

lumaxis: Yep! Just force-pushed, I think this should fix all of them.

gr2m (approved):
Looks good, thanks a lot!

lumaxis:
Is this good to merge? 🙂

wolfy1339:
You'll need to merge in the `master` branch

lumaxis:
Done!

wolfy1339 (approved):

PR #452, opened 2021-02-08
refactor(types): rename EmitterWebhookEventName -> EmitterEventName

jablko:
Is there any objection to renaming `EmitterWebhookEventName` -> `EmitterEventName`, now that `"*"` and `"error"` have been removed?
- https://github.com/octokit/webhooks.js/pull/444#issuecomment-774778669
- https://github.com/octokit/webhooks.js/pull/444#discussion_r572263542

-----
[View rendered README.md](https://github.com/jablko/webhooks.js/blob/patch-5/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@
   - [webhooks.middleware()](#webhooksmiddleware)
   - [Webhook events](#webhook-events)
 - [TypeScript](#typescript)
-  - [`EmitterWebhookEventName`](#emitterwebhookeventname)
-  - [`EmitterWebhookEvent`](#emitterwebhookevent)
+  - [`EmitterEventName`](#emittereventname)
+  - [`EmitterEvent`](#emitterevent)
 - [License](#license)
 
 <!-- tocstop -->
@@ -652,11 +652,11 @@ In addition to these types, `@octokit/webhooks` exports 2 types specific to itse
 
 Note that changes to the exported types are not considered breaking changes, as the changes will not impact production code, but only fail locally or during CI at build time.
 
-### `EmitterWebhookEventName`
+### `EmitterEventName`
 
 A union of all possible events supported by the event emitter.
 
-### `EmitterWebhookEvent`
+### `EmitterEvent`
 
 The object that is emitted by `@octokit/webhooks` as an event; made up of an `id`, `name`, and `payload` properties.
 An optional generic parameter can be passed to narrow the type of the `payload` property to be based on the `name` of the event.
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -1,6 +1,6 @@
 import type {
-  EmitterWebhookEvent,
-  EmitterWebhookEventName,
+  EmitterEvent,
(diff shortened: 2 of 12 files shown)


G-Rath:
Is there any reason to? The name is still accurate, and I'd prefer being explicit for now than saving a few characters.

The current name was picked to try and make it a bit clearer what the types represented, which still has value given we've only just released the new major that removes the `*` and `error` events.

Plus this'll be another change for TS users updating to v8, as they'll have to update their types all over again.

jablko:
> Is there any reason to? The name is still accurate, and I'd prefer being explicit for now than saving a few characters.
> 
> The current name was picked to try and make it a bit clearer what the types represented, which still has value given we've only just released the new major that removes the `*` and `error` events.

I think `EmitterEventName` helps to distinguish `EmitterWebhookEventName` (`check_run` or `check_run.completed`) from `WebhookEventName` (`check_run` only) https://github.com/octokit/webhooks/blob/03b633b2aed02709c55e3df897747f278ad24397/schema.d.ts#L5729

and `EmitterWebhookEvent` (`{ id, name, payload }`) from `WebhookEvent` (payload only)?
https://github.com/octokit/webhooks/blob/03b633b2aed02709c55e3df897747f278ad24397/schema.d.ts#L5731

I think "Webhook" makes these names more similar rather than more explicit, without clarifying what they represent (since we don't emit anything else)?

`EmitterEventName` is also more consistent with e.g. `emitterEventNames` and `TEmitterEvent` vs. `TWebhookEvent`, which I think helps reinforce the distinction:
https://github.com/octokit/webhooks.js/blob/1891034b602a2193a08f1ac3ad31775df65ca578/src/types.ts#L8
https://github.com/octokit/webhooks.js/blob/1891034b602a2193a08f1ac3ad31775df65ca578/src/types.ts#L11

> Plus this'll be another change for TS users updating to v8, as they'll have to update their types all over again.

Yes, although `EmitterEventName` is in some cases the name of the type in v7: https://github.com/octokit/webhooks.js/blob/79d9db0089e8ea038445cc67247d259e47a0df19/src/index.ts#L30

G-Rath:
While I agree with you about them being more consistent, I don't think the difference is great enough to counter [my previous points](https://github.com/octokit/webhooks.js/pull/452#issuecomment-775412242).

I'd prefer to leave the names as they are for now to give the types a chance to settle.

gr2m:
> I'd prefer to leave the names as they are for now to give the types a chance to settle.

I agree. The change is not worth the friction it will cause. 

Thanks for bringing it up @jablko, we will revisit the exported type names in future anyway. We will likely split up `@octokit/webhooks` into two packages (`@octokit/webhooks` and `@octokit/events`), and that will be a good time to revisit the naming conventions.

jablko:
I'm happy with punting this change until then :+1:

gr2m:
I'll close this PR for now as there is nothing actionable. Let's revisit once we did the split into `@octokit/webhooks` and `@octokit/events`

PR #455, opened 2021-02-10
feat(types): accept read-only event names

jablko:
Add `readonly` modifier to event names parameters.

### DIFF
--- a/src/event-handler/on.ts
+++ b/src/event-handler/on.ts
@@ -19,10 +19,16 @@ function handleEventHandlers(
 }
 export function receiverOn(
   state: State,
-  webhookNameOrNames: EmitterWebhookEventName | EmitterWebhookEventName[],
+  webhookNameOrNames:
+    | EmitterWebhookEventName
+    | readonly EmitterWebhookEventName[],
   handler: Function
 ) {
-  if (Array.isArray(webhookNameOrNames)) {
+  if (
+    (Array.isArray as (arg: any) => arg is readonly unknown[])(
+      webhookNameOrNames
+    )
+  ) {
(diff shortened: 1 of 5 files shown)


src/event-handler/on.ts
```diff
@@ -19,10 +19,16 @@ function handleEventHandlers(
...
   state: State,
-  webhookNameOrNames: EmitterWebhookEventName | EmitterWebhookEventName[],
+  webhookNameOrNames:
+    | EmitterWebhookEventName
+    | readonly EmitterWebhookEventName[],
   handler: Function
 ) {
-  if (Array.isArray(webhookNameOrNames)) {
+  if (
+    (Array.isArray as (arg: any) => arg is readonly unknown[])(
```
G-Rath, line 28: Not a fan of this - I don't think `readonly` gives us much advantages to be worth needing to cast like this, nor to have extra runtime code.

jablko: What's the best workaround?

oscard0m: @G-Rath

oscard0m: @jablko is there a specific use case where you find these changes valuable? Maybe with the example you would like to get type-covered we can come up together with other alternatives.

G-Rath: > What's the best workaround?

There isn't really a workaround.

gr2m:
I agree with @G-Rath's point here: https://github.com/octokit/webhooks.js/pull/455/files#r574053711. Is there any advantage of using a readonly array here? I don't think a problem ever came up from users that would be resolved by this change

gr2m:
closing due to inactivity

PR #457, opened 2021-02-10
feat(logging): add log option

jablko:
I'd like to be able to redirect the log messages, so I added a `{ log }` option, consistent with the other `@octokit/*` packages, based on this feedback https://github.com/octokit/webhooks.js/pull/201#pullrequestreview-463894384

@oscard0m is already working on this in #201 (where there's existing discussion/feedback) and I'm happy to set the base branch of this PR to https://github.com/octokit/webhooks.js/tree/use-conole-log-level-package-instead-of-debug-package, however I think that branch has been removed.

If you resurrect that branch I'm happy to request pulling this PR into it. Whichever you prefer! I want to contribute to, not supersede, #201 :heart:

### DIFF
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -32,8 +32,8 @@ export function createEventHandler(options: Options<any>): EventHandler {
   const state: State = {
     hooks: {},
     log: {
-      debug: () => {},
-      info: () => {},
+      debug: /* istanbul ignore next: unused */ () => {},
+      info: /* istanbul ignore next: unused */ () => {},
       warn: console.warn.bind(console),
       error: console.error.bind(console),
       ...options.log,
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -2,6 +2,7 @@ import type {
   EmitterWebhookEvent,
   EmitterWebhookEventName,
   HandlerFunction,
+  Logger,
   Options,
   State,
   WebhookEventHandlerError,
@@ -31,13 +32,7 @@ interface EventHandler<TTransformed = unknown> {
 export function createEventHandler(options: Options<any>): EventHandler {
   const state: State = {
     hooks: {},
-    log: {
-      debug: /* istanbul ignore next: unused */ () => {},
-      info: /* istanbul ignore next: unused */ () => {},
-      warn: console.warn.bind(console),
-      error: console.error.bind(console),
-      ...options.log,
-    },
+    log: createLogger(options.log),
   };
 
   if (options && options.transform) {
@@ -50,5 +45,15 @@ export function createEventHandler(options: Options<any>): EventHandler {
     onError: onError.bind(null, state),
     removeListener: removeListener.bind(null, state),
     receive: receive.bind(null, state),
+  };
+}
+
+export function createLogger(logger?: Partial<Logger>) {
+  return {
+    debug: () => {},
+    info: /* istanbul ignore next: unused */ () => {},
+    warn: console.warn.bind(console),
+    error: console.error.bind(console),
+    ...logger,
   };
 }
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -31,6 +31,13 @@ interface EventHandler<TTransformed = unknown> {
 export function createEventHandler(options: Options<any>): EventHandler {
   const state: State = {
     hooks: {},
+    log: {
+      debug: () => {},
+      info: () => {},
+      warn: console.warn.bind(console),
+      error: console.error.bind(console),
+      ...options.log,
+    },
   };
 
   if (options && options.transform) {
--- a/src/middleware/index.ts
+++ b/src/middleware/index.ts
@@ -1,3 +1,4 @@
+import { debug } from "debug";
 import { createEventHandler } from "../event-handler/index";
 import { middleware } from "./middleware";
 import { Options, State } from "../types";
@@ -13,7 +14,7 @@ export function createMiddleware(options: Options<any>) {
     secret: options.secret,
     hooks: {},
     log: {
-      debug: () => {},
+      debug: options.log ? () => {} : debug("webhooks:receiver"),
       info: /* istanbul ignore next: unused */ () => {},
       warn: console.warn.bind(console),
       error: console.error.bind(console),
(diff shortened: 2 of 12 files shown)


src/types.ts
```diff
@@ -49,6 +50,12 @@ type Hooks = {
 export interface State extends Options<any> {
   eventHandler?: any;
   hooks: Hooks;
+  log: {
+    debug: (message: string) => unknown;
+    info: (message: string) => unknown;
+    warn: (message: string) => unknown;
+    error: (message: string) => unknown;
+  };
```
jablko: Copied from https://github.com/octokit/core.js/blob/124bf5a64a1aef1fe0591ccff9b8cca87663cdbf/src/types.ts#L15-L20

src/event-handler/index.ts
```diff
@@ -31,6 +31,13 @@ interface EventHandler<TTransformed = unknown> {
 export function createEventHandler(options: Options<any>): EventHandler {
   const state: State = {
     hooks: {},
+    log: {
+      debug: () => {},
+      info: () => {},
+      warn: console.warn.bind(console),
+      error: console.error.bind(console),
```
jablko: Copied from https://github.com/octokit/core.js/blob/124bf5a64a1aef1fe0591ccff9b8cca87663cdbf/src/index.ts#L110-L113

src/middleware/middleware.ts
```diff
@@ -26,6 +26,7 @@ export function middleware(
     }
 
     debugWebhooks(`ignored: ${request.method} ${request.url}`);
+    state.log.debug(`ignored: ${request.method} ${request.url}`);
```
gr2m, line 25: I wonder that if `state.log` is set, should we still call `debugWebhooks`? We could introduce the `{ log }` option without a breaking change that way. With the next breaking version, we will simply remove the `debugWebhooks` altogether.

In either way, we should add a test that captures the stdout output

jablko: Good idea, I've revised it to not call `debugWebhooks()` if the log option is set and added a basic test. What do you think?

gr2m: Looks good 👍🏼 

Could you adjust the test so that they don't log out anything, instead pass a mock and add an assertion that the expected output was logged?

image

src/types.ts
G-Rath: I think it'd be good to have these as an exported interface marked as optional, for reuse (I could see myself doing `const mockLogger: Logger = { ... }` in tests related to this package), and then do `Required<>` here.

src/event-handler/index.ts
```diff
@@ -31,6 +31,13 @@ interface EventHandler<TTransformed = unknown> {
 export function createEventHandler(options: Options<any>): EventHandler {
   const state: State = {
     hooks: {},
+    log: {
```
G-Rath: I wonder if maybe we should have a `createLogger` helper function that'll let us consolidate these, and ideally remove some of the `istanbul ignore`s?

I'm thinking something like:

```
const createLogger = (debugNamespace: string, logger: Logger | boolean): Logger => ({
  debug: logger ? () => {} : debug(`webhooks:${debugNamespace}`),
  info: console.info.bind(console),
  warn: console.warn.bind(console),
  error: console.error.bind(console),
  ...logger,
});
```

It also looks like we're not using `info` anywhere, so we could probably just omit that for now - if we wanted to use it in the future we could add it in as a non-breaking feature.

G-Rath: (`createLogger` could also be helpful for us, as we could then mock it in tests to return a set of mock functions)

src/middleware/middleware.ts
jablko: Done, I've revised the tests so they don't output anything and instead check the output is what's expected.

src/types.ts
jablko: Done, I added a `Logger` interface.

src/event-handler/index.ts
jablko: Done, I've extracted `createLogger()`.

My preference is to keep `info`, for consistency with the other `@octokit/*` packages. I can think of a couple ways to pacify Istanbul, what I like about the comment is that it's obvious why it's there, and it's trivial to remove when it's no longer needed.

src/event-handler/index.ts
```diff
@@ -45,3 +47,13 @@ export function createEventHandler(options: Options<any>): EventHandler {
     receive: receive.bind(null, state),
   };
 }
+
+export function createLogger(logger?: Partial<Logger>) {
```
G-Rath, line 51: I think it would be better if this lived in its own file somewhere, as the some of the functions are marked as being usable standalone and this is creating an implicit dependency on `event-handler`.

Having it in its own file means it won't trigger any import side-effects and make it easier to pull the standalones into their own packages if we decide to go that way in the future.

src/middleware/index.ts
```diff
@@ -12,6 +13,7 @@ export function createMiddleware(options: Options<any>) {
     path: options.path || "/",
     secret: options.secret,
     hooks: {},
+    log: createLogger(options.log || { debug: debug("webhooks:receiver") }),
```
G-Rath, line 16: nice! that's even cleaner 😅

src/event-handler/index.ts
jablko: Currently there isn't a problem pulling the standalones into their own packages or triggering import side effects because everything that imports `createLogger()` already imports `createEventHandler()`, which is why I thought it best to put it here. I can imagine that changing in the future, but I think my choice would be to cross that bridge when we come to it?

gr2m (approved):
This is great, thanks a lot Jack! 💐

src/event-handler/index.ts
G-Rath: I still would have preferred to have had this somewhere else, as it would make it easier to test in isolation and be more inline with the single responsibility principle. (but it's not a deal breaker - we can always fix it up later)

oscard0m:
> If you resurrect that branch I'm happy to request pulling this PR into it. Whichever you prefer! I want to contribute to, not supersede, #201 ❤️

Sorry for arriving late here. No problem in not using my original PR in this one :) @jablko

PR #458, opened 2021-02-11
feat(typescript): compile-time missing options error

jablko:
Like at run time https://github.com/octokit/webhooks.js/blob/80ee43fc95d8853ba1439d55a54f1b9cee7813e1/test/integration/server-test.ts#L26-L27

### DIFF
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -30,11 +30,11 @@ interface EventHandler<TTransformed> {
 }
 
 export function createEventHandler<TTransformed>(
-  options: Options<TTransformed>
+  options?: Options<TTransformed>
 ): EventHandler<TTransformed> {
   const state: State = {
     hooks: {},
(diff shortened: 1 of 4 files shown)


src/event-handler/index.ts
```diff
@@ -28,7 +28,7 @@ interface EventHandler<TTransformed = unknown> {
   receive(event: EmitterWebhookEvent): Promise<void>;
 }
 
-export function createEventHandler(options: Options<any>): EventHandler {
+export function createEventHandler(options?: Options<any>): EventHandler {
```
jablko: The `createEventHandler()` argument is optional?

gr2m: I'd say yes

gr2m (dismissed):

gr2m (dismissed):

gr2m (approved):

PR #459, opened 2021-02-11
fix(typescript): infer TTransformed from createEventHandler() options

jablko:
Should we get the `TTransformed` type from the `createEventHandler()` options, like with do with the `Webhooks` options? https://github.com/octokit/webhooks.js/blob/80ee43fc95d8853ba1439d55a54f1b9cee7813e1/src/index.ts#L45


G-Rath (approved):
oh good catch - I missed that one!

PR #460, opened 2021-02-12
refactor(typescript): simplify TransformMethod type

jablko:
Can we simplify the `TransformMethod` type as follows?
```Diff
-type TransformMethod<T extends EmitterWebhookEvent, V = T> = (
+type TransformMethod<T> = (
   event: EmitterWebhookEvent
-) => V | PromiseLike<V>;
+) => T | PromiseLike<T>;
```

### DIFF
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -29,7 +29,7 @@ interface EventHandler<TTransformed> {
 }
 
 export function createEventHandler<TTransformed>(
-  options: Options<any, TTransformed>
+  options: Options<TTransformed>
 ): EventHandler<TTransformed> {
   const state: State = {
     hooks: {},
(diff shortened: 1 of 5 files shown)


gr2m (dismissed):
Looks good to me! I'm always a bit nervous about TypeScript changes though, and given that we just broke `@octokit/webhooks` for some TypeScript users, I'd appreciate a 2nd review

G-Rath (dismissed):

G-Rath:
@gr2m this should be fine - the reason I didn't originally do this was for similar reasons: lack of familiarity with the codebase + didn't want to make too many waves all at once :)

(in fact this might fix the issue in `@octokit/app.js`)

G-Rath:
(thinking about it, this is probably meant to be `event: T`, but I can't think of a situation where that'd improve things - worse case we can adjust the type later 🤷)

gr2m (approved):
Thanks @G-Rath! I resolved the merge conflict which invalidated your review, but this is good to merge now as soon as CI is happy

gr2m:
I've merged it as a fix so that users will get the new version, in case something blows up it will be easier to track down and revert

PR #461, opened 2021-02-12
feat(typescript): compile-time missing secret error

jablko:
Should this be a compile-time as well as a run-time error?
```Diff
--- a/test/unit/middleware-constructor-test.ts
+++ b/test/unit/middleware-constructor-test.ts
@@ -1,5 +1,6 @@
 import { createMiddleware as Middleware } from "../../src/middleware";
 
 test("options: none", () => {
+  // @ts-expect-error
   expect(() => Middleware({})).toThrow();
 });

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -88,7 +88,9 @@ class Webhooks<TTransformed = unknown> {
 }
 
 /** @deprecated `createWebhooksApi()` is deprecated and will be removed in a future release of `@octokit/webhooks`, please use the `Webhooks` class instead */
-const createWebhooksApi = <TTransformed>(options: Options<TTransformed>) => {
+const createWebhooksApi = <TTransformed>(
+  options: Options<TTransformed> & { secret: string }
+) => {
   const log = createLogger(options.log);
   log.warn(
     "[@octokit/webhooks] `createWebhooksApi()` is deprecated and will be removed in a future release of `@octokit/webhooks`, please use the `Webhooks` class instead"
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,6 +10,7 @@ import {
   EmitterWebhookEventName,
   HandlerFunction,
   Options,
+  State,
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/index.ts [hunk 1]
```diff
@@ -42,12 +41,12 @@ class Webhooks<
...
   ) => Promise<void>;
 
-  constructor(options?: Options<E, TTransformed>) {
+  constructor(options?: Options<E, TTransformed> & { secret: string }) {
     if (!options || !options.secret) {
       throw new Error("[@octokit/webhooks] options.secret required");
     }
 
-    const state: State = {
+    const state = {
```
G-Rath:
```suggestion
    const state: State & { secret: string }= {
```

This needs to be explicitly typed as it's an object

src/middleware/index.ts
```diff
@@ -1,13 +1,13 @@
...
+import { Options } from "../types";
 
-export function createMiddleware(options: Options<any>) {
+export function createMiddleware(options: Options<any> & { secret: string }) {
   if (!options || !options.secret) {
     throw new Error("[@octokit/webhooks] options.secret required");
   }
 
-  const state: State = {
+  const state = {
```
G-Rath: This needs to be explicitly typed as its an object.

G-Rath (dismissed):

src/index.ts
jablko: I've applied this change although I confess I don't understand it: why do objects need to be explicitly typed?

wolfy1339 (dismissed):

src/index.ts
jablko: Answered in https://github.com/octokit/webhooks.js/pull/466#discussion_r575863299, thanks!

gr2m (approved):
Thanks! We shoudl have that a long time ago! I am sure we will be able to make the types much more elegant once we split out the event handling into its own module

PR #463, opened 2021-02-12
refactor: simplify conditions

G-Rath:
Noticed a few conditions which could be simplified with optional chaining, optional assignment, and a few other ways.

Making this a draft for now as we've got a few PRs changing some optional/non-optional types which'll affect these changes, so aiming to land after they're landed.

### DIFF
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -37,7 +37,7 @@ export function createEventHandler<TTransformed>(
     log: createLogger(options && options.log),
   };
 
-  if (options && options.transform) {
+  if (options?.transform) {
     state.transform = options.transform;
   }
 
--- a/src/event-handler/receive.ts
+++ b/src/event-handler/receive.ts
@@ -42,7 +42,7 @@ export function receiverHandle(state: State, event: EmitterWebhookEvent) {
     return Promise.reject(error);
   }
 
-  if (!event || !event.name) {
+  if (!event?.name) {
     throw new AggregateError(["Event name not passed"]);
   }
 
--- a/src/event-handler/wrap-error-handler.ts
+++ b/src/event-handler/wrap-error-handler.ts
@@ -11,10 +11,8 @@ export function wrapErrorHandler(handler: Function, error: Error) {
     console.log(error);
   }
 
-  if (returnValue && returnValue.catch) {
-    returnValue.catch((error: Error) => {
-      console.log('FATAL: Error occurred in "error" event handler');
-      console.log(error);
-    });
-  }
+  returnValue?.catch?.((error: Error) => {
+    console.log('FATAL: Error occurred in "error" event handler');
+    console.log(error);
+  });
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -47,7 +47,7 @@ class Webhooks<TTransformed = unknown> {
   ) => void | Promise<void>;
 
   constructor(options: Options<TTransformed> & { secret: string }) {
-    if (!options || !options.secret) {
+    if (!options?.secret) {
       throw new Error("[@octokit/webhooks] options.secret required");
     }
 
--- a/src/middleware-legacy/index.ts
+++ b/src/middleware-legacy/index.ts
@@ -5,7 +5,7 @@ import { middleware } from "./middleware";
 import { Options, State } from "../types";
 
 export function createMiddleware(options: Options & { secret: string }) {
-  if (!options || !options.secret) {
+  if (!options?.secret) {
     throw new Error("[@octokit/webhooks] options.secret required");
   }
 
--- a/src/event-handler/on.ts
+++ b/src/event-handler/on.ts
@@ -37,7 +37,7 @@ export function receiverOn(
     throw new Error(message);
   }
 
-  if (emitterEventNames.includes(webhookNameOrNames)) {
+  if (!emitterEventNames.includes(webhookNameOrNames)) {
     state.log.warn(
       `"${webhookNameOrNames}" is not a known webhook name (https://developer.github.com/v3/activity/events/types/)`
     );
--- a/src/event-handler/on.ts
+++ b/src/event-handler/on.ts
@@ -11,10 +11,7 @@ function handleEventHandlers(
   webhookName: EmitterWebhookEventName | "error" | "*",
   handler: Function
 ) {
-  if (!state.hooks[webhookName]) {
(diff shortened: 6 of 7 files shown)


src/event-handler/index.ts
```diff
@@ -37,7 +37,7 @@ export function createEventHandler<TTransformed>(
     log: createLogger(options.log),
   };
 
-  if (options && options.transform) {
+  if (options?.transform) {
```
gr2m, line 40: I had problems with optional chaining on the past. The compiled JS would continue to have it which broke lots of runtime environments. Can you check the compiled code in the pkg folder if the .? gets replaced?

G-Rath: TypeScript should be transpiling it out of the box, but I'll confirm to be safe :)

gr2m: it was in the `dist-web` folder here: https://github.com/gr2m/octokit-plugin-create-pull-request/issues/68
We don't do a web build for `@octokit/webhooks` yet so we should be fine

oscard0m (dismissed):
LGTM

gr2m:
I don't think that `jest` likes `options?.transform` syntax? Maybe there is some config we could use to make it happy?

wolfy1339:
~I don't think that the syntax is the problem.~ **EDIT**: So it seems optional chaining only has support for NodeJS 14+

There was a condition where the was a "not" operator misssing, therefore it made tests error whenever a valid webhook name was passed instead of only erroring when an invalid name was passed

wolfy1339:
Here's what's going on:
* We seem to support NodeJS 10, 12; or at least we test against them
* We target ES2020 in our `tsconfig.json`, which doesn't have great support in v10 and v12
* Optional chaining and Nullish coalescing are only supported in NodeJS 13.14+
* To use these features would require using the `--harmony` flag for node (only suitable for v13.14+)

There needs to be proper documentation of which Node versions we support in order to establish the baseline target.

gr2m:
We support support Node 10, 12, and 14 right now. Node 10 will be out of after April 30, 2021, so we will drop support across the octokit modules and release breaking changes. When we drop Node 10, we will start testing against Node 16, which is planned to be released on April 20, 2021

> Optional chaining and Nullish coalescing are only supported in NodeJS 13.14+

We shouldn't use it then.



> We target ES2020 in our `tsconfig.json`, which doesn't have great support in v10 and v12

We should update https://github.com/octokit/tsconfig/blob/104b85db45120e71119840489b6372d764441ed5/tsconfig.json#L8 then I guess? What target should we use for v12 support?

wolfy1339:
> > Optional chaining and Nullish coalescing are only supported in NodeJS 13.14+
> We shouldn't use it then.

That shouldn't be an issue if the next point is addressed, as TS will transpile it down to a format that the NodeJS version can understand.

> What target should we use for v12 support?
 
ES2019 should work fine

As long as the right target version is set, we can theoretically use any new features in lower versions of Node

gr2m:
> ES2019 should work fine

just double checking, this page says ES2018:
https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-10

I think there was a reason that we went with ES2020. The compilation result was really bad with other targets, which made debugging harder. I think it was about async/await and/or classes.

Maybe we should stick with 2020 and restrain ourselves from using to advanced features? I always make sure we test in the node version we want to support, so we shouldn't break anyone's usage

src/event-handler/index.ts
```diff
@@ -37,7 +37,7 @@ export function createEventHandler<TTransformed>(
     log: createLogger(options && options.log),
   };
 
-  if (options && options.transform) {
+  if (options?.transform) {
```
gr2m, line 40: please revert as per discussion

src/event-handler/receive.ts
```diff
@@ -42,7 +42,7 @@ export function receiverHandle(state: State, event: EmitterWebhookEvent) {
     return Promise.reject(error);
   }
 
-  if (!event || !event.name) {
+  if (!event?.name) {
```
gr2m, line 45: please revert as per discussion

src/event-handler/wrap-error-handler.ts
```diff
@@ -11,10 +11,8 @@ export function wrapErrorHandler(handler: Function, error: Error) {
     console.log(error);
   }
 
-  if (returnValue && returnValue.catch) {
-    returnValue.catch((error: Error) => {
-      console.log('FATAL: Error occurred in "error" event handler');
-      console.log(error);
-    });
-  }
+  returnValue?.catch?.((error: Error) => {
```
gr2m, line 14: please revert as per discussion

src/index.ts
```diff
@@ -47,7 +47,7 @@ class Webhooks<TTransformed = unknown> {
   ) => void | Promise<void>;
 
   constructor(options: Options<TTransformed> & { secret: string }) {
-    if (!options || !options.secret) {
+    if (!options?.secret) {
```
gr2m, line 50: please revert as per discussion

src/middleware-legacy/index.ts
```diff
@@ -5,7 +5,7 @@ import { middleware } from "./middleware";
 import { Options, State } from "../types";
 
 export function createMiddleware(options: Options & { secret: string }) {
-  if (!options || !options.secret) {
+  if (!options?.secret) {
```
gr2m, line 8: please revert as per discussion

gr2m (changes requested):
please revert the optional chaining changes, as we want to continue to use ES2020 as build target which keeps them in, but that in turn breaks node 12.

G-Rath:
@gr2m if we're wanting to still support node 12 we should remain on [ES2019, with a sprinkle of ES2020](https://github.com/tsconfig/bases/blob/master/bases/node12.json#L6)

Either way it's fine - I'm going to close this PR for now because I'm a bit swamped and the reverts mean this PR boils down to two small changes (using `includes` instead of `indexOf`, and replacing an `if` with a `return`), as I should also revert the logical OR assignment as that's not supported until Node 15 and so would need to be transpiled like optional chaining if we wanted to use it.

I'm happy if someone else wants to cherry-pick the landable changes into a new PR :)

wolfy1339:
I just verified the output produced, and everything is getting transpiled down to ES2019. I don't personally see any issue with these changes.

However, all the changes related to the middleware are no longer relevant or needed, which makes some of these reverts un-needed

PR #464, opened 2021-02-13
fix(typescript): HandlerFunction argument

jablko:
What do you think about revising `HandlerFunction` as follows, to make the argument `TTransformed` if the `transform` option is set and `EmitterWebhookEvent<TName>` if not, instead of always the intersection?
```Diff
--- a/src/types.ts
+++ b/src/types.ts
@@ -32,8 +32,8 @@ type TransformMethod<T> = (event: EmitterWebhookEvent) => T | PromiseLike<T>;
 
 export type HandlerFunction<
   TName extends EmitterWebhookEventName,
-  TTransformed
-> = (event: EmitterWebhookEvent<TName> & TTransformed) => any;
+  TTransform
+> = (event: TTransform extends TransformMethod<infer T> ? T : EmitterWebhookEvent<TName>) => any;
 
 type Hooks = {
   [key: string]: Function[];
```

This fixes the types in cases like:
```Diff
--- a/test/integration/event-handler-test.ts
+++ b/test/integration/event-handler-test.ts
@@ -131,15 +131,15 @@ test("options.transform", (done) => {
   const eventHandler = createEventHandler({
     transform: (event) => {
       expect(event.id).toBe("123");
       return "funky";
     },
   });
 
-  eventHandler.on("push", (event: EmitterWebhookEvent) => {
+  eventHandler.on("push", (event: string) => {
     expect(event).toBe("funky");
 
     done();

### DIFF
--- a/test/integration/event-handler-test.ts
+++ b/test/integration/event-handler-test.ts
@@ -135,7 +135,7 @@ test("options.transform", (done) => {
     },
   });
 
-  eventHandler.on("push", (event: EmitterWebhookEvent) => {
+  eventHandler.on("push", (event: string) => {
     expect(event).toBe("funky");
 
     done();
@@ -155,7 +155,7 @@ test("async options.transform", (done) => {
     },
   });
 
-  eventHandler.on("push", (event: EmitterWebhookEvent) => {
+  eventHandler.on("push", (event: string) => {
     expect(event).toBe("funky");
     done();
   });
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -15,23 +15,23 @@ import {
 import { receiverHandle as receive } from "./receive";
 import { removeListener } from "./remove-listener";
 
-interface EventHandler<TTransformed> {
+interface EventHandler<T extends Options> {
   on<E extends EmitterWebhookEventName>(
     event: E | E[],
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: test/integration/event-handler-test.ts [hunk 1]
```diff
@@ -135,7 +135,7 @@ test("options.transform", (done) => {
     },
   });
 
-  eventHandler.on("push", (event: EmitterWebhookEvent) => {
+  eventHandler.on("push", (event: string) => {
```
gr2m, line 138: why this change?

G-Rath: It's because [the `transform` method is transforming the type into `string`](https://github.com/octokit/webhooks.js/pull/464/files#diff-19ed883872f2faee0e85fe413f91e9718bfa273b0309cc7be942453f6eec30a6R132-R135) (technically, it's type is `"funky"`).

gr2m: Ah of course, yes, thank you!

G-Rath (dismissed):
Very clever! I clearly need to step up my conditional typing game 😉

gr2m (approved):
Looks good to me 👍🏼

wolfy1339:
@G-Rath can you do a final review and merge this

G-Rath (approved):

PR #465, opened 2021-02-14
fix(typescript): compile-time invalid argument errors

jablko:
This minor change to the test types confirms that an invalid algorithm and the `"*"` and `"error"` event names are compile-time as well as run-time errors, by replacing a couple of type assertions with `// @ts-expect-error` comments. Doesn't touch the implementation, only the tests.

The first commit is a distantly related cosmetic change: replaces a try/catch test with the standard `.toThrow()` matcher. I'm happy to move it out of this PR if you prefer?

### DIFF
--- a/test/unit/event-handler-on-test.ts
+++ b/test/unit/event-handler-on-test.ts
@@ -1,5 +1,5 @@
 import { receiverOn } from "../../src/event-handler/on";
-import { EmitterWebhookEventName, State } from "../../src/types";
+import { State } from "../../src/types";
 
 function noop() {}
 
@@ -24,17 +24,15 @@ test("receiver.on with invalid event name", () => {
 });
 
 test("receiver.on with event name of '*' throws an error", () => {
-  expect(() =>
-    receiverOn(state, "*" as EmitterWebhookEventName, noop)
-  ).toThrowError(
+  // @ts-expect-error
+  expect(() => receiverOn(state, "*", noop)).toThrowError(
     'Using the "*" event with the regular Webhooks.on() function is not supported. Please use the Webhooks.onAny() method instead'
   );
 });
 
 test("receiver.on with event name of 'error' throws an error", () => {
-  expect(() =>
-    receiverOn(state, "error" as EmitterWebhookEventName, noop)
-  ).toThrowError(
+  // @ts-expect-error
+  expect(() => receiverOn(state, "error", noop)).toThrowError(
     'Using the "error" event with the regular Webhooks.on() function is not supported. Please use the Webhooks.onError() method instead'
   );
 });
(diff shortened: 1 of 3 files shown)


test/unit/event-handler-on-test.ts
```diff
@@ -24,17 +24,15 @@ test("receiver.on with invalid event name", () => {
 });
 
 test("receiver.on with event name of '*' throws an error", () => {
-  expect(() =>
-    receiverOn(state, "*" as EmitterWebhookEventName, noop)
-  ).toThrowError(
+  // @ts-expect-error
+  expect(() => receiverOn(state, "*", noop)).toThrowError(
```
G-Rath:
```suggestion
  expect(() => receiverOn(state, "*", noop)).toThrow(
```

[`toThrowError` is an alias](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-alias-methods.md), which we'd ideally like to remove at some point.

test/unit/event-handler-on-test.ts
```diff
@@ -24,17 +24,15 @@ test("receiver.on with invalid event name", () => {
...
     'Using the "*" event with the regular Webhooks.on() function is not supported. Please use the Webhooks.onAny() method instead'
   );
 });
 
 test("receiver.on with event name of 'error' throws an error", () => {
-  expect(() =>
-    receiverOn(state, "error" as EmitterWebhookEventName, noop)
-  ).toThrowError(
+  // @ts-expect-error
+  expect(() => receiverOn(state, "error", noop)).toThrowError(
```
G-Rath:
```suggestion
  expect(() => receiverOn(state, "error", noop)).toThrow(
```

[`toThrowError` is an alias](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-alias-methods.md), which we'd ideally like to remove at some point.

G-Rath:
I'm not the biggest fan of the cast -> `@ts-expect-error` changes, as we have other tests (specifically, `typescript-validate.ts`) to cover the compile-side errors whereas the focus of these tests is on the runtime behaviour.

I do approve of the refactoring of the tests to use native jest matchers however 🙂

test/unit/event-handler-on-test.ts
jablko: I'm in favor :+1: I've applied these changes.

jablko:
At worst, the cast -> `// @ts-expect-error` change is benign, neither positive nor negative?

gr2m (approved):
I like the changes 👍🏼 It makes it more clear that we are testing runtime errors and want to explicitly ignore compile time errors in order to test them

Issue #425, opened 2021-02-02
Improve internal Types


oscard0m:
> > From your explanation seems like we should invest some more time on better typing this
> 
> The types for this PR are correct it's our internal types that could use improvement, and definitely worth investing improving. However, the gains vs amount of work that'd be involved would be pretty low, so I'd not say it's a super high priority (in particular because of the `State` type - i.e. we mark `eventEmitter` as optional, but don't actually check for its existence in a few places; our implementation + test suite ensure it's not `undefined` at runtime, but the types are not representing that properly).

We can create an issue of it (`maintenance` label) so at least we keep track of it. Thumbs Up if you are ok with it.

_Originally posted by @oscard0m in https://github.com/octokit/webhooks.js/issues/422#issuecomment-771928554_

PR #466, opened 2021-02-14
refactor(typescript): type State#eventHandler

jablko:
What do you think about changing `State#eventHandler` from `any` to `EventHandler`? I don't think this affects the external types, just makes the internal types more accurate.
```Diff
--- a/src/types.ts
+++ b/src/types.ts
@@ -40,7 +41,7 @@ type Hooks = {
 };
 
 export interface State extends Options<any> {
-  eventHandler?: any;
+  eventHandler?: EventHandler;
   hooks: Hooks;
   log: Logger;
 }

### DIFF
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -15,7 +15,7 @@ import {
 import { receiverHandle as receive } from "./receive";
 import { removeListener } from "./remove-listener";
 
-interface EventHandler<TTransformed> {
+export interface EventHandler<TTransformed = unknown> {
   on<E extends EmitterWebhookEventName>(
     event: E | E[],
     callback: HandlerFunction<E, TTransformed>
@@ -26,7 +26,7 @@ interface EventHandler<TTransformed> {
     event: E | E[],
     callback: HandlerFunction<E, TTransformed>
   ): void;
-  receive(event: EmitterWebhookEvent): Promise<void>;
+  receive(event: EmitterWebhookEvent | Error): Promise<void>;
 }
 
 export function createEventHandler<TTransformed>(
--- a/src/event-handler/receive.ts
+++ b/src/event-handler/receive.ts
@@ -29,7 +29,10 @@ function getHooks(
 }
 
 // main handler function
-export function receiverHandle(state: State, event: EmitterWebhookEvent) {
+export function receiverHandle(
+  state: State,
+  event: EmitterWebhookEvent | Error
+) {
   const errorHandlers = state.hooks.error || [];
 
   if (event instanceof Error) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 import { IncomingMessage, ServerResponse } from "http";
 import { createLogger } from "./createLogger";
-import { createEventHandler } from "./event-handler/index";
+import { createEventHandler, EventHandler } from "./event-handler/index";
 import { createMiddleware } from "./middleware/index";
 import { middleware } from "./middleware/middleware";
 import { verifyAndReceive } from "./middleware/verify-and-receive";
@@ -10,6 +10,7 @@ import {
   EmitterWebhookEventName,
   HandlerFunction,
   Options,
+  State,
   WebhookError,
   WebhookEventHandlerError,
 } from "./types";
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

File: src/event-handler/index.ts [hunk 1]
```diff
@@ -26,6 +26,7 @@ interface EventHandler<TTransformed> {
     event: E | E[],
     callback: HandlerFunction<E, TTransformed>
   ): void;
+  receive(event: Error): Promise<void>;
   receive(event: EmitterWebhookEvent): Promise<void>;
```
G-Rath:
```suggestion
  receive(event: EmitterWebhookEvent | Error): Promise<void>;
```

src/index.ts
```diff
@@ -45,7 +44,7 @@ class Webhooks<TTransformed> {
       throw new Error("[@octokit/webhooks] options.secret required");
     }
 
-    const state: State = {
+    const state = {
```
G-Rath: I think this should remain, since it should be conforming to the `State` type (which itself is in a bit of a mess)

G-Rath:
We'd love it! This is part of #425 - the whole `State` type needs a bit of a cleanup as it's not being used consistently.

You may find that the needs to happen for this change to be landable without having to cast or otherwise modify types elsewhere, but I'd be ok with that if it was only in a few places and wasn't anything too magical.

src/event-handler/index.ts
jablko: Done :+1:

src/index.ts
jablko: I'm in favor of dropping it because the inferred type is narrower than `State`:
- It's assignable to `State` (the following parameters are `State`) https://github.com/octokit/webhooks.js/blob/dd65cf5e97225a8a216af818a4d7f8c830621674/src/index.ts#L62-L63
- The explicit type discards information: the fact that `eventHandler` is non-null

That said it's a minor detail, give the word and I'll make this change.

G-Rath: I hear what you're saying, and am not wholly against it - however, I always tend to prefer explicitly typing object and arrays in most cases, as it makes it clearer what's in the wrong when there's an error (since it'll error on this assignment being wrong, rather than the usage being wrong), which actually leads me to my second point:

> The explicit type discards information: the fact that eventHandler is non-null

Exactly, which may or may not be correct and is effectively the tl;dr of #425 😅 

`State` looks to me like the classic genetically named internal-only state type that tend to popup when you move from JS to TS.

Because of that, I'd prefer to keep the explicit type annotation, to capture what type we're aiming for and hopefully make it easier to spot where our inconsistencies are when refactoring that type.

wolfy1339: > `State` looks to me like the classic genetically named internal-only state type that tend to popup when you move from JS to TS.

That is pretty much what happened there when I initially moved to TS.

TBF, we haven't really done much refactoring since, I think it's time to revisit a lot of this stuff with the new experiences we have and new TS features as well

jablko: Okay, thanks for this thoughtful explanation. Done :heavy_check_mark:

wolfy1339:
What needs to happen here in order to get this PR going ahead? @G-Rath @jablko

src/event-handler/index.ts
```diff
@@ -26,7 +26,7 @@ interface EventHandler<TTransformed> {
     event: E | E[],
     callback: HandlerFunction<E, TTransformed>
   ): void;
-  receive(event: EmitterWebhookEvent): Promise<void>;
+  receive(event: EmitterWebhookEvent | Error): Promise<void>;
```
gr2m, line 29: we shouldn't do that as we have the `onError` handler

src/event-handler/receive.ts
```diff
@@ -29,7 +29,10 @@ function getHooks(
 }
 
 // main handler function
-export function receiverHandle(state: State, event: EmitterWebhookEvent) {
+export function receiverHandle(
+  state: State,
+  event: EmitterWebhookEvent | Error
```
gr2m, line 34: we shouldn't accept `Error` here

gr2m (commented):
I'm okay with doing this change, but we'd need to resolve some conflicts.

I feel like the code is getting a bit messy, but that's okay, we'll clean it up soon (last famous words)

src/event-handler/index.ts
gr2m: nevermind me, maybe I shouldn't review PRs when the kids wake me at 4am 🤷🏼

gr2m:
> What needs to happen here in order to get this PR going ahead? @G-Rath @jablko

could one of you let us know what the next is here?

gr2m:
shall we close this PR?

wolfy1339:
Yes, lets close this. We can revisit later

PR #469, opened 2021-02-15
ci: fix `@octokit/webhooks-definitions` updater

wolfy1339:
Remove the `--save-dev` flag as `@octokit/webhooks-definitions` is now a production dependency as of v8


G-Rath (approved):
(there is a `ci` commit type you can use for these sort of changes i.e `ci: fix @octokit/webhooks-definitions updater`)

PR #472, opened 2021-02-16
ci(update): await `@octokit/webhooks-definitions` package version to be published to npm before downloading it from the registry

### LINKED ISSUE #470: Wait for NPM to update `@latest` version for the `update` workflow
I've built https://github.com/gr2m/await-npm-package-version-action/ to address the delay between an npm publish and the availablity of the new version in the registry, e.g. see

https://github.com/octokit/openapi-types.ts/blob/ef4ea6bdb24d85cb094bba8751829bbecc835ffd/.github/workflows/update.yml#L34-L37

Let's add that to this repository's update workflow

_Originally posted by @gr2m in https://github.com/octokit/webhooks.js/issues/449#issuecomment-774752812_

wolfy1339:
Fixes #470

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -12,6 +12,10 @@ jobs:
       - uses: actions/setup-node@v2
       - run: npm ci
       - run: npm install --save-exact @octokit/webhooks-definitions@latest
+      - uses: gr2m/await-npm-package-version-action@v1
+        with:
+          package: "@octokit/webhooks-definitions"
+          version: ${{ github.event.client_payload.release.tag_name }}
       - run: npm run generate-types
       - name: create pull request
         uses: gr2m/create-or-update-pull-request-action@v1.x


gr2m (approved):

.github/workflows/update.yml
```diff
@@ -12,6 +12,10 @@ jobs:
       - uses: actions/setup-node@v2
       - run: npm ci
       - run: npm install --save-exact @octokit/webhooks-definitions@latest
+      - uses: gr2m/await-npm-package-version-action@v1
```
G-Rath, line 15: Shouldn't this be *before* we install the package, if its meant to making us wait for the package to be found?

wolfy1339: I believe so

gr2m: Oopsies sorry I should have caught that myself. Great catch @G-Rath

PR #480, opened 2021-02-27
Export `HandlerFunction`

JuanM04:


wolfy1339:
Hi @JuanM04 :wave:,

Can you explain the reasoning behind your changes?

JuanM04:
I was trying to create a handler with TypeScript and `probot`. The only type I couldn't import was this one

wolfy1339:
I see

wolfy1339:
However, you can import it this way:
```ts
import { HandlerFunction } from "@octokit/webhooks/dist-types/types"
```

JuanM04:
Hmm.. I didn't think about that.

That works for me. If you want, you can close this PR

gr2m:
> ```ts
> import { HandlerFunction } from "@octokit/webhooks/dist-types/types"
> ```

I wouldn't consider this public API though. If you import from a path within the package I'd suggest you pin the version. We don't currently plan on it but we might change the structure of the published package in future, at which point the above line would break. And this would be an internal change that would not require a breaking version release, it could happen in a fix version bump

PR #485, opened 2021-03-05
Explicitly specify that TS 4.1 is required

jstastny:
The compilation under TypeScript 4.0 fails.

```
$ npm install typescript@4.0
$ npm run validate:ts


> @octokit/webhooks@0.0.0-development validate:ts /Users/honza/projects/webhooks.js
> tsc --noEmit --noImplicitAny --target es2020 --esModuleInterop --moduleResolution node test/typescript-validate.ts

src/types.ts:12:27 - error TS1110: Type expected.

12 > = TEmitterEvent extends `${infer TWebhookEvent}.${infer TAction}`
                             ~~~

src/types.ts:12:36 - error TS1005: '}' expected.

12 > = TEmitterEvent extends `${infer TWebhookEvent}.${infer TAction}`
                                      ~~~~~~~~~~~~~

src/types.ts:12:49 - error TS1128: Declaration or statement expected.

12 > = TEmitterEvent extends `${infer TWebhookEvent}.${infer TAction}`
                                                   ~

src/types.ts:12:50 - error TS1128: Declaration or statement expected.

12 > = TEmitterEvent extends `${infer TWebhookEvent}.${infer TAction}`
                                                    ~

src/types.ts:12:52 - error TS1005: ';' expected.

12 > = TEmitterEvent extends `${infer TWebhookEvent}.${infer TAction}`
                                                      ~

src/types.ts:12:59 - error TS1005: ';' expected.

12 > = TEmitterEvent extends `${infer TWebhookEvent}.${infer TAction}`
                                                             ~~~~~~~

src/types.ts:53:55 - error TS1005: ';' expected.

53  * Error object with optional properties coming from `octokit.request` errors
                                                         ~~~~~~~

src/types.ts:77:1 - error TS1160: Unterminated template literal.

77



Found 8 errors.

npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! @octokit/webhooks@0.0.0-development validate:ts: `tsc --noEmit --noImplicitAny --target es2020 --esModuleInterop --moduleResolution node test/typescript-validate.ts`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the @octokit/webhooks@0.0.0-development validate:ts script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

```

I am not sure if the `package-json.lock` should be modified as well or not. I have modified it to match the minimum required version and therefore make the continuous integration run on that version, which is IMO an advantage.

PR #487, opened 2021-03-06
ci: test against all supported TS versions

### LINKED ISSUE #486: CI should test against supported versions of TypeScript
_Originally posted by @wolfy1339 in https://github.com/octokit/webhooks.js/pull/485#issuecomment-791747067_

wolfy1339:
Fixes #486
Tested locally using [act](https://github.com/nektos/act)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,6 +24,22 @@ jobs:
           node-version: ${{ matrix.node }}
       - run: npm ci
       - run: npm test
+  typescript_matrix:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        typescript:
+          - 4.1
+          - 4.2
+    name: TypeScript ${{matrix.typescript}}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - run: npm ci
+      - run: npm i typescript@${{matrix.typescript}}
+      - run: npm test
 
   # separate job to set as required in branch protection,
   # as the above change each time the Node versions change


.github/workflows/test.yml
```diff
@@ -24,6 +24,22 @@ jobs:
           node-version: ${{ matrix.node }}
       - run: npm ci
       - run: npm test
+  typescript_matrix:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        typescript:
+          - 4.1
+          - 4.2
```
oscard0m, line 33: Until which version of support do we want to arrive?

.github/workflows/test.yml
```diff
@@ -24,6 +24,22 @@ jobs:
...
+    name: TypeScript ${{matrix.typescript}}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - run: npm ci
+      - run: npm i typescript@${{matrix.typescript}}
+      - run: npm test
```
oscard0m, line 42: Are there other places in Octokit ecosystem where this job would have sense? Could it be interesting to create the GitHub Action and extend it in all the workflows we consider it necessary?

.github/workflows/test.yml
```diff
@@ -24,6 +24,22 @@ jobs:
...
+          - 4.1
+          - 4.2
+    name: TypeScript ${{matrix.typescript}}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - run: npm ci
+      - run: npm i typescript@${{matrix.typescript}}
```
oscard0m, line 41: As mentioned in the issue, is the extra time of this PR Check something to consider? How much time it takes with and without this check?

.github/workflows/test.yml
wolfy1339: Those are the current supported versions. `>=4.1, < 5` (as per `package.json`)

.github/workflows/test.yml
wolfy1339: It would probably make sense to use this in `@octokit/webhooks-definitions` as well. The other projects, it could be useful, but the TS version range needs to be changed per project (usually `>=4.0, < 5`)

.github/workflows/test.yml
wolfy1339: The full jest suite takes around `63s` when I ran the action locally, using [act](https://github.com/nektos/act)

.github/workflows/test.yml
oscard0m: > Those are the current supported versions. `>=4.1, < 5` (as per `package.json`)

Yep, I meant to say if we should consider give support to older versions of TypeScript. Tried to look for official supported versions by TypeScript team but I did not find any, do you know if there is a place with as NodeJS has for example?

.github/workflows/test.yml
oscard0m: > It would probably make sense to use this in `@octokit/webhooks-definitions` as well. The other projects, it could be useful, but the TS version range needs to be changed per project (usually `>=4.0, < 5`)

Sounds like a mission for [Octoherd](https://github.com/octoherd) team 💪🏽 . If you like the idea I can open an issue per each project in Octokit with Typescript and link to an issue from Octoherd to create this script ❤️

.github/workflows/test.yml
wolfy1339: We can't support older versions of TS here or in `@octokit/webhook-definitions`. Since we use features only available in 4.2+

.github/workflows/test.yml
wolfy1339: Sure. Let's maybe wait for @gr2m to see what he thinks before we move forward

.github/workflows/test.yml
oscard0m: > The full jest suite takes around `63s` when I ran the action locally, using [act](https://github.com/nektos/act)

GitHub Actions in production will be faster than that right? (unless you have a NASA 💻  @wolfy1339 🤣 ). To me is an _acceptable_ timing and... we can always invest in improving check run times in a different issue if it becomes a problem

---

Didn't know about `act`. Love it!

.github/workflows/test.yml
oscard0m: > We can't support older versions of TS here or in `@octokit/webhook-definitions`. Since we use features only available in 4.2+

Yep, I was wondering if we would like to support older versions of TS and if it's worth it. Specially for the exposed types to the world. The internal part is fine to use the LTS.

---

Let's go for `TS v4.1` and `TS 4.2` and opening a discussion for giving support to older versions of TS if you think is worth having the discussion :)

.github/workflows/test.yml
wolfy1339: > GitHub Actions in production will be faster than that right?

They seem to be usually around 30-40s
>(unless you have a NASA computer @wolfy1339 :rofl:)

Nah, nothing of the sorts

.github/workflows/test.yml
wolfy1339: These are used in types exported publically.
See the `HandlerFunction` type, it is used by `Webhook#on` function and the type itself uses `EmitterWebhookEvent` which is incompatible with any version lower than `TS 4.1`

oscard0m: for other readers: [this is the feature](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#template-literal-types) which requires `TS >= 4.1`

gr2m:
I'm reluctant to start testing with older versions of libraries. It adds a maintenance overhead and the tests will get out of date ... I'd rather just support the latest version of TypeScript, I don't think it's too much to ask for people to upgrade. And there is no way to define a peer dependency of devDependencies. Even if we test in multiple versions, people will still run into the same problem.

Maybe I'm missing something here, but for now I'd just bump the `typescript` minimal version in `devDependencies`

gr2m:
Basically this PR is good enough from my perspective:
https://github.com/octokit/webhooks.js/pull/485/files

G-Rath:
I agree with @gr2m - while I don't mind bumping the version in `devDependencies`, I'm not sure why the original PR was doing `npm install typescript@4.0` as the entire point of the lock file is to track the exact version to use.

You'd have the same problem doing that with just any dependency on a project 🤷

oscard0m:
I'm ok with closing this PR then and stick supporting Typescript `@latest`

wolfy1339:
Closing per the feedback

PR #492, opened 2021-03-12
WIP fix(types): transform method only accepts current event type

oscard0m:


src/types.ts
```diff
@@ -21,7 +21,10 @@ interface BaseWebhookEvent<TName extends WebhookEventName> {
   payload: WebhookEventMap[TName];
 }
 
-export interface Options<TTransformed = unknown> {
+export interface Options<
+  E extends EmitterWebhookEventName = any,
+  TTransformed extends EmitterWebhookEvent<E> = any
```
oscard0m, line 26: I would like to improve these `any` default values here. Is it possible?

wolfy1339: You can default to `EmitterWebhookEventName`, it's done elsewhere in the code

wolfy1339:
I tested your patch and it doesn't seem to fix the issue at hand

wolfy1339:
I think some next steps would be to check if reverting these PRs fixes the problem
https://github.com/octokit/webhooks.js/pull/459 https://github.com/octokit/webhooks.js/pull/464

oscard0m:
> I think some next steps would be to check if reverting these PRs fixes the problem
> #459 #464

Opened a new PR reverting these 2 PR commits: https://github.com/octokit/webhooks.js/pull/494

Closing this one since it's not fixing the issue, as @wolfy1339 pointed

PR #493, opened 2021-03-13
ci(test): rename jobs in test.yml

oscard0m:
Proposal to rename jobs to better describe what's happening inside each job

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ name: Test
       - opened
       - synchronize
 jobs:
-  test_matrix:
+  test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -27,11 +27,11 @@ jobs:
...
     runs-on: ubuntu-latest
-    needs: test_matrix
+    needs: test
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
       - run: npm ci
-      - run: "npm run validate:ts"
\ No newline at end of file
+      - run: "npm run validate:ts"
```
gr2m, line 37: the "test" job name is significant, it is part of the branch protection for this repository. The main reason why I split up tests into two jobs, is that the names for the node version test matrix change each time we change the versions we want to test:

image

while the final "test" job name remains the same.

image

All this to change, the job names are named that way for the sake of simpler branch protection settings

oscard0m: Created a [post](https://github.community/t/if-a-check-run-name-changes-it-keeps-being-required/167669?u=oscard0m) in GitHub Community to see if we can come up with a better solution. This looks a bit hacky to me but I guess there is no choice for the moment.

oscard0m:
Closing PR and creating an issue in the future depending on this: https://github.com/octokit/webhooks.js/pull/493#discussion_r593740058

PR #494, opened 2021-03-14
fix(typescript): revert #459 & #464

### LINKED ISSUE #490: Passing down event name into `TransformMethod` to allow for the right payload types to be used
I'm trying to fix the issue octokit/app.js#212, and I've hit a wall with types with the `transform` method.

How do I pass down the `TName` type parameter from the `HandlerFunction` type to `TransformMethod`?
The end result wanted is to keep all the same type inference as not having the `transform` method, and just apply the modifications from that function.

In this case, we would like the return type to be `EmiterWebhookEvent<EventName extends EmitterWebhookEventName> & { octokit: Octokit }`

oscard0m:
## 📝 Summary

Reverts the following PR's: 
* https://github.com/octokit/webhooks.js/pull/464
* https://github.com/octokit/webhooks.js/pull/459

## ⛱ Motivation and Context

Types for webhooks are broken when a `transform` function is passed in the options: 
https://github.com/octokit/app.js/pull/213#issuecomment-798928569

Follow up conversation:
https://github.com/octokit/webhooks.js/pull/492#issuecomment-798938503

### DIFF
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -30,7 +30,7 @@ interface EventHandler<TTransformed> {
 }
 
 export function createEventHandler<TTransformed>(
-  options: Options<any, TTransformed>
+  options: Options<TTransformed>
 ): EventHandler<TTransformed> {
   const state: State = {
     hooks: {},
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -30,7 +30,7 @@ interface EventHandler<TTransformed> {
 }
 
 export function createEventHandler<TTransformed>(
-  options?: Options<TTransformed>
+  options: Options<any, TTransformed>
 ): EventHandler<TTransformed> {
   const state: State = {
     hooks: {},
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -15,23 +15,23 @@ import {
 import { receiverHandle as receive } from "./receive";
 import { removeListener } from "./remove-listener";
 
-interface EventHandler<T extends Options> {
+interface EventHandler<TTransformed> {
   on<E extends EmitterWebhookEventName>(
     event: E | E[],
-    callback: HandlerFunction<E, T["transform"]>
+    callback: HandlerFunction<E, TTransformed>
   ): void;
   onAny(handler: (event: EmitterWebhookEvent) => any): void;
   onError(handler: (event: WebhookEventHandlerError) => any): void;
   removeListener<E extends EmitterWebhookEventName>(
     event: E | E[],
-    callback: HandlerFunction<E, T["transform"]>
+    callback: HandlerFunction<E, TTransformed>
   ): void;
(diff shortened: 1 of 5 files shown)


wolfy1339:
This fixes the problem.

What about the scenario brought up in #464? Since we only allow to append to the event and not change it completely.

oscard0m:
> This fixes the problem.

I would like to add a test in `typescript.validate.ts` to cover the case we are fixing so we avoid regressions in the future

> What about the scenario brought up in #464? Since we only allow to append to the event and not change it completely.

I would create an issue (I think there was not one before) so we keep discussing this. What do you think?

wolfy1339:
> I would like to add a test in `typescript.validate.ts` to cover the case we are fixing so we avoid regressions in the future

That should be as easy, just add an event handler for a certain event (ex: `issues`), and then try to access a property that exists only on that event (ex in the `issues` event: `issue`)

> I would create an issue (I think there was not one before) so we keep discussing this. What do you think?

Yes, we should create an issue.

oscard0m:
> > I would like to add a test in `typescript.validate.ts` to cover the case we are fixing so we avoid regressions in the future
> 
> That should be as easy, just add an event handler for a certain event (ex: `issues`), and then try to access a property that exists only on that event (ex in the `issues` event: `issue`)

We have this already, and seems it did not prevent us to break types:

https://github.com/octokit/webhooks.js/blob/01c7655b556fea9ed0b89aeb7a9afdf8f2acc2cc/test/typescript-validate.ts#L164-L166

This is the test example you were suggesting right?

wolfy1339:
That doesn't have the `transform` method, but it is the exact sort of test that needs to be added. Look further down, there is another section of the tests where we have the `transform` method

https://github.com/octokit/webhooks.js/blob/01c7655b556fea9ed0b89aeb7a9afdf8f2acc2cc/test/typescript-validate.ts#L179-L182

oscard0m:
Ok, but with changes introduced by this PR it passes the `transform` method in [L75](https://github.com/octokit/webhooks.js/pull/494/files#diff-f0e1dffba6fc83b2e2bfe81aeab3e340f2b50b5cbf6aac225bfc5dc2536ea4eaR75) to all following lines so we are covering the use case now, right?

wolfy1339:
After a second look, I can say that that should do the trick for the purposes of this PR

oscard0m:
> After a second look, I can say that that should do the trick for the purposes of this PR

Added a comment linking to this PR for that line to give context for other developers before modifying it in the future

G-Rath (approved):
> Added a comment linking to this PR for that line to give context for other developers before modifying it in the future

Per my comment [here](https://github.com/octokit/webhooks.js/issues/495#issuecomment-798968014), at some point we should modify the runtime code to prevent this as much as possible.

wolfy1339 (approved):

wolfy1339:
Closes octokit/app.js#213

PR #509, opened 2021-03-26
feat: `createNodeMiddleware()`

### LINKED ISSUE #507: Replace `webhooks.middleware` with `createNodeMiddleware()`
**What’s missing?**

For better consistency with other modules (compare https://github.com/octokit/app.js/, https://github.com/octokit/oauth-app.js/), better compatibility with browsers, and better composability, we should

1. Export a `createNodeMiddleware(app, { path })` function
2. Deprecate `webhooks.middleware` and the [`path` constuctor option](https://github.com/octokit/webhooks.js/#constructor)
3. Default the webhooks path to `/api/github/webhooks`, for better compatibility with the defaults of https://github.com/octokit/oauth-app.js#createnodemiddlewareapp-options

**Why?**

http request routing is not a concern of the `@octokit/webhooks` library. When it was created, long-living node processes were the norm. Today, serverless enviroments gain more and more traction, and they come with their own http routing. Moving it out of the core library will make it easier to use in serverless environments, and easier to compose with other libraries, such as https://github.com/octokit/app.js/, https://github.com/octokit/octokit.js, and github.com/probot/probot/

**Alternatives you tried**

It's possible to use & compose the library as is today, but it's unnecessary complicated.

### LINKED ISSUE #510: Document `log` option
this is a follow up to https://github.com/octokit/webhooks.js/pull/457 @jablko

`new Webhooks({ log })` was implemented in #457, but not documented.

gr2m:
fixes #507. fixes #510
Part of https://github.com/octokit/octokit.js/pull/15

### TODOs

- [x] implement `log` option: `createNodeMiddleware(webhooks, { log })`

-----
[View rendered README.md](https://github.com/octokit/webhooks.js/blob/507/createNodeMiddleware/README.md)
[View rendered src/middleware-legacy/README.md](https://github.com/octokit/webhooks.js/blob/507/createNodeMiddleware/src/middleware-legacy/README.md)


src/middleware/node/index.ts
```diff
@@ -0,0 +1,18 @@
+import { Webhooks } from "../../index";
+import { EmitterWebhookEvent } from "../../types";
+import { middleware } from "./middleware";
+import { onUnhandledRequestDefault } from "./on-unhandled-request-default";
+import { MiddlewareOptions } from "./types";
+
+export function createNodeMiddleware(
+  webhooks: Webhooks<EmitterWebhookEvent>,
```
gr2m: @G-Rath @oscard0m  @wolfy1339 sorry I think we discussed this recently, I've been working on other parts of @octokit the past few weeks. Is there a reason why I have to set a type if I want to use the `Webhooks` constructor in TypeScript code? I don't think that should be the case. Is there a reason why we don't set a default?

wolfy1339: This seems to have been an oversight in #494, it was defaulted to `unknown` before v8

src/middleware/node/middleware.ts
```diff
@@ -0,0 +1,88 @@
...
+  // GitHub will abort the request if it does not receive a response within 10s
+  // See https://github.com/octokit/webhooks.js/issues/185
+  let didTimeout = false;
+  const timeout = setTimeout(() => {
+    didTimeout = true;
+    response.statusCode = 202;
+    response.end("still processing\n");
+  }, 9000).unref();
+
+  return getPayload(request)
```
G-Rath: nit: I think this'd be better with `async/await`

test/integration/node-middleware.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        headers: {
+          "X-GitHub-Delivery": "123e4567-e89b-12d3-a456-426655440000",
+          "X-GitHub-Event": "push",
+          "X-Hub-Signature-256": signatureSha256,
+        },
+        body: JSON.stringify(pushEventPayload),
+      }
+    );
+
+    expect(response.status).toEqual(200);
```
G-Rath, line 46:
```suggestion
    expect(response.status).toBe(200);
```

nit: prefer using `toBe` for literal matchers (or otherwise `toStrictEqual`)/

test/integration/node-middleware.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(await response.text()).toEqual("ok\n");
```
G-Rath:
```suggestion
    await expect(response.text()).resolves.toBe("ok\n");
```

nit: `await expect(...).resolves` fails nicer, as `expect(await ...)` will fail the test with "an uncaught error was thrown" whereas this fails it with "expected promise to resolve, but it rejected with ...".

test/integration/node-middleware.test.ts
```diff
@@ -0,0 +1,90 @@
...
+import { pushEventPayload } from "../fixtures";
+
+const signatureSha256 = sign(
+  { secret: "mySecret", algorithm: "sha256" },
+  JSON.stringify(pushEventPayload)
+);
+
+describe("createNodeMiddleware(webhooks)", () => {
+  it("README example", async () => {
+    const webhooks = new Webhooks<EmitterWebhookEvent>({
```
G-Rath:
```suggestion
  it("README example", async () => {
    expect.assertions(3);
  
    const webhooks = new Webhooks<EmitterWebhookEvent>({
```

as you've got an expect in an event (`webhooks.on("push", (event) => {`)

test/integration/node-middleware.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      secret: "mySecret",
+    });
+
+    webhooks.on("push", (event) => {
+      expect(event.id).toBe("123e4567-e89b-12d3-a456-426655440000");
```
G-Rath, line 25: because this is async, ideally you should use the optional `done` callback or (preferably) create a promise that gets resolved in this event, which you can `await` at the end of the test body.

test/integration/node-middleware.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    );
+
+    expect(response.status).toEqual(200);
+    expect(await response.text()).toEqual("ok\n");
+
+    server.close();
+  });
+
+  test("request.body already parsed (e.g. Lambda)", async () => {
+    expect.assertions(2);
```
G-Rath:
```suggestion
    expect.assertions(3);
```

you've got three assertions in this test - ones in the webhook `on` event.

test/integration/node-middleware.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        headers: {
+          "X-GitHub-Delivery": "123e4567-e89b-12d3-a456-426655440000",
+          "X-GitHub-Event": "push",
+          "X-Hub-Signature-256": signatureSha256,
+        },
+        body: JSON.stringify(pushEventPayload),
+      }
+    );
+
+    expect(response.status).toEqual(200);
```
G-Rath, line 89:
```suggestion
    expect(response.status).toBe(200);
```

nit: prefer using `toBe` for literal matchers (or otherwise `toStrictEqual`)

test/integration/node-middleware.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(await response.text()).toEqual("ok\n");
```
G-Rath, line 90:
```suggestion
    await expect(response.text()).resolves.toBe("ok\n");
```

nit: `await expect(...).resolves` fails nicer, as `expect(await ...)` will fail the test with "an uncaught error was thrown" whereas this fails it with "expected promise to resolve, but it rejected with ...".

src/middleware/node/middleware.ts
```diff
@@ -0,0 +1,88 @@
...
+    .catch((error: WebhookEventHandlerError) => {
+      clearTimeout(timeout);
+
+      console.log(`error`);
+      console.log(error);
+
+      if (didTimeout) return;
+
+      const statusCode = Array.from(error)[0].status;
+      response.statusCode = statusCode || 500;
```
G-Rath:
```suggestion
      response.statusCode = statusCode ?? 500;
```

nit: since the `statusCode` could in theory be 0.

G-Rath (commented):
Have left a few nity comments - the test ones are a bit more nity (a result of being a core maintainer of `eslint-plugin-jest` 😅)

test/integration/node-middleware.test.ts
gr2m: Good to know, thank you!

src/middleware/node/index.ts
gr2m: created a PR here #514

test/integration/node-middleware.test.ts
gr2m: could you elaborate, I'd like to better understand your point. I use `.equalTo` for comparing simple types such as numbers. I don't think there is a difference to `.toStrictEqual()` when comparing numbers?

test/integration/node-middleware.test.ts
gr2m: I'm not sure if I follow you here. The event handler is guaranteed to be executed before sending a response, as long as it doesn't take more than 9s, so the code should be executed before the `await fetch` below resolves

gr2m:
This is ready for review. Review with the "ignore whitespace" option enabled as I did some unrelated changes in the README which changed lots of indentation 

https://github.com/octokit/webhooks.js/pull/509/files?diff=unified&w=1

gr2m:
hold on I forgot to implement deprecations for the `path` option and `octokit.webhooks`

gr2m:
@octokit/maintainers this is good to go now

Ignore white space when reviewing changes: https://github.com/octokit/webhooks.js/pull/509/files?w=1

src/middleware/node/get-missing-headers.ts
```diff
@@ -0,0 +1,12 @@
+import { IncomingMessage } from "http";
+
+const WEBHOOK_HEADERS = [
+  "x-github-event",
+  "x-hub-signature-256",
```
gr2m, line 5: I intentionally left out the `x-hub-signature` header here, we want to nudge users to use best practices. GHES support is good enough for the new header

src/index.ts
```diff
@@ -53,21 +59,38 @@ class Webhooks<TTransformed = unknown> {
...
+      return middleware(state, request, response, next);
+    };
   }
 }
 
 /** @deprecated `createWebhooksApi()` is deprecated and will be removed in a future release of `@octokit/webhooks`, please use the `Webhooks` class instead */
 const createWebhooksApi = <TTransformed>(options: Options<TTransformed>) => {
-  console.error(
+  const log = createLogger(options.log);
+  log.warn(
```
wolfy1339, line 93: This isn't strictly related to this PR, but I get the changes

test/unit/deprecation.test.ts
```diff
@@ -1,12 +1,67 @@
...
+        info: () => {},
+        warn: warn,
+        error: () => {},
+      },
+    });
+    expect(warn).toBeCalledWith(
       "[@octokit/webhooks] `createWebhooksApi()` is deprecated and will be removed in a future release of `@octokit/webhooks`, please use the `Webhooks` class instead"
     );
-    spy.mockClear();
+    warn.mockClear();
```
wolfy1339, line 19: Not strictly related to this PR.

wolfy1339 (commented):
Nothing stands out to me.

There were 2 changes that weren't strictly related to this PR

src/index.ts
gr2m: yeah sorry for that, you are totally right, I've been lazy creating separate PRs for that. Same with the formatting changes in the README, I've to show `octokit` on Wednesday, I'm in a bit of a rush

gr2m:
I have to go ahead and merge this, but please review and let me know if I missed something, I'll fix it in a follow up PR.

I'll also start a beta release PR for v9 that I can use for `@octokit/app` and `octokit`

honnix:
Thanks for making the change!

With this change, what is the correct way to set middleware to express app?

The deprecated way as following still works, that's great.
```js
const { Webhooks, createNodeMiddleware } = require("@octokit/webhooks");
const webhooks = new Webhooks({
  secret: "mysecret",
  path: "/webhooks",
});

webhooks.onAny(({ id, name, payload }) => {
  console.log(name, "event received");
});

const express = require("express");
const app = express();

app.use(webhooks.middleware);

app.get("/", function (req, res) {
  res.send("hello world");
});

app.listen(3000);
```

However I tried the following, but it doesn't work well. It seems the middleware hijacks all routes and I got
```shell
$ curl localhost:3000/
{"error":"Required headers missing: x-github-event, x-hub-signature-256, x-github-delivery"}% 
```
```js
const { Webhooks, createNodeMiddleware } = require("@octokit/webhooks");
const webhooks = new Webhooks({
  secret: "mysecret",
});

webhooks.onAny(({ id, name, payload }) => {
  console.log(name, "event received");
});

const express = require("express");
const app = express();

app.use(createNodeMiddleware(webhooks, { path: "/webhooks" }));

app.get("/", function (req, res) {
  res.send("hello world");
});

app.listen(3000);
```

I tried to do `app.use` after setting up routes then it works. Is this intended?

gr2m:
> I guess the new middleware stops forwarding to next route

that was an unintended change, it slipped due to lack of tests, sorry for that! And thanks for the PR 👍🏼

PR #511, opened 2021-03-27
feat: deprecate `createWebhooksApi()`

### LINKED ISSUE #454: Deprecate `createWebhooksApi`
The `createWebhooksApi` export is there for legacy reasons. We use constructors across the other @octokit modules

https://github.com/octokit/webhooks.js/blob/1891034b602a2193a08f1ac3ad31775df65ca578/src/index.ts#L69

We should deprecate `createWebhooksApi` by exporting a function with that name that logs a deprecation message and returns a `Webhooks` instance with the same options, then remove it as v9. 

I like creating a dedicated `deprecations-test.ts` file where we test that the deprecation message is logged and that the API still works. If we stay consistent with that flow, removing deprecated APIs as part of a breaking release becomes trivial

wolfy1339:
Fixes #454

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -65,7 +65,13 @@ class Webhooks<TTransformed> {
   }
 }
 
-const createWebhooksApi = Webhooks.prototype.constructor;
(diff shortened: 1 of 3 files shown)


gr2m (approved):
Thanks!

PR #518, opened 2021-03-30
feat: v9

### LINKED ISSUE #475: Project: Web platform compatibility (Browsers, Deno, Cloudflare Workers)
`@octokit/webhooks` can already be used in browser's today, but the build size is huge, because [Node's `crypto` API](https://nodejs.org/api/crypto.html) needs to be polyfilled. We need to create a universal library to sign a JSON payload, which uses the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) for browser and Deno builds. This will make `@octokit/webhooks` and other libraries and framework that use it such as `@octokit/app` and `probot` work better with Cloudflare Workers. It is working today ([example](https://github.com/gr2m/cloudflare-worker-github-app-example)), but the build size could be reduced by 90%+. Another API change that would improve `@octokit/webhooks`'s compatibility with browser/Deno/Cloudflare Workers is to remove the `webhooks.middleware` API, and replace it with a dedicated `createNodeMiddleware` export, similar to what we do with [`@octokit/app`](https://github.com/octokit/app.js/#getnodemiddlewareapp-options).

I created a similar package for the JSON Web Token authentication used by GitHub Apps: [`universal-github-app-jwt`](https://github.com/gr2m/universal-github-app-jwt/).

The challenge for this module is that different environment should receive different code. In the past, the `"browse"` package field was used by several build tools. But with the coming support of ES Modules by Node.js, we have a convention that the build tools are adapting now, which is to use the `"exports"` key, see https://docs.skypack.dev/package-authors/package-checks#export-map

It might be a good time to just create a native ES Module library without any build step at this point, so that we have full control over the `package.json` file published to npm. The `@pika/pack` build tool we are currently using in @octokit does not support the `"exports"` key at this point, and the module is currently not maintained by the Pika/Snowpack team

### LINKED ISSUE #476: docs: Recommend to use log client
I would not use it, it increases the bundle size, and I want `@octokit/webhooks` to work well in contraint environments such as browsers and Cloudflare Workers in future.

We can recommend to use it to users as we do in `@octokit/core`: https://github.com/octokit/core.js#logging

_Originally posted by @gr2m in https://github.com/octokit/webhooks.js/pull/201#discussion_r578752175_

### LINKED ISSUE #477: remove "debug" property
We introduced a `log` option as a replacement for the `debug` module in https://github.com/octokit/webhooks.js/pull/457. As part of the next breaking release, we shoudl remove the `debug` module altogether.

### LINKED ISSUE #515: remove deprecated `webhooks.middleware`
follow up to https://github.com/octokit/webhooks.js/pull/509

gr2m:
### ⚠️ do not rebase this pull request, it will break the beta release flow for `semantic-release`

I've pushed a [`beta-rebased`](https://github.com/octokit/webhooks.js/compare/master...beta-rebased?expand=1) branch which resolves the conflicts. 

fixes #477, fixes #515, fixes #476, fixes #475

### Breaking changes

- default `sign` algorithm for `webhooks.sign` is now `sha256`. Use `webhooks.sign({ algorithm: "sha1", secret }, payload)` to fall back to `sha1`.
- static `sign` and `verify` methods are no longer exported. Use `@octokit/webhooks-methods` package instead
- `webhooks.sign()` and `webhooks.verify()` are now asynchronous

### TODOs

- [x] default sign algorithm to `sha256`
- [x] Make `webhooks.sign` and `webhooks.verify` async, for compatibility with Browsers and Deno
- [x] Remove exports of standalone `sign` and `verify` methods

-----
[View rendered src/middleware-legacy/README.md](https://github.com/octokit/webhooks.js/blob/beta/src/middleware-legacy/README.md)


gr2m:
v9 beta.1 release: https://github.com/octokit/webhooks.js/releases/tag/v9.0.0-beta.1

gr2m:
v9 beta.2 release: https://github.com/octokit/webhooks.js/releases/tag/v9.0.0-beta.2

gr2m:
This is ready to be merged. I've pushed a [`beta-rebased`](https://github.com/octokit/webhooks.js/compare/master...beta-rebased?expand=1) branch which resolves the conflicts. 

@octokit/maintainers anyone of you would like to give this a review?

wolfy1339:
Fixes #476 by removing the deprecated middleware

gr2m:
v9 beta.4 release: https://github.com/octokit/webhooks.js/releases/tag/v9.0.0-beta.4

I tested it in the browser console with

```js
pkg = await import("https://cdn.skypack.dev/@octokit/webhooks@9.0.0-beta.4")
const webhooks = new pkg.Webhooks({
  secret: "secret"
})
await webhooks.sign("payload")
```

and it works great

gr2m:
Unless there are any objects, I'd like to release this on Tuesday

PR #519, opened 2021-03-30
fix: use `options.onUnhandledRequest` in `createNodeMiddleware(webhooks, options)`

gr2m:


src/middleware/node/middleware.ts
```diff
@@ -21,8 +20,7 @@ export async function middleware(
   const isUnknownRoute = request.method !== "POST" || pathname !== options.path;
   const isExpressMiddleware = typeof next === "function";
   if (!isExpressMiddleware && isUnknownRoute) {
-    options.log.debug(`not found: ${request.method} ${request.url}`);
```
gr2m, line 24: the log was not a good idea here, because we use `options.onUnhandledRequest` to combine middlewares

wolfy1339 (approved):

PR #521, opened 2021-03-30
fix: "Unknown route: POST /"

jmopel:
If just omit the options as in the example, the server will return GitHub this error for any webhooks: [404] "Unknown route: POST /"

-----
[View rendered README.md](https://github.com/Jmopel/webhooks.js/blob/patch-1/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ webhooks.onAny(({ id, name, payload }) => {
   console.log(name, "event received");
 });
 
-require("http").createServer(createNodeMiddleware(webhooks)).listen(3000);
+require("http").createServer(createNodeMiddleware(webhooks, { path: "/" })).listen(3000);
 // can now receive webhook events at /api/github/webhooks
 ```
 


gr2m:
The comment below states that the path is `/api/github/webhooks`, not `/`, by default?

jmopel:
Oh, I didn't pay attention to this comment, sorry. With the previous `webhooks.middleware` it was `/` by default. So after the upgrade, some inattentive people like me today may have the same problem 😊

gr2m:
> Oh, I didn't pay attention to this comment, sorry. With the previous `webhooks.middleware` it was `/` by default. So after the upgrade, some inattentive people like me today may have the same problem 😊

yes, I fear that will happen. We want to normalize the `/api/github/*` routes to make it easier to move between the different @octokit modules, and between a long living server process and serverless deployment environments. There is only one route for webhooks, but a bunch for OAuth. Here is a list: https://github.com/octokit/octokit.js/#app-server

PR #524, opened 2021-04-01
fix(types): allow `IncomingMessage` to have bodies of other types

G-Rath:
This is more accurate, as we don't actually know for sure that `body` will be our type; this also should stop conflicts with other packages that do this.


wolfy1339 (dismissed):

G-Rath:
@wolfy1339 could I get a quick reapprove?

wolfy1339:
Of course, I was just waiting on the tests, and doing other things at the same time

wolfy1339 (approved):

PR #534, opened 2021-04-15
fix: pass on to the next middleware in case of express

honnix:
As I explained [here](https://github.com/octokit/webhooks.js/pull/509#issuecomment-811379891), it is now required to always install this middleware to the end of the chain, otherwise it rejects all requests it cannot handle.

### DIFF
--- a/src/middleware/node/middleware.ts
+++ b/src/middleware/node/middleware.ts
@@ -25,7 +25,7 @@ export async function middleware(
   const isExpressMiddleware = typeof next === "function";
   if (isUnknownRoute) {
     if (isExpressMiddleware) {
-      return next();
+      return next && next();
     } else {
       return options.onUnhandledRequest(request, response);
     }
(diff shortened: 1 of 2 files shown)


honnix:
@gr2m PTAL and help me understand whether this is a valid change. Thanks a lot.

honnix:
Sorry I couldn't figure out how to add a correct label. I guess this is a `bug`.

gr2m (approved):
This looks great, thank you @honnix 👍🏼 I'm not very familiar with express, if there is anything we can do to make it easier to use for express users, please let us know.

I also made follow up issue to make sure we test the express integration in `@octokit/oauth-app` and `@octokit/app`:  https://github.com/octokit/app.js/issues/236, https://github.com/octokit/oauth-app.js/issues/222

honnix:
@gr2m Thanks for approving and merging. I cannot say myself familiar with express either but I thinkthe added test cases should help catch similar problem in the future.

PR #539, opened 2021-04-20
feat: update package to use new packages split from `@octokit/webhooks-defintions`

wolfy1339:
I updated the package to use the newly published packages that were split from `@octokit/webhooks-definitions`.

`@octokit/webhooks-schemas` was placed in `devDependencies` as it is not used in thew local package.
`@octokit/webhooks-types` was added in `dependencies` as it is used in published code.


gr2m (dismissed):
tests need to be fixed, otherwise looking good 👍🏼

gr2m (approved):

PR #548, opened 2021-04-24
refactor: use `Array#includes` instead of `indexOf`

wolfy1339:
This is #463 but rebased, and up to date.

Optional chaining changes were reverted due to support being introduced into browsers recently


src/event-handler/on.ts
```diff
@@ -11,10 +11,7 @@ function handleEventHandlers(
   webhookName: EmitterWebhookEventName | "error" | "*",
   handler: Function
 ) {
-  if (!state.hooks[webhookName]) {
-    state.hooks[webhookName] = [];
-  }
-
+  state.hooks[webhookName] ||= [];
```
G-Rath: You'll want to revert this too, as it's even newer than optional chaining

G-Rath (commented):
Has the `if` -> `return` refactor already happened?

src/event-handler/on.ts
wolfy1339: Yes, but TS transpliles it down to ES2019, for all platforms.

The problem was with optional chaining showing up in browser builds when support for it was just recently introduced (early 2020)

wolfy1339:
> Has the `if` -> `return` refactor already happened?

The middleware was rewritten/replaced and it doesn't seem to be an issue anymore

src/event-handler/on.ts
G-Rath: But then won't we have the same issue when browsers start supporting this syntax too?

gr2m: It's currently transpiled to 

```js
  var _a;

  (_a = state.hooks)[webhookName] || (_a[webhookName] = []);
```

using the `es2020` target from https://github.com/octokit/tsconfig.

So I think we are good?

wolfy1339: The current tooling creates a build optimized for bundlers and tooling, which is not optimised to be used in the browser directly. It bases these builds off of the `dist-src` builds produced by TypeScript (which is set to output ES2020), and then passes it into `rollup`, which basically creates a bundle directly from that build and **doesn't do any transpiling.**

`@pika/plugin-bundle-web` (requires `@pika/plugin-build-web`) **does the transpiling** using `babel` with the `@babel/preset-env` plugin, for which we can specify the targets that we want using the `.babelrc` file at the root of the repository. It defaults to browsers that support ESM, which [this table over at MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#browser_support) explains

gr2m: honestly I'd probably just revert that change and leave the other one. We'll change the whole build setup anyway eventually, and then we can take the opportunity to use more modern ES features

```suggestion
  if (!state.hooks[webhookName]) {
    state.hooks[webhookName] = [];
  }

```

gr2m (approved):

PR #563, opened 2021-05-10
chore: pin prettier version and fix "Update Prettier" workflow

wolfy1339:


wolfy1339:
I made an octoherd script to partly automate this: https://github.com/wolfy1339/octoherd-script-fix-prettier-update

It doesn't currently check for any open PRs though, or that the file has already been updated or is correct

gr2m (approved):

PR #586, opened 2021-06-17
feat:`octokit.verifyAndReceive()` accepts raw string payload

### LINKED ISSUE #420: `verifyAndReceive` should accept raw string payload
**What happened?**
Using `verifyAndReceive` method with a payload in `string` format does not work

Details here: https://github.com/probot/probot/pull/1467#issuecomment-767944136

**What did you expect to happen?**
`verifyAndReceive` should accept the raw `payload` payload string besides the parsed object

**What the problem might be**
To improve signature of `verifyAndReceive` to support both types

gr2m:
fixes #420

-----
[View rendered README.md](https://github.com/octokit/webhooks.js/blob/420/verifyAndReceive-with-string-payload/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,7 +260,7 @@ webhooks.verifyAndReceive({ id, name, payload, signature });
           payload
         </code>
         <em>
-          Object
+          Object or String
         </em>
       </td>
       <td>
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ webhooks.verify(eventPayload, signature);
           eventPayload
         </code>
         <em>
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -189,7 +189,7 @@ webhooks.verify(eventPayload, signature);
           eventPayload
         </code>
         <em>
-          (Object)
+          (Object or String)
```
gr2m, line 192: this was already supported, just not documented

wolfy1339 (approved):

PR #587, opened 2021-06-17
fix: bump `@octokit/webhooks-methods` to `^2.0.0`

gr2m:
follow up to https://github.com/octokit/webhooks-methods.js/pull/21

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1644,9 +1644,9 @@
       }
     },
     "@octokit/webhooks-methods": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/@octokit/webhooks-methods/-/webhooks-methods-1.0.0.tgz",
-      "integrity": "sha512-pVceMQcj9SZ5p2RkemL0TuuPdGULNQj9F3Pq1cNM1xH+Kst1VNt0dj3PEGZRZV473njrDnYdi/OG4wWY9TLbbA=="
+      "version": "2.0.0",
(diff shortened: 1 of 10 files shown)


wolfy1339 (approved):

rr-codes:
@wolfy1339 This is a breaking change, and led to all my GitHub payloads breaking. This is because `@octokit/webhooks-methods` had a breaking change requiring the payloads to be `string` instead of JSON objects, both of which worked fine until this PR, and now only `string` works.

rr-codes:
for context, I am using Express, and had this middleware:


```js
app.use((req, res, next) => {
  if (req.originalUrl.includes('/stripe/webhook')) {
    bodyParser.raw({ type: 'application/json' })(req, res, next);
  } else {
    bodyParser.json()(req, res, next);
  }
});
```

However, with this PR, I had to change it to
```js
app.use((req, res, next) => {
  if (req.originalUrl.includes('/stripe/webhook')) {
    bodyParser.raw({ type: 'application/json' })(req, res, next);
  } else if (req.originalUrl.includes('/webhooks/github')) {
    bodyParser.text({ type: 'application/json' })(req, res, next);
  } else {
    bodyParser.json()(req, res, next);
  }
});
```

wolfy1339:
It was a breaking change in the `@octokit/webhools-methods` package, we added a compatibility layer to this package in the meantime until the change is implemented in this package as well.

rr-codes:
@wolfy1339 Ya I saw the changes, but the compatibility layer didn't seem to have any effect

wolfy1339:
Could you open a new issue with all the details, and we can start looking from there

rr-codes:
sure, will do

PR #623, opened 2021-09-01
fix(types): add ability to remove onAny listeners again

### LINKED ISSUE #622: Can not remove onAny listener
**What happened?**

- Updated to the latest version (9.12.0)
- Can no longer remove `onAny` based listeners

```
error TS2345: Argument of type '"*"' is not assignable to parameter of type '"push" | "release" | "public" | "label" | "meta" | "status" | "check_run" | "check_suite" | "code_scanning_alert" | "commit_comment" | "content_reference" | "create" | "delete" | ... 213 more ... | ("push" | ... 224 more ... | "workflow_run.requested")[]'.

100       this.webhooks.removeListener('*', this.webhooksListener);
```

wolfy1339:


gr2m (commented):
The changes work for me 👍🏼 Could you also update the README to document how to remove an `onAny` event listener?

wolfy1339:
> Could you also update the README to document how to remove an onAny event listener?

It is already documented. https://github.com/octokit/webhooks.js#webhooksremovelistener

wolfy1339:
The `validate:ts` test seems to be stalling

wolfy1339:
Something is definitely up with this patch. It makes typescript run forever during the `validate:ts` script.

I'm not exactly sure what's going on here.

gr2m:
I tried upgrading typescript and adding `--strict` but neither helped. I looked a bit into the code changes and tried some alternative approaches but no dime :(

wolfy1339:
@G-Rath @oscard0m Do you guys have any idea how to remedy this situation?

G-Rath:
@wolfy1339 this looks like we're just doing a lot of work (which isn't surprising) because of how large the union is - there's not really much we can do aside from changing the type in a way that's looser so that less work is done.

wolfy1339:
Closing in favor of #645

PR #626, opened 2021-09-03
feat: export emitterEventNames

### LINKED ISSUE #550: List of possible events in webhooks is not available
**What’s missing?**
There is `WebhookEventMap` type that contains mapping of events to their spec. There is also `WebhookEventName` which is a string literal type containing all possible Webhook Events that can be checked during ts compilation. However, there is no list of all available events, e.g.: as array of strings: `const WebhookEventNames=[...] as const;`

**Why?**
This is needed to provide list of available events in GUI or validate provided event in runtime.

**Alternatives you tried**

I tried to implement it in `octokit-types.ts` but I realized it only generates `schema.d.ts` which does not allow to create enums or arrays of strings.

I think, there should be an additional file containing this definition. @wolfy1339 do you have an idea how should it be implemented?

pawelmitka:
`emitterEventNames` is a list of possible webhook event names that is useful e.g.: in GUI or validation of input. Therefore, it is very useful also outside of Octokit. Hence it has been exported.

This PR fixes: https://github.com/octokit/webhooks/issues/550


src/index.ts
```diff
@@ -16,6 +16,7 @@ import {
 } from "./types";
 
 export { createNodeMiddleware } from "./middleware/node/index";
+export { emitterEventNames } from "./generated/webhook-names";
```
gr2m, line 19: Can you add a test to https://github.com/octokit/webhooks.js/blob/master/test/integration/smoke-test.ts? 

I think we should also document the export in the README.

pawelmitka: sure, will do 😄

wolfy1339 (dismissed):
LGTM

README.md
```diff
@@ -694,6 +696,11 @@ If there are actions for a webhook, events are emitted for both, the webhook nam
 
 <!-- /autogenerated via scripts/generate-types.ts -->
 
+### emitterEventNames
+
+The read only tuple containing all the possible webhook event names listed above.
+This might be useful in GUI and input validation.
```
gr2m: It's not only the event names, but the combination of event names and actions. How about this?

````suggestion
A read only tuple containing all the possible combinations of the webhook events + actions listed above. This might be useful in GUI and input validation.

```js
import { emitterEventNames } from "@octokit/webhooks";
emitterEventNames // ["check_run", "check_run.completed", ...]
```
````

File: test/integration/smoke-test.ts [hunk 3]
```diff
@@ -23,3 +23,13 @@ test('require("@octokit/webhooks").createEventHandler', () => {
...
 });
+
+test('require("@octokit/webhooks").emitterEventNames', () => {
+  const emitWarningSpy = jest.spyOn(process, "emitWarning");
+
+  const allEvents = emitterEventNames;
+  expect(typeof allEvents).toBe("object");
+  expect(typeof allEvents[0]).toBe("string");
+
+  expect(emitWarningSpy).not.toHaveBeenCalled();
```
gr2m: no need to spy on process.emitWarning for this test

```suggestion
  const allEvents = emitterEventNames;
  expect(typeof allEvents).toBe("object");
  expect(typeof allEvents[0]).toBe("string");
```

gr2m (changes requested):

gr2m (dismissed):
perfect, thank you 💐

wolfy1339 (approved):
:+1:

PR #645, opened 2021-10-04
fix(types): Add ability to remove onAny listeners again

daniel-beard:
- Attempt a fix for https://github.com/octokit/webhooks.js/issues/622 that doesn't blow up the type checking duration.
- A few key differences from the original PR at: https://github.com/octokit/webhooks.js/issues/622 (I've commented in the diff where these differ)

### DIFF
--- a/src/event-handler/remove-listener.ts
+++ b/src/event-handler/remove-listener.ts
@@ -2,7 +2,7 @@ import { EmitterWebhookEventName, State } from "../types";
 
 export function removeListener(
   state: State,
-  webhookNameOrNames: EmitterWebhookEventName | EmitterWebhookEventName[],
+  webhookNameOrNames: "*" | EmitterWebhookEventName | EmitterWebhookEventName[],
   handler: Function
 ) {
   if (Array.isArray(webhookNameOrNames)) {
--- a/src/types.ts
+++ b/src/types.ts
@@ -45,6 +45,11 @@ export type HandlerFunction<
   TTransformed
 > = (event: EmitterWebhookEvent<TName> & TTransformed) => any;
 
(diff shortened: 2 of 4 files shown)


src/event-handler/remove-listener.ts
```diff
@@ -2,7 +2,7 @@ import { EmitterWebhookEventName, State } from "../types";
 
 export function removeListener(
   state: State,
-  webhookNameOrNames: EmitterWebhookEventName | EmitterWebhookEventName[],
+  webhookNameOrNames: '*' | EmitterWebhookEventName | EmitterWebhookEventName[],
```
daniel-beard: This differs from the original proposed fix in that I omit the following type union:

```
| (EmitterWebhookEventName | "*")[]
```

Which I think is fine here, as you want to remove everything OR individual event names. I don't see a use case for ever specifying both like:

```
webhooks.removeListener(["check_run.created", "*"], ...
```

src/types.ts
```diff
@@ -45,6 +45,11 @@ export type HandlerFunction<
   TTransformed
 > = (event: EmitterWebhookEvent<TName> & TTransformed) => any;
 
+export type RemoveHandlerFunction<
```
daniel-beard, line 48: Through experimentation, I found that the changes to `HandlerFunction` here: https://github.com/octokit/webhooks.js/pull/623/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR43 were causing the blowup in duration.

I made `removeListener` have it's own variation `RemoveHandlerFunction`

wolfy1339 (approved):
Thanks alot for figuring this out

PR #694, opened 2022-06-27
ci: adds codeql-analysis.yml for codeql execution

nickfloyd:
This PR:

- enables codeql for this repo to meet the security and stability requirements of the octokit org.
- Removes the error and stack from being returned from the middleware
- Logs errors to stdout via logger

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'javascript' ]
+        language: [ 'typescript' ]
         # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
         # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
 
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+  schedule:
+    - cron: '28 11 * * 6'
+
+jobs:
+  analyze:
+    name: Analyze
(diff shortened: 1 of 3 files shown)


G-Rath (dismissed):
I think both of what's been flagged are false positives (the one about `$` definitely is)

gr2m (dismissed):
we can make the replace a global replace to appease CodeQL

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,72 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
```
wolfy1339: Shouldn't this be `typescript`

nickfloyd: 👍 Great catch, I'll make sure that gets updated.  Thanks!

nickfloyd:
```suggestion
        language: [ 'typescript' ]
```

timrogers:
It feels like it would be reasonable to change `response.end(String(error));`. There isn't really any need for that to expose quite so much information, although I don't consider it to be a big deal by any means.

nickfloyd:
> It feels like it would be reasonable to change `response.end(String(error));`. There isn't really any need for that to expose quite so much information, although I don't consider it to be a big deal by any means.

I could definitely go to the logging to server log route.  I didn't want to assume I knew why that was there (Chesterton's fence and all 😉).  Perhaps, @gr2m, @wolfy1339, or @G-Rath would have more historical knowledge on why the error is being returned the way it is  - my assumption is that it eases debugging, but again, we should be able to post and get that from server logs.

wolfy1339:
That has been there since the very beginnings of this project.

It would make sense to remove it and log it instead

nickfloyd:
> It would make sense to remove it and log it instead

Solid.  I'll take care of it in this PR.  Thanks for the 👀!

.github/workflows/codeql-analysis.yml
G-Rath: Is there actually a difference in this? My assumption was that `javascript` did cover TypeScript even though it is listed in the docs as a different language.

I'm now wondering if a bunch of my repos might have been getting only half-checked because of this 😅

nickfloyd: Yeah after reading through the docs the best strategy seemed to be to use both for any repo where the source had both TypeScript and JavaScript i.e.:

`language: [ 'javascript', 'typescript' ]` < this seems to be working correctly for mixed source.

wolfy1339 (approved):

.github/workflows/codeql-analysis.yml
G-Rath: That sounds like something that the default codeql workflow generator should include (at the least, mention `typescript` as one of the options in it's comment) - can I assume from this you'll pass this on to the right folks at GitHub to action that, or should I open a ticket or issue somewhere?

PR #704, opened 2022-07-07
ci: stop testing against NodeJS v10, v12

wolfy1339:
BREAKING CHANGE: Drop support for NodeJS v10, v12

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,9 @@
         "@pika/plugin-build-node",
         {
           "minNodeVersion": "14"
+        },
+        {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -59,6 +59,9 @@
       ],
       [
         "@pika/plugin-build-node",
+        {
+          "minNodeVersion": "14"
+        },
```
oscard0m, line 64: This is wrong. Looks like a bug in Octoherd script

wolfy1339: yes, it is

PR #740, opened 2022-09-11
Removes the `onUnhandledRequest` middleware option

baoshan:
__expected to be merged into `beta` branch__

__What?__

This PR removes the `onUnhandledRequest` middleware option.

__Why?__

To respect a proposed general rule that all Octokit middlewares (`oauth-app.js`, `webhooks.js`, `app.js`):

- do NOT handle requests outside `pathPrefix` (or `path`) option but provide a natural way for concrete runtimes/frameworks to handle unhandled requests
- always handle requests within `pathPrefix` (or `path`) option (returns `404` for unknown routes)

__But my code relies on the `onUnhandledRequest` option!__

Because `webhook.js` only exposes a (shared) middleware for Node.js and Express.js middleware currently, the natural way (to handle unhandled requests) is (AFAICT):

- For native Node.js (`http.createServer`), check returned boolean value:

```js
const middleware = createNodeMiddleware(webhooks, {});
createServer(async (req, res) => {
  if (await middleware(req, res)) return;
  res.writeHead(404);
  res.end();
}).listen();
```

- For Express.js, rely on proper handler(s) registered at your desired path(s).

-----
[View rendered README.md](https://github.com/baoshan/octokit-webhooks.js/blob/remove_on-unhandled-request/README.md)

### DIFF
--- a/src/middleware/node/types.ts
+++ b/src/middleware/node/types.ts
@@ -1,9 +1,3 @@
-// remove type imports from http for Deno compatibility
-// see https://github.com/octokit/octokit.js/issues/24#issuecomment-817361886
-// import { IncomingMessage, ServerResponse } from "http";
-type IncomingMessage = any;
-type ServerResponse = any;
-
 import { Logger } from "../../createLogger";
 
 export type MiddlewareOptions = {
--- a/src/middleware/node/types.ts
+++ b/src/middleware/node/types.ts
@@ -9,8 +9,4 @@ import { Logger } from "../../createLogger";
 export type MiddlewareOptions = {
   path?: string;
   log?: Logger;
-  onUnhandledRequest?: (
-    request: IncomingMessage,
-    response: ServerResponse
-  ) => void;
 };
--- a/README.md
+++ b/README.md
@@ -599,32 +599,6 @@ Used for internal logging. Defaults to [`console`](https://developer.mozilla.org
 
 </td>
     </tr>
-    <tr>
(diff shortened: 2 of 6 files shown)


wolfy1339 (approved):

wolfy1339:
Test failures are unrelated to this PR.

They should have been caught in #732

gr2m:
> Because `webhook.js` only exposes a (shared) middleware for Node.js and Express.js middleware currently, the natural way (to handle unhandled requests) is (AFAICT):
> 
>     * For native Node.js (`http.createServer`), check returned boolean value:
> 
> 
> ```js
> const middleware = createNodeMiddleware(webhooks, {});
> createServer(async (req, res) => {
>   if (await middleware(req, res)) return;
>   res.writeHead(404);
>   res.end();
> }).listen();
> ```

Should we add that to the README?

baoshan:
Thanks! I’ve fixed that.

src/middleware/node/types.ts
```diff
@@ -9,8 +9,4 @@ import { Logger } from "../../createLogger";
 export type MiddlewareOptions = {
   path?: string;
   log?: Logger;
-  onUnhandledRequest?: (
```
wolfy1339, line 12: Can you remove the type declarations, and the associated comments that are now unused

baoshan: Thanks. Done.

PR #741, opened 2022-09-11
tests: fix jest errors

wolfy1339:
Fixes test failures introduced in #732


wolfy1339:
@octokit/js-community We will need to double check all the `octokit` repos to make sure that the inline snapshots are corrected.

G-Rath (approved):

PR #777, opened 2022-12-06
Accept array of secrets, not just one

### LINKED ISSUE #770: [FEAT]: Facilitate webhook secret rotation
### Describe the need

1. Generate a new secret.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

nwf-msr:
Resolves #770.

----

### Before the change?

* Only one secret could be specified for a WebHook object.

### After the change?

* A bare secret or an array of secrets are each accepted.  In the latter case, any match will be accepted.

----

### Pull request checklist
- [ ] Added the appropriate label for the given change

- [x] No

### Pull request type

I do not seem to be able to add labels, but please add `Type: Feature` for me?

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ require("http").createServer(createNodeMiddleware(webhooks)).listen(3000);
 To ease key rotation, one can also feed the `Webhook` constructor an array of
 secrets. Generate a new secret and update your Webhook receiver to accept both
 this new one and the current secret, then update GitHub to send the new secret,
-then update your Webhook receiver to accept only the new secret.  Signatures will
+then update your Webhook receiver to accept only the new secret. Signatures will
 always be made with the first secret.
 
 ```js
--- a/README.md
+++ b/README.md
@@ -56,6 +56,17 @@ require("http").createServer(createNodeMiddleware(webhooks)).listen(3000);
 // can now receive webhook events at /api/github/webhooks
 ```
 
+To ease key rotation, one can also feed the `Webhook` constructor an array of
+secrets. Generate a new secret and update your Webhook receiver to accept
+both this new one and the current secret, then update GitHub to send the new
+secret, then update your Webhook receiver to accept only the new secret.
+
+```js
+const webhooks = new Webhooks({
+  secret: ["mynewsecret", "myoldsecret"],
+});
+```
+
 ## Local development
 
 You can receive webhooks on your local machine or even browser using [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) and [smee.io](https://smee.io/).
@@ -111,11 +122,13 @@ new Webhooks({ secret /*, transform */ });
         <code>
           secret
         </code>
-        <em>(String)</em>
+        <em>(String or String Array)</em>
       </td>
       <td>
         <strong>Required.</strong>
-        Secret as configured in GitHub Settings.
+        Secret as configured in GitHub Settings.  If an array of strings, an
+        incoming webhook event will be accepted if its signature verifies
+        with any of the secrets given.
       </td>
     </tr>
     <tr>
--- a/README.md
+++ b/README.md
@@ -57,9 +57,10 @@ require("http").createServer(createNodeMiddleware(webhooks)).listen(3000);
 ```
 
 To ease key rotation, one can also feed the `Webhook` constructor an array of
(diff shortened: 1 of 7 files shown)


gr2m (changes requested):
PR looks great so far! Can you please also update the `### Constructor` options table?

Also can you run `npm run lint:fix` to make the prettier linter happy?

gr2m:
can you look into the failing tests?

Please don't force-push changes, we don't mind a messy commit history in pull requests, we squash & merge them in the end anyway

nwf-msr:
What'd I do to make CodeQL unhappy? :(

wolfy1339:
It's not you, it's CodeQL being uncooperative.

Just ignore it

README.md
```diff
@@ -56,6 +56,18 @@ require("http").createServer(createNodeMiddleware(webhooks)).listen(3000);
 // can now receive webhook events at /api/github/webhooks
 ```
 
+To ease key rotation, one can also feed the `Webhook` constructor an array of
+secrets. Generate a new secret and update your Webhook receiver to accept both
+this new one and the current secret, then update GitHub to send the new secret,
+then update your Webhook receiver to accept only the new secret. Signatures will
+always be made with the first secret.
```
gr2m, line 63:
```suggestion
then update your Webhook receiver to accept only the new secret. New signatures will
be calculated using the first secret.
```

README.md
```diff
@@ -111,11 +123,14 @@ new Webhooks({ secret /*, transform */ });
         <code>
           secret
         </code>
-        <em>(String)</em>
+        <em>(String or String Array)</em>
```
gr2m, line 126:
```suggestion
        <em>(String or Array of Strings)</em>
```

README.md
```diff
@@ -111,11 +123,14 @@ new Webhooks({ secret /*, transform */ });
...
       </td>
       <td>
         <strong>Required.</strong>
-        Secret as configured in GitHub Settings.
+        Secret as configured in GitHub Settings.  If an array of strings, an
+        incoming webhook event will be accepted if its signature verifies
+        with any of the secrets given and signatures will always be generated
+        with the first secret.
```
gr2m, line 133:
```suggestion
        Secret as configured in GitHub Settings.  If set to an array of strings, a
        webhook request will be accepted if its signature verifies
        with any of the given secrets. New signatures will always be
        calculated using the first secret.
```

gr2m (commented):
I think the pull request looks good as is.

Good point about the `.sign()` method. Given that the first secret in a given array of secrets has special meaning, I wonder if it would make more sense to keep the secret as is, but add an additional option such as `alternativeSecrets` or maybe some other name that better conveys that these secrets are used for incoming request verification only? 

My understanding is that in your case, there will always only be one other secret. Do you see a use case where more than two secrets would be needed to be provided?

nwf-msr:
> I think the pull request looks good as is.
> 
> Good point about the `.sign()` method. Given that the first secret in a given array of secrets has special meaning, I wonder if it would make more sense to keep the secret as is, but add an additional option such as `alternativeSecrets` or maybe some other name that better conveys that these secrets are used for incoming request verification only?

That makes sense.  I'll take a stab at that approach.

> My understanding is that in your case, there will always only be one other secret. Do you see a use case where more than two secrets would be needed to be provided?

That's true and no, I don't see reason to need more than one active and one retiring secret, at least, at the moment.  I'm tempted to leave `alternativeSecrets` as an array of strings, tho', just in case.  It's not like it's hard or on the hot path most of the time.

gr2m:
maybe call them `alternativeSecretsForVerification` to make clear they won't be used for the `.sign()` method?

nwf-msr:
Belatedly, I've raised https://github.com/octokit/webhooks.js/pull/811 .

wolfy1339:
Instead of opening a separate PR, you could have simply reused this one.

Closing in favour of #811

nwf-msr:
I didn't reuse this one because I was advised not to force-push changes and enough time had elapsed that doing the work on a new checkout seemed wise.

wolfy1339:
When you need to update using a rebase, a force-push can be warranted.
When you simply always squash your changes, it can become harder to review changes since the last time it was reviewed, and we don't mind having many commits.

PR #780, opened 2022-12-07
ci(codeql): set env variable `CODEQL_ACTION_RUN_MODE`

wolfy1339:
The CodeQL tests are failing, and this was a solution in another of the repos
ref: https://github.com/octokit/fixtures/commit/8642c27ddee746ec9a00a97ac28bdaf823046b2b

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -30,7 +30,7 @@ jobs:
           languages: ${{ matrix.language }}
       - name: Autobuild
         env:
-          CODEQL_ACTION_RUN_MODE: action
+          CODEQL_ACTION_RUN_MODE: Action
         uses: >-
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -29,6 +29,8 @@ jobs:
         with:
           languages: ${{ matrix.language }}
       - name: Autobuild
+        env:
+          CODEQL_ACTION_RUN_MODE: action
```
kfcampbell:
```suggestion
          CODEQL_ACTION_RUN_MODE: Action
```

I'm seeing some capitalized usages of this (like [here](https://github.com/github/codeql-action/issues/912)); it can't hurt to try it that way.

kfcampbell:
I can't find the "action" vs. "Action" thing documented anywhere unfortunately, but the capitalized version seems to be correct.

kfcampbell (approved):

Issue #775, opened 2022-11-29
[BUG]: dependabot_alert events fail to verify


timomeh:
## Observed Behavior

I noticed that `webhooks.verify()` returns false for dependabot_alert events, but only for this event type. All other events verify successfully.

To reproduce, I created a small script:

```ts
import { Webhooks } from '@octokit/webhooks'

const webhooks = new Webhooks({ secret: 'redacted' })

const check_run_event = {
  sign: "sha256=redacted", // from the X-Hub-Signature-256 header of the check_run event
  data: { /* copied a check_run event payload from the App Settings */ }
}
const dependabot_alert_event = {
  sign: "sha256=redacted", // from the X-Hub-Signature-256 header of the dependabot_alert event
  data: { /* copied a dependabot_alert event payload from the App Settings */ }
}

await webhooks.verify(dependabot_alert_event.data, dependabot_alert_event.sign)
// -> false

await webhooks.verify(check_run_event.data, check_run_event.sign)
// -> true
```

## Expected Behavior

Given that other events verify successfully, I would've expected that the dependabot_alert event also verifies successfully.

## Note

I'm unsure if this is a bug of @octokit/webhooks, or maybe even a miscalculation of the `X-Hub-Signature-256` header coming from GitHub-Hookshot. In case it really seems to be an issue of Github-Hookshot, where can I report this?

I'm aware that the dependabot_alert event is still in beta.

### Versions

@octokit/webhooks v10.3.1

### Relevant log output

```shell
No logs are printed
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

wolfy1339:
This isn't the right repository for issues with the `@octokit/webhooks` npm package. The proper repo would be https://github.com/octokit/webhooks.js

@gr2m @timrogers Can this issue be transferred to the proper repo.

***

That is odd. There's nothing printed to the console at all?
Would you mind sharing the payload you received that fails the validation?

timomeh:
@wolfy1339 Ah, sorry for the wrong repo – too many tabs open simultaneously.

Looking at the source of the `@octokit/webhooks` npm package, the `verify` call doesn't include any debug logging. The verfication itself is in the `@octokit/webhooks-methods` package, which also doesn't seem to include any debug logging.

Here's the output of the failing payload. It's a private repo, so I redacted some sensitive info.

```json
{
  "action": "created",
  "alert": {
    "number": 45,
    "state": "open",
    "dependency": {
      "package": {
[...]

PR #790, opened 2023-01-04
feat: deprecate `onUnhandledRequest` and accepting object payloads for  for `webhooks.verify()` and `webhooks.verifyAndReceive()`

### LINKED ISSUE #589: Only accept string payloads for `webhooks.verify()` and `webhooks.verifyAndReceive()`
**What’s missing?**

**Alternatives you tried**

n/a

### LINKED ISSUE #784: Deprecate `onUnhandledRequest` middleware option
Followup to #740, we should deprecate the `onUnhandledRequest` middleware option, and release that PR that is currently sitting in the `beta` branch

wolfy1339:
Resolves #784
Resolves #589
See also #775

----

### Before the change?

* The `onUnhandledRequest` option was not deprecated
* One could pass an object to the API and due to how Node works, it could change the object slightly so it wouldn't validate

### After the change?

* Deprecate the `onUnhandledRequest` option
* Users will be required to pass a string version of the JSON object

### Other information

* 

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----

### DIFF
--- a/src/middleware/node/index.ts
+++ b/src/middleware/node/index.ts
@@ -12,9 +12,15 @@ export function createNodeMiddleware(
     log = createLogger(),
   }: MiddlewareOptions = {}
 ) {
+  const deprecateOnUnhandledRequest = (request: any, response: any) => {
+    console.error(
+      "[@octokit/webhooks] `onUnhandledRequest()` is deprecated and will be removed in a future release of `@octokit/webhooks`"
+    );
+    return onUnhandledRequest(request, response);
+  };
   return middleware.bind(null, webhooks, {
     path,
-    onUnhandledRequest,
+    onUnhandledRequest: deprecateOnUnhandledRequest,
     log,
   } as Required<MiddlewareOptions>);
 }
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ webhooks.verify(eventPayload, signature);
           eventPayload
         </code>
         <em>
-          (Object or String)
+          (Object [deprecated] or String)
         </em>
       </td>
       <td>
@@ -262,7 +262,7 @@ webhooks.verifyAndReceive({ id, name, payload, signature });
           payload
         </code>
         <em>
-          Object or String
+          Object (deprecated) or String
         </em>
       </td>
       <td>
@@ -601,7 +601,7 @@ Used for internal logging. Defaults to [`console`](https://developer.mozilla.org
     </tr>
     <tr>
       <td>
-        <code>onUnhandledRequest</code>
+        <code>onUnhandledRequest (deprecated)</code>
         <em>
           function
         </em>
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,12 +26,8 @@ export type Iverify = {
 };
 export type IverifyAndReceive = {
   /** @deprecated Passing a JSON payload object to `verifyAndReceive()` is deprecated and the functionality will be removed in a future release of `@octokit/webhooks`. */
-  (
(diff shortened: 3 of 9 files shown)


wolfy1339:
The deprecations still need to be added to the middleware.
That will be another PR, as it requires modifying the middleware to accept string payloads

nickfloyd (approved):
LGTM! ❤️ Thanks for getting these issues knocked out!

src/middleware/node/index.ts
```diff
@@ -12,9 +12,15 @@ export function createNodeMiddleware(
     log = createLogger(),
   }: MiddlewareOptions = {}
 ) {
+  const deprecateOnUnhandledRequest = (request: any, response: any) => {
+    console.error(
```
AaronDewes, line 16: I know this has been merged, but shouldn't this be a warning instead of an error? It being an error seems to confuse Jest in some tests I run, and a warning would fit here better in my opinion.

austenstone:
Should this [example ](https://github.com/octokit/webhooks.js#local-development) be updated? How do we do the equivalent of this with just a string?

wolfy1339:
Simply don't pass the string received into `JSON.parse()` and the example should work

austenstone:
I see payload can't be an object.
```ts
    this.source = new EventSource(webhookProxyUrl);
    this.source.onmessage = (event) => {
      const webhookEvent = JSON.parse(event.data);
      this.app.webhooks
        .verifyAndReceive({
          id: webhookEvent["x-request-id"],
          name: webhookEvent["x-github-event"],
          payload: JSON.stringify(webhookEvent.body), // <= HERE
          signature: webhookEvent["x-hub-signature"]
        })
        .catch(console.error);
    };
```

wolfy1339:
What version of `@octokit/webhooks` are you using?

What is the exact message you are getting and from where?

I don't see why it would error

austenstone:
> What is the exact message you are getting and from where?

I don't see why it would error
> [@octokit/webhooks] Passing a JSON payload object to `verifyAndReceive()` is deprecated and the functionality will be removed in a future release of `@octokit/webhooks`

When I pass object instead of string the payload in `verifyAndReceive` param.

`"octokit": "^2.0.15"`

`App` from `import { App, createNodeMiddleware } from "octokit";`

wolfy1339:
I cannot reproduce this issue.

The following code works for me

```js
const { App } = require('octokit'); // Octokit.js v2.0.15 and v2.0.16
const EventSource = require('eventsource');

const webhookProxyUrl = "https://smee.io/IrqK0nopGAOc847";
const source = new EventSource(webhookProxyUrl);

const app = new App({
  appId: 95608,
  privateKey: readFileSync('./private-key.pem').toString(),
  webhooks: {
    secret: 'thisismysecret'
  }
});

source.onmessage = (event) => {
  const webhookEvent = JSON.parse(event.data);
  app.webhooks
    .verifyAndReceive({
      id: webhookEvent["x-request-id"],
      name: webhookEvent["x-github-event"],
      signature: webhookEvent["x-hub-signature"],
      payload: JSON.stringify(webhookEvent.body),
    })
    .catch(console.error);
};
```

austenstone:
```payload: JSON.stringify(webhookEvent.body)```

If this isn't a string the warning will appear.

Ex:
```payload: webhookEvent.body```

wolfy1339:
That is the intended behaviour

PR #792, opened 2023-01-08
feat: v11

### LINKED ISSUE #589: Only accept string payloads for `webhooks.verify()` and `webhooks.verifyAndReceive()`

### LINKED ISSUE #775: [BUG]: dependabot_alert events fail to verify

### LINKED ISSUE #828: Modify `package.json` fields for ESM and browser builds best practices
> Should we change the current value of the `module` field to point to `dist-src/` and add a new `browser` field that points to `dist-web/` ?
> 
> This would alleviate issues like [octokit/auth-app.js#59](https://github.com/octokit/auth-app.js/issues/59) until work on `octokit-next` is finished
> 
> ```diff
> diff --git a/scripts/build.mjs b/scripts/build.mjs
> index 3ef8688..73d9ef9 100644
> --- a/scripts/build.mjs
> +++ b/scripts/build.mjs
> @@ -72,9 +71,9 @@ async function main() {
>          ...pkg,
>          files: ["dist-*/**", "bin/**"],
>          main: "dist-node/index.js",
> -        module: "dist-web/index.js",
> +        browser: "dist-web/index.js",
>          types: "dist-src/index.d.ts",
> -        source: "dist-src/index.js",
> +        module: "dist-src/index.js",
>        },
> ```

I would do that in a separate PR (feel free to create a new issue mentioning this comment) so we do not add more things on top of this migration. I think it's good the differences are minimum in this PR. What do you think?

_Originally posted by @oscard0m in https://github.com/octokit/webhooks.js/pull/819#issuecomment-1498924723_

wolfy1339:
Fixes #589, fixes #775, fixes #828

### Breaking changes
* middleware only and (always) handles requests at `path`
* Only accept string payloads for `webhooks.verify()` and `webhooks.verifyAndReceive()` and `webhooks.sign()`
* The middleware now expects only string payloads for `request.body`
* Modifies the `module` field in `package.json` to point to the source ESM, and adds a `browser` field that points to the browser bundle


src/middleware/node/get-payload.ts
```diff
@@ -12,20 +11,11 @@ import AggregateError from "aggregate-error";
...
 
-  if (request.body) {
-    if (typeof request.body !== "string") {
-      console.error(
-        "[@octokit/webhooks] Passing the payload as a JSON object in `request.body` is deprecated and will be removed in a future release of `@octokit/webhooks`, please pass it as a a `string` instead."
-      );
-    }
-    return Promise.resolve(request.body as WebhookEvent | string);
-  }
+  if (request.body) return Promise.resolve(request.body);
```
philipandersson, line 18: Finally no more objects 🙏

src/verify-and-receive.ts
```diff
@@ -35,9 +29,6 @@ export async function verifyAndReceive(
   return state.eventHandler.receive({
     id: event.id,
     name: event.name,
-    payload:
-      typeof event.payload === "string"
-        ? JSON.parse(event.payload)
-        : event.payload,
+    payload: JSON.parse(event.payload),
```
philipandersson, line 32: Great that we are addressing: https://github.com/octokit/webhooks.js/issues/589#issuecomment-1376814798
I guess you could argue that it should be the `receive` function's responsibility to parse the `JSON`.

oscard0m (dismissed):

wolfy1339:
@octokit/js @octokit/extensibility-sdks Is there anything else that you could think of that should be included in this major version

oscard0m:
> @octokit/js @octokit/extensibility-sdks Is there anything else that you could think of that should be included in this major version

As you mentioned, I agree on adding this as well: https://github.com/octokit/webhooks.js/issues/828

wolfy1339:
Once we land #819, I can work on addressing #828 in this PR

wolfy1339:
I addressed #828 in ffbf0ac70ae8fb901029a169b1c677e6a55bef7d

kfcampbell (approved):
I'm :+1: on this, thank you @wolfy1339!

PR #793, opened 2023-01-08
feat: only accept string payloads for `webhooks.verify()` and `webhooks.verifyAndReceive()`

### LINKED ISSUE #589: Only accept string payloads for `webhooks.verify()` and `webhooks.verifyAndReceive()`

wolfy1339:
Resolves #589

**NOTE**: This is getting merged into the `beta` branch

----

### Before the change?

* The ability to pass webhook payloads was deprecated

### After the change?

* The deprecated code has been removed, and now users no longer have the ability to pass object payloads to `webhooks.verify()` and `webhooks.verifyAndReceive()`

### Other information

* 

----

### Pull request checklist

- [x] Yes (Please add the `Type: Breaking change` label)
- [ ] No

If `Yes`, what's the impact:  

* Users can no longer pass object payloads to the `webhooks.verify()` and `webhooks.verifyAndReceive()` methods

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


test/unit/deprecation.test.ts
```diff
@@ -1,45 +1,3 @@
...
-        { secret, algorithm: "sha256" },
-        pushEventPayloadString
-      ),
-    });
-    expect(spy).toBeCalledWith(
-      "[@octokit/webhooks] Passing a JSON payload object to `verifyAndReceive()` is deprecated and the functionality will be removed in a future release of `@octokit/webhooks`"
-    );
-    spy.mockClear();
-  });
+  test("there are currently no deprecations", () => {});
```
G-Rath: Ehh empty tests are generally considered a mistake in jest - I think it'd be better to just delete this file

G-Rath (approved):

test/unit/deprecation.test.ts
wolfy1339: I don't mind, this is just how the file was when we last did a major release removing deprecated methods

G-Rath:
Fwiw I've just double checked an Azure function project and AWS lambda project which are used by GitHub webhooks, to confirm that getting the raw body is very straightforward 👍

PR #794, opened 2023-01-08
feat: only accept string payloads for `webhooks.sign()`

wolfy1339:
### Before the change?

* One could pass a JSON object or a string to the function, and have it work normally

### After the change?

* One can no longer pass a JSON object, and must pass it as a string instead

### Other information

* 

----

### Pull request checklist

- [x] Yes (Please add the `Type: Breaking change` label)
- [ ] No

If `Yes`, what's the impact:  

* Passing the payload as a JSON object is no longer supported

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


G-Rath:
Is the only motivation here to mirror the fact that `verify` now only takes a string? as if so then I'm less sure about this change because while the symmetry is nice, unlike `verify` we know exactly what to do if we're passed an object to `sign` (call `JSON.stringify`) and one of the expected use-cases for this is testing where there could be a bunch of uses per file so handling this in the library  rather than in userland could be nice?

It seems to me like we could instead just replace `toNormalizedJsonString` with a standard `JSON.stringify` call instead - what do other people think?

wolfy1339:
> Is the only motivation here to mirror the fact that `verify` now only takes a string? as if so then I'm less sure about this change because while the symmetry is nice, unlike `verify` we know exactly what to do if we're passed an object to `sign` (call `JSON.stringify`) and one of the expected use-cases for this is testing where there could be a bunch of uses per file so handling this in the library rather than in userland could be nice?

The function uses the method exported from `@octokit/webhooks-methods`, which itself only accepts a string payload. Instead of wrapping it, we can simply point to it directly.

The reasoning for only accepting string payloads for all methods is explained here: https://github.com/octokit/webhooks-methods.js/pull/21#issuecomment-863497657

> It seems to me like we could instead just replace `toNormalizedJsonString` with a standard `JSON.stringify` call instead - what do other people think?

This is the original PR that added the `toNormalizedJsonString`:
https://github.com/octokit/webhooks.js/pull/72

G-Rath:
hmm ok - I think I'm happy to go with this for now, especially since I've not actively worked with GH webhooks in a while, but I guess mark me down as open to reconsidering this in future because I think there could be a valid argument for this behaviour in `sign` but not in `verify`.

G-Rath (approved):

sweets2222 (approved):

gr2m (approved):

sweets2222 (approved):

PR #795, opened 2023-01-08
fix: return more helpful error when `Content-Type` header is not `application/json`

### LINKED ISSUE #158: throw a helpful error if received request has content type set to `application/x-www-form-url-encoded`
I ran into [this issue][1] and though it was never resolved, it appears tha the underlying problem is that `@octokit/wehooks` is [assuming that the incoming request body is `application/json`][2]. However, github can also deliver payloads in `application/x-www-form-url-encoded` which is, in fact, the default delivery format when creating a webhook via the github UI

image

In the event that the payload is encoded as `application/x-www-form-urlencoded`, the middleware will try to parse it as JSON.

I could see two possible solutions:

1. asserting that the `content-type` header is a`application/json`, and rejecting with a 400 and a status text that clearly indicates that `x-form-url-encoded` is not acceptable, and that the webhook configuration should be updated.
2. checking the content-type and decoding the payload appropriately for both `application/json` and `application/x-form-www-url-encoded`.

Would be happy to help implement either.

[1]: https://github.com/octokit/webhooks.js/issues/75
[2]: https://github.com/octokit/webhooks.js/blob/f6552815c887b3651a57e92d9e9b7821dc90ff91/src/middleware/get-payload.js#L18

wolfy1339:
Resolves #158

----

### Before the change?

* The middleware would fail since it would assume that the `Content-Type` header is set to `application/json`, and it would return an error

### After the change?

* The middleware will return an [HTTP 415](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415) response explaining that only `application/json` is supported 

### Other information

* 

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----

### DIFF
--- a/src/middleware/node/middleware.ts
+++ b/src/middleware/node/middleware.ts
@@ -44,7 +44,10 @@ export async function middleware(
     }
   }
 
-  if (request.headers["content-type"] !== "application/json") {
+  // Check if the Content-Type header is `application/json` and allow for charset to be specified in it
+  // Otherwise, return a 415 Unsupported Media Type error
+  // See https://github.com/octokit/webhooks.js/issues/158
+  if (!request.headers["content-type"].startsWith("application/json")) {
     response.writeHead(415, {
       "content-type": "application/json",
       accept: "application/json",
--- a/src/middleware/node/middleware.ts
+++ b/src/middleware/node/middleware.ts
@@ -1,5 +1,5 @@
 // remove type imports from http for Deno compatibility
-// see https://github.com/octokit/octokit.js/issues/24#issuecomment-817361886
+// see https://github.com/octokit/octokit.js/issues/2075#issuecomment-817361886
 // import { IncomingMessage, ServerResponse } from "http";
 type IncomingMessage = any;
 type ServerResponse = any;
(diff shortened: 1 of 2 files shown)


src/middleware/node/middleware.ts
```diff
@@ -44,6 +44,19 @@ export async function middleware(
     }
   }
 
+  if (request.headers["content-type"] !== "application/json") {
```
gr2m: Maybe check with `.startsWith()`? Just in case GitHub decides to add character encoding to the header (`Content-type: application/json; charset=utf-8`)

wolfy1339: I hadn't thought of that, good idea!

src/middleware/node/middleware.ts
```diff
@@ -1,5 +1,5 @@
 // remove type imports from http for Deno compatibility
-// see https://github.com/octokit/octokit.js/issues/24#issuecomment-817361886
+// see https://github.com/octokit/octokit.js/issues/2075#issuecomment-817361886
```
gr2m, line 2: wow, nice find!

gr2m (approved):

PR #811, opened 2023-02-06
Ease key rotation via an array of fallback secrets

### LINKED ISSUE #770: [FEAT]: Facilitate webhook secret rotation

nwf-msr:
The core change is in src/verify.ts.  We don't really have to do anything special here for constant-timedness.  Since each comparison is constant time, the only timing differential is between all checks (completely) failing and some check (completely) succeeding, and so the only thing learned is whether a guess was completely right or wrong.

The rest of the changes here are plumbing this new mechanism up to the top level and testing.

As suggested by @gr2m on https://github.com/octokit/webhooks.js/pull/777

Resolves #770.

----

### Before the change?

* Only one secret could be specified for a WebHook object.

### After the change?

* Additional secrets are accepted for verification

----

### Pull request checklist
- [ ] Added the appropriate label for the given change

- [x] No

### Pull request type

I do not seem to be able to add labels, but please add `Type: Feature` for me?

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,17 @@ require("http").createServer(createNodeMiddleware(webhooks)).listen(3000);
 // can now receive webhook events at /api/github/webhooks
 ```
 
+To ease key rotation, the `Webhooks` constructor also accepts an array of additional secrets used as fallbacks for verification.
+That is, if a message fails to verify with the provided `secret`, the receiver will also test against each of these additional secrets.
+The message will be accepted if it verifies against any secret.
+
+```js
+const webhooks = new Webhooks({
+  secret: "mysecret",
+  alternativeSecretsForVerification: ["oldsecret1", "oldsecret2"],
+});
+```
+
 ## Local development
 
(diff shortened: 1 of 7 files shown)


wolfy1339:
Would it be better to implement the behaviour in `@octokit/webhooks-methods` instead?

In the next major release in the works #792, there is no longer any custom verify methods in the codebase.

nwf-msr:
Could do if that's the better thing to do. Who else sits downstream of `@octokit/webhooks-methods`?

wolfy1339:
> Could do if that's the better thing to do. Who else sits downstream of `@octokit/webhooks-methods`?

I don't think there are any of our packages that are downstream except for this one

wolfy1339:
Closing per author comment, https://github.com/octokit/webhooks.js/issues/770#issuecomment-1724404682

PR #818, opened 2023-03-06
chore: explicitly mark imports/exports that are only used as types

wolfy1339:
----

### Before the change?

* Some compilers that don't fully parse typescript, ex: Babel and esbuild, would include these imports/exports in the build output

### After the change?

* With the explicit markers, those imports/exports aren't included in the build output

### Other information

*

----

### Pull request checklist

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/main/community/breaking_changes.md) to help!

- [x] No

If `Yes`, what's the impact:

* N/A

### Pull request type

- Bugfix: `Type: Bug`

----

### DIFF
--- a/src/event-handler/on.ts
+++ b/src/event-handler/on.ts
@@ -1,5 +1,5 @@
 import { emitterEventNames } from "../generated/webhook-names";
-import {
+import type {
   EmitterWebhookEvent,
   EmitterWebhookEventName,
   State,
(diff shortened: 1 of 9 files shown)


G-Rath (approved):

oscard0m:
Should we lint this? https://typescript-eslint.io/rules/consistent-type-imports/ @wolfy1339

wolfy1339:
It would be nice, but i don't think introducing another linter is the best idea

oscard0m:
> It would be nice, but i don't think introducing another linter is the best idea

Some extra step in test.yml maybe? Still creates friction to contributors, right?

wolfy1339:
If it were a plugin for prettier, it would work just fine. Introducing eslint into the mix is probably not the best call.

For now, it will just have to be something to watch out for

PR #819, opened 2023-03-06
build: replace pika with esbuild and tsc

wolfy1339:
Pika has been deprecated for a while now, and the project has now been archived.

Uses esbuild to transpile the TS source code into an ESM source, NodeJs bundle, and a browser bundle

Uses the TypeScript compiler to generate the types

----

### Before the change?

* Uses pika for the build system

### After the change?

* Uses esbuild to generate the JS bundles
* Uses the TypeScript compiler to generate the type definitions

### Other information

*

----

### Pull request checklist

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/main/community/breaking_changes.md) to help!

- [x] No

If `Yes`, what's the impact:

* N/A

### Pull request type

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "license": "MIT",
   "author": "Gregor Martynus (https://twitter.com/gr2m)",
   "scripts": {
-    "build": "pika-pack build",
+    "build": "ts-node --transpile-only scripts/build.ts && tsc -p tsconfig.json",
     "coverage": "jest --coverage && open coverage/lcov-report/index.html",
     "generate-types": "ts-node --transpile-only scripts/generate-types.ts",
     "lint": "prettier --check 'src/**/*.{ts,json}' 'scripts/**/*' 'test/**/*.ts' README.md package.json",
@@ -49,17 +49,15 @@
     "@jest/types": "^29.0.0",
     "@octokit/tsconfig": "^1.0.1",
     "@octokit/webhooks-schemas": "6.11.0",
-    "@pika/pack": "^0.3.7",
-    "@pika/plugin-build-node": "^0.9.2",
-    "@pika/plugin-build-web": "^0.9.2",
-    "@pika/plugin-ts-standard-pkg": "^0.9.2",
     "@types/jest": "^29.0.0",
     "@types/json-schema": "^7.0.7",
     "@types/node": "^18.0.0",
     "@types/node-fetch": "^2.5.8",
     "@types/prettier": "^2.0.0",
     "axios": "^1.0.0",
+    "esbuild": "^0.17.11",
     "express": "^4.17.1",
+    "glob": "^9.2.1",
     "jest": "^29.0.0",
     "node-fetch": "^2.6.7",
     "prettier": "2.8.7",
@@ -73,21 +71,5 @@
   },
   "publishConfig": {
     "access": "public"
-  },
-  "@pika/pack": {
-    "pipeline": [
-      [
-        "@pika/plugin-ts-standard-pkg"
-      ],
-      [
-        "@pika/plugin-build-node",
-        {
-          "minNodeVersion": "14"
-        }
-      ],
-      [
-        "@pika/plugin-build-web"
-      ]
-    ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "license": "MIT",
   "author": "Gregor Martynus (https://twitter.com/gr2m)",
   "scripts": {
-    "build": "node scripts/build.ts && tsc -p tsconfig.json",
+    "build": "node scripts/build.mjs && tsc -p tsconfig.json",
     "coverage": "jest --coverage && open coverage/lcov-report/index.html",
     "generate-types": "ts-node --transpile-only scripts/generate-types.ts",
     "lint": "prettier --check 'src/**/*.{ts,json}' 'scripts/**/*' 'test/**/*.ts' README.md package.json",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "license": "MIT",
   "author": "Gregor Martynus (https://twitter.com/gr2m)",
   "scripts": {
-    "build": "ts-node --transpile-only scripts/build.ts && tsc -p tsconfig.json",
+    "build": "node scripts/build.ts && tsc -p tsconfig.json",
     "coverage": "jest --coverage && open coverage/lcov-report/index.html",
     "generate-types": "ts-node --transpile-only scripts/generate-types.ts",
     "lint": "prettier --check 'src/**/*.{ts,json}' 'scripts/**/*' 'test/**/*.ts' README.md package.json",
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -75,6 +75,7 @@ async function main() {
         module: "dist-web/index.js",
         types: "dist-src/index.d.ts",
         source: "dist-src/index.js",
+        sideEffects: false
       },
       null,
       2
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -71,10 +71,10 @@ async function main() {
       {
         ...pkg,
         files: ["dist-*/**", "bin/**"],
-        main: "./dist-node/index.js",
-        module: "./dist-web/index.js",
-        types: "./dist-src/index.d.ts",
-        source: "./dist-src/index.js",
+        main: "dist-node/index.js",
+        module: "dist-web/index.js",
+        types: "dist-src/index.d.ts",
+        source: "dist-src/index.js",
       },
       null,
       2
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -21,7 +21,7 @@ async function main() {
     platform: "neutral",
     format: "esm",
     ...sharedOptions,
-    sourcemap: false
+    sourcemap: false,
   });
 
   // Remove the types file from the dist-src folder
@@ -77,7 +77,7 @@ async function main() {
         module: "dist-web/index.js",
         types: "dist-src/index.d.ts",
         source: "dist-src/index.js",
-        sideEffects: false
+        sideEffects: false,
       },
       null,
       2
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -11,7 +11,7 @@ const sharedOptions = {
 
 async function main() {
   // Start with a clean slate
-  await rm("pkg", { recursive: true });
+  await rm("pkg", { recursive: true, force: true });
   // Build the source code for a neutral platform as ESM
   await esbuild.build({
     entryPoints: await glob(["./src/*.ts", "./src/**/*.ts"]),
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -4,7 +4,7 @@ import { glob } from "glob";
 
 const sharedOptions = {
   sourcemap: "external",
-  sourcesContent: false,
+  sourcesContent: true,
   minify: false,
   allowOverwrite: true,
   packages: "external",
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -1,10 +1,8 @@
-#!/usr/bin/env ts-node-transpile-only
-
-import esbuild, { type BuildOptions } from "esbuild";
+import esbuild from "esbuild";
 import { copyFile, readFile, writeFile, rm } from "fs/promises";
 import { glob } from "glob";
 
-const sharedOptions: BuildOptions = {
+const sharedOptions = {
   sourcemap: true,
   minify: false,
   allowOverwrite: true,
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -24,8 +24,6 @@ async function main() {
 
   // Remove the types file from the dist-src folder
   const typeFiles = await glob([
-    "./pkg/dist-src/types.js",
-    "./pkg/dist-src/types.js.map",
     "./pkg/dist-src/**/types.js.map",
     "./pkg/dist-src/**/types.js",
   ]);
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -3,7 +3,8 @@ import { copyFile, readFile, writeFile, rm } from "fs/promises";
 import { glob } from "glob";
 
 const sharedOptions = {
-  sourcemap: true,
+  sourcemap: "external",
+  sourcesContent: false,
   minify: false,
   allowOverwrite: true,
   packages: "external",
@@ -20,6 +21,7 @@ async function main() {
     platform: "neutral",
     format: "esm",
     ...sharedOptions,
+    sourcemap: false
   });
 
   // Remove the types file from the dist-src folder
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -75,7 +75,7 @@ async function main() {
         files: ["dist-*/**", "bin/**"],
         main: "dist-node/index.js",
         module: "dist-web/index.js",
-        types: "dist-src/index.d.ts",
+        types: "dist-types/index.d.ts",
         source: "dist-src/index.js",
         sideEffects: false,
       },
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -8,6 +8,7 @@ const sharedOptions: BuildOptions = {
   sourcemap: true,
   minify: false,
   allowOverwrite: true,
+  packages: "external",
 }
 
 async function main() {
@@ -34,7 +35,6 @@ async function main() {
     outdir: "pkg/dist-node",
     bundle: true,
     platform: "node",
-    packages: "external",
     target: "node14",
     format: "cjs",
     ...sharedOptions,
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -1,0 +1,73 @@
+#!/usr/bin/env ts-node-transpile-only
+
+import esbuild, { type BuildOptions } from "esbuild";
+import { copyFileSync, readFileSync, writeFileSync } from "fs";
+import { globSync as glob } from "glob";
+
+const sharedOptions: BuildOptions = {
+  sourcemap: true,
+  minify: false,
+  allowOverwrite: true,
+}
+
+async function main() {
+  // Build the source code for a neutral platform as ESM
+  await esbuild.build({
+    entryPoints: [...glob("./src/*.ts"), ...glob("./src/**/*.ts")],
+    outdir: "pkg/dist-src",
+    bundle: false,
+    platform: "neutral",
+    format: "esm",
+    ...sharedOptions,
+  });
+
+  const entryPoints = [
+    ...glob("./pkg/dist-src/*.js"),
+    ...glob("./pkg/dist-src/**/*.js"),
+  ];
+
+  // Build the source code as a bundle for Node.js as CJS
+  await esbuild.build({
+    entryPoints,
+    outdir: "pkg/dist-node",
+    bundle: true,
+    platform: "node",
+    packages: "external",
+    target: "node14",
+    format: "cjs",
+    ...sharedOptions,
+  });
+  // Build the source code as a bundle for the browser as ESM
+  await esbuild.build({
+    entryPoints,
+    outdir: "pkg/dist-web",
+    bundle: true,
+    platform: "browser",
+    format: "esm",
+    ...sharedOptions,
+  });
+
+  // Copy the README, LICENSE and package.json to the pkg folder
+  copyFileSync("LICENSE.md", "pkg/LICENSE");
+  copyFileSync("README.md", "pkg/README.md");
+  let pkg = JSON.parse(readFileSync("package.json", "utf8").toString());
+  delete pkg.scripts;
+  delete pkg.prettier;
+  delete pkg.release;
+  writeFileSync(
+    "pkg/package.json",
+    JSON.stringify(
+      {
+        ...pkg,
+        files: ["dist-*/**", "bin/**"],
+        main: "./dist-node/index.js",
+        module: "./dist-web/index.js",
+        types: "./dist-src/index.d.ts",
+        source: "./dist-src/index.js",
+      },
+      null,
+      2
+    )
+  );
+}
+main();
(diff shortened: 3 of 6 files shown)


kfcampbell:
I'm :+1: on esbuild and tsc, and this approach seems reasonable to me.

scripts/build.ts
```diff
@@ -0,0 +1,86 @@
...
+    entryPoints,
+    outdir: "pkg/dist-web",
+    bundle: true,
+    platform: "browser",
+    format: "esm",
+    ...sharedOptions,
+  });
+
+  // Copy the README, LICENSE to the pkg folder
+  copyFileSync("LICENSE.md", "pkg/LICENSE");
```
oscard0m: We are missing the format of 
```suggestion
  copyFileSync("LICENSE.md", "pkg/LICENSE.md");
```

oscard0m (commented):
# ❤️‍🔥 Feedback
Amazing job @wolfy1339! Thanks for taking the time and effort on this. This will have a lot of benefits for the community and us 👏🏽 🥳 

I checked your changes and ran `diff` to compare both outputs. I see some differences, but I'm unsure if they are expected.

`Node version used`: `14.21.3`
`npm version used`: `9.6.0`

## 📦⚖️  Output differences
I see some differences which I'm not sure if they are expected or not:
### `pkg/LICENSE.md`
missing `.md` extension. Is it intended?

### `pkg/package.json`
- The `author` appears with ESBuild. I think this is an improvement 👏🏽 
- Fixed `files` patterns ✅ 
- With `pika`, the `main`, `module`, `types` and `source` paths were not starting with a relative path (`./`). Is this an intended change?

### `pkg/dist-node/index.js` and `pkg/dist-web/index.js`
I see noticeable file size differences:
image

image

Checking the differences, I see some variable naming differences and spacing/indentation, so I'm assuming ESBuild does a better job there. Still, the file size increases ⚠️ from `19kb` to `21kb`, which I guess is something we don't want.

### `pkg/dist-src/*`
We are emitting source maps, but I think we don't want the `dist-src/`, right? I guess the non-source-map files increased their size because of sourcemaps.
image

### `pkg/dist-types`
It looks identical to me 🍾 ✅ 


## 📊 Benchmarks
I leave here some screenshots of benchmarks I run on my local machine.
**command:** `hyperfine --runs 3 'npm run build'`

I don't have much experience with `esbuild` but I was expecting better execution times. Maybe is `tsc` consuming our time here? Or maybe the `js` file invoking `esbuild`? Since the time is almost the same this is something I would be happy to nerd about in a different PR 🤓 unless you see any easy quick-win here.

I would be happy to support you here. Let me know if you want to split the work or 
### Pika (`main` branch)
image

### ESBuild (`esbuild` branch)
image

## Migration Plan
Do you have any plan in mind for extending this to the rest of the repositories of the ecosystem? I'm happy to help with octoherd-scripts or any other approach we plan to do to migrate this.

package.json
```diff
@@ -49,17 +49,15 @@
     "@jest/types": "^29.0.0",
     "@octokit/tsconfig": "^1.0.1",
     "@octokit/webhooks-schemas": "6.10.0",
-    "@pika/pack": "^0.3.7",
-    "@pika/plugin-build-node": "^0.9.2",
-    "@pika/plugin-build-web": "^0.9.2",
-    "@pika/plugin-ts-standard-pkg": "^0.9.2",
```
oscard0m, line 55: We need to update `.gitignore` comment as well: `// # Pika builds`

scripts/build.ts
```diff
@@ -0,0 +1,86 @@
...
+    entryPoints,
+    outdir: "pkg/dist-node",
+    bundle: true,
+    platform: "node",
+    target: "node14",
+    format: "cjs",
+    ...sharedOptions,
+  });
+  // Build the source code as a bundle for the browser as ESM
+  await esbuild.build({
```
oscard0m: Would it speed things up if we wrap these commands in a `Promise.all()`?
```js
await Promise.all([
  esbuild.build(...), // dist-node
  esbuild.build(...), // dist-web
  esbuild.build(...), // dist-src
])`
```

scripts/build.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const entryPoints = ["./pkg/dist-src/index.js"];
+
+  // Build the source code as a bundle for Node.js as CJS
+  await esbuild.build({
+    entryPoints,
+    outdir: "pkg/dist-node",
+    bundle: true,
+    platform: "node",
+    target: "node14",
```
oscard0m: 💅🏽 It would be nice to correlate this with `packageJson.engines.node`, but we can tackle this in a different PR.

```suggestion
    target: pkgJson.engines.node,
```

scripts/build.ts
```diff
@@ -0,0 +1,86 @@
...
+        main: "./dist-node/index.js",
+        module: "./dist-web/index.js",
+        types: "./dist-src/index.d.ts",
+        source: "./dist-src/index.js",
+      },
+      null,
+      2
+    )
+  );
+}
```
oscard0m: I would check if making the `read/write` actions async gives us some benefits.

wolfy1339:
> With pika, the main, module, types and source paths were not starting with a relative path (./). Is this an intended change?

I can fix that

> I don't have much experience with esbuild but I was expecting better execution times. Maybe is tsc consuming our time here?

I'm not sure. We can certainly play around with this later.
One thing that might help is not using `ts-node` for the build script.

> Do you have any plan in mind for extending this to the rest of the repositories of the ecosystem? I'm happy to help with octoherd-scripts or any other approach we plan to do to migrate this.

I didn't have any specific plan.
An octoherd script can do most of the work, but since we modify `package.json` and `package-lock.json` we will need manual input

scripts/build.ts
wolfy1339: We have to do `build-src` first, as the other builds are based off of it.

For the other 2, we can wrap them in `Promise.all()`

scripts/build.ts
```diff
@@ -0,0 +1,86 @@
...
+const sharedOptions: BuildOptions = {
+  sourcemap: true,
+  minify: false,
+  allowOverwrite: true,
+  packages: "external",
+};
+
+async function main() {
+  // Start with a clean slate
+  rmSync("pkg", { recursive: true });
```
oscard0m: 💅🏽 I found this plugin: https://github.com/AkaraChen/esbuild-plugin-clean/blob/main/esm.mjs. 

Maybe we can use it or take it as inspiration for our purposes: `build.onStart(...)`

scripts/build.ts
```diff
@@ -0,0 +1,86 @@
...
+    outdir: "pkg/dist-web",
+    bundle: true,
+    platform: "browser",
+    format: "esm",
+    ...sharedOptions,
+  });
+
+  // Copy the README, LICENSE to the pkg folder
+  copyFileSync("LICENSE.md", "pkg/LICENSE.md");
+  copyFileSync("README.md", "pkg/README.md");
```
oscard0m: I found these 2 plugins which maybe we want to keep track of:
- https://github.com/nickjj/esbuild-copy-static-files
- https://github.com/sprout2000/esbuild-copy-plugin
- https://github.com/ozanozbek/esbuild-plugin-copy-file

package.json
```diff
@@ -7,7 +7,7 @@
   "license": "MIT",
   "author": "Gregor Martynus (https://twitter.com/gr2m)",
   "scripts": {
-    "build": "pika-pack build",
+    "build": "ts-node --transpile-only scripts/build.ts && tsc -p tsconfig.json",
```
oscard0m: Maybe we can get better timings with [`tsup`](https://tsup.egoist.dev/#generate-declaration-file) by esbuild? I will experiment with that in the upcoming days but I don't want to block you.

scripts/build.ts
wolfy1339: It would be nice, but the `engines` field isn't just a node version, it's a semver expression

oscard0m: True. It's not an easy task.

scripts/build.ts
wolfy1339: Done in 1cdd9b4f15c2d85ad5d54a76006e234a42addf94

scripts/build.ts
wolfy1339: Switching to the Promise based `fs` api, and using async glob, it is roughly a second faster

package.json
wolfy1339: We can try `tsup`, I'm not sure if you can specify different output folders for the different builds.

I just went this way to keep the same package structure

package.json
wolfy1339: Done in 7d570d736f90104e09ed029ccb95a6d84de72766

wolfy1339:
With all changes since your review @oscard0m, the build time has reduced by 2 seconds.
```
$ hyperfine --runs 3 'npm run build'
Benchmark 1: npm run build
  Time (mean ± σ):      7.026 s ±  0.204 s    [User: 14.463 s, System: 0.413 s]
  Range (min … max):    6.864 s …  7.255 s    3 runs
```

That's 4 seconds faster than Pika

```
$ node --version
v18.14.0
$ npm --version
9.3.1
```

wolfy1339:
Should we change the current value of the `module` field to point to `dist-src/` and add a new `browser` field that points to `dist-web/` ?

This would alleviate issues like https://github.com/octokit/auth-app.js/issues/59 until work on `octokit-next` is finished

```diff
diff --git a/scripts/build.mjs b/scripts/build.mjs
index 3ef8688..73d9ef9 100644
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -72,9 +71,9 @@ async function main() {
         ...pkg,
         files: ["dist-*/**", "bin/**"],
         main: "dist-node/index.js",
-        module: "dist-web/index.js",
+        browser: "dist-web/index.js",
         types: "dist-src/index.d.ts",
-        source: "dist-src/index.js",
+        module: "dist-src/index.js",
       },
```

ollie-iterators:
What is blocking this PR from getting merged?

wolfy1339:
Nothing specific at the moment 

It just needs a review and approval from another maintainer

oscard0m:
@ollie-iterators @wolfy1339 I plan to review this on Thursday. 

Sorry for the late reply here.

scripts/build.mjs
```diff
@@ -0,0 +1,84 @@
...
+const sharedOptions = {
+  sourcemap: true,
+  minify: false,
+  allowOverwrite: true,
+  packages: "external",
+};
+
+async function main() {
+  // Start with a clean slate
+  await rm("pkg", { recursive: true });
```
oscard0m: This line fails in local development when the `pkg/` folder does not exist. Maybe we can apply the [`force: true` option](https://nodejs.org/docs/latest-v18.x/api/fs.html#fsrmpath-options-callback) here?

```suggestion
  await rm("pkg", { recursive: true, force: true });
```

oscard0m:
> Should we change the current value of the `module` field to point to `dist-src/` and add a new `browser` field that points to `dist-web/` ?
> 
> This would alleviate issues like [octokit/auth-app.js#59](https://github.com/octokit/auth-app.js/issues/59) until work on `octokit-next` is finished
> 
> ```diff
> diff --git a/scripts/build.mjs b/scripts/build.mjs
> index 3ef8688..73d9ef9 100644
> --- a/scripts/build.mjs
> +++ b/scripts/build.mjs
> @@ -72,9 +71,9 @@ async function main() {
>          ...pkg,
>          files: ["dist-*/**", "bin/**"],
>          main: "dist-node/index.js",
> -        module: "dist-web/index.js",
> +        browser: "dist-web/index.js",
>          types: "dist-src/index.d.ts",
> -        source: "dist-src/index.js",
> +        module: "dist-src/index.js",
>        },
> ```

I would do that in a separate PR (feel free to create a new issue mentioning this comment) so we do not add more things on top of this migration. I think it's good the differences are minimum in this PR. What do you think?

scripts/build.mjs
```diff
@@ -0,0 +1,84 @@
+import esbuild from "esbuild";
+import { copyFile, readFile, writeFile, rm } from "fs/promises";
+import { glob } from "glob";
+
+const sharedOptions = {
+  sourcemap: true,
```
oscard0m: I'm seeing noticeable differences in the generated sourcemaps from one version and the other. The 2 main differences I see are: 

- `Pika`'s generated sourcemaps are pointing to `dist-src/` and the ones generated by ESBuild are pointing to `src/`. I guess we don't want this so we need the `sourcemaps: "external"` option. Does this make sense?
- `Pika`'s generated sourcemaps files are not including `sourcesContent` but the ones generated by ESBuild yes. I guess we don't want to emit that by using `sourcesContent: false`. What do you think?

```suggestion
  sourcemap: external,
  sourcesContent: false
```

scripts/build.mjs
```diff
@@ -0,0 +1,84 @@
...
+  // Start with a clean slate
+  await rm("pkg", { recursive: true });
+  // Build the source code for a neutral platform as ESM
+  await esbuild.build({
+    entryPoints: await glob(["./src/*.ts", "./src/**/*.ts"]),
+    outdir: "pkg/dist-src",
+    bundle: false,
+    platform: "neutral",
+    format: "esm",
+    ...sharedOptions,
```
oscard0m, line 23: We were not emitting sourcemaps in `dist-src/` before so I think we should preserve that.

```suggestion
    ...sharedOptions,
    sorucemaps: false
```

scripts/build.mjs
```diff
@@ -0,0 +1,84 @@
...
+    "pkg/package.json",
+    JSON.stringify(
+      {
+        ...pkg,
+        files: ["dist-*/**", "bin/**"],
+        main: "dist-node/index.js",
+        module: "dist-web/index.js",
+        types: "dist-src/index.d.ts",
+        source: "dist-src/index.js",
+      },
```
oscard0m: comparing outputs, seems like `pika` was generating the `package.json` file with [`sideEffects: false`](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free). To prevent this to be potentially a breaking change, I would keep it.

```suggestion
      {
        ...pkg,
        files: ["dist-*/**", "bin/**"],
        main: "dist-node/index.js",
        module: "dist-web/index.js",
        types: "dist-src/index.d.ts",
        source: "dist-src/index.js",
        sideEffects: false
      },
```

oscard0m (changes requested):

scripts/build.mjs
wolfy1339: I'll have to test that to make sure it doesn't negatively effect the other builds

scripts/build.mjs
wolfy1339: Wouldn't it be better to have them point to the source TS files instead of `dist-src/`?

That way the files can stay independent from each other

wolfy1339:
> > Should we change the current value of the `module` field to point to `dist-src/` and add a new `browser` field that points to `dist-web/` ?
> > 
> > This would alleviate issues like [octokit/auth-app.js#59](https://github.com/octokit/auth-app.js/issues/59) until work on `octokit-next` is finished
> > 
> > ```diff
> > diff --git a/scripts/build.mjs b/scripts/build.mjs
> > index 3ef8688..73d9ef9 100644
> > --- a/scripts/build.mjs
> > +++ b/scripts/build.mjs
> > @@ -72,9 +71,9 @@ async function main() {
> >          ...pkg,
> >          files: ["dist-*/**", "bin/**"],
> >          main: "dist-node/index.js",
> > -        module: "dist-web/index.js",
> > +        browser: "dist-web/index.js",
> >          types: "dist-src/index.d.ts",
> > -        source: "dist-src/index.js",
> > +        module: "dist-src/index.js",
> >        },
> > ```
> 
> I would do that in a separate PR (feel free to create a new issue mentioning this comment) so we do not add more things on top of this migration. I think it's good the differences are minimum in this PR. What do you think?

Sure thing 👍

Let's do it that way

scripts/build.mjs
oscard0m: We are not publishing the `/src/` folder with the `.ts` files so I think that would not work. Am I correct in my assumption?

image

package.json
oscard0m: Sure, let's not add more complexity to this and stay with what we have.

scripts/build.ts
oscard0m: Let's keep this for a future and not add more complexity to this PR

scripts/build.ts
oscard0m: Let's keep this for a future and not add more complexity to this PR

oscard0m:
> I didn't have any specific plan. An octoherd script can do most of the work, but since we modify `package.json` and `package-lock.json` we will need manual input

Let's have the first version for `webhooks.js` and once we are happy we work an octoherd script together :)

oscard0m:
@wolfy1339 let me know if I can help you with this in some way. I have some availability for pair-program as well.

scripts/build.mjs
wolfy1339: This makes it so that the other builds reference `dist-src/`

scripts/build.mjs
wolfy1339: The map contains the source files within the `sourceContent` property. Which makes them completely indepedent

scripts/build.mjs
wolfy1339: Let's just make it output the same as Pika for now, and we can revisit these points later

scripts/build.mjs
wolfy1339: Let's just make it output the same as Pika for now, and we can revisit these points later

wolfy1339:
I addressed the review comments and rebased the branch

oscard0m (dismissed):
Awesome job @wolfy1339 ! Thanks for the time and effort. Do you think it's interesting to do a beta release first? I would like to try upgrading the package in local for other octokit repos and also try the sourcemaps/bundle size once published in npm.

wolfy1339:
We could definitely do a beta release first

wolfy1339:
Actually, that would conflict with the current `beta` branch which holds v11 in #792

wolfy1339:
This won't trigger a release, as it only touches the build system

oscard0m:
> This won't trigger a release, as it only touches the build system

Or we mark it as a breaking change (in alpha) or we do a second PR in alpha which triggers a release.

Going to bed now. I read you tomorrow morning CET

wolfy1339:
> > This won't trigger a release, as it only touches the build system
> 
> Or we mark it as a breaking change (in alpha) or we do a second PR in alpha which triggers a release.
> 
> Going to bed now. I read you tomorrow morning CET

The semantic release configuration doesn't have alpha set up.

Anyways, I mirrored this branch to `next` and created an empty commit to trigger a patch release. It is available on the `@next` npm tag

scripts/build.mjs
```diff
@@ -0,0 +1,87 @@
...
+  delete pkg.release;
+  await writeFile(
+    "pkg/package.json",
+    JSON.stringify(
+      {
+        ...pkg,
+        files: ["dist-*/**", "bin/**"],
+        main: "dist-node/index.js",
+        module: "dist-web/index.js",
+        types: "dist-src/index.d.ts",
```
oscard0m: I'm trying [this branch](https://github.com/octokit/app.js/pull/405) for `@octokit/app.js` and found the following issue:

```suggestion
        types: "dist-types/index.d.ts",
```

scripts/build.mjs
```diff
@@ -0,0 +1,87 @@
+import esbuild from "esbuild";
+import { copyFile, readFile, writeFile, rm } from "fs/promises";
+import { glob } from "glob";
+
+const sharedOptions = {
+  sourcemap: "external",
+  sourcesContent: false,
```
oscard0m:
```suggestion
  sourcesContent: true,
```
> Let's just make it output the same as Pika for now, and we can revisit these points later

I'm sorry here, I don't know what I was checking but the `sourceContent` is served in Pika so I think we should be keeping it for ESBuild for the moment.

oscard0m:
Thanks @wolfy1339! Do you mind pinging me here when `next` branch is up to date with these changes and release in @next done so I try again the test I'm doing in `@octokit/app.js`?

If you don't have time because is late in your timezone I can take it.

---

EDIT: I'm applying changes to `next` branch FYI

oscard0m:
The changes look good to me. I applied them to `next` branch ([`v10.7.3`](https://www.npmjs.com/package/@octokit/webhooks/v/10.7.3)) and [the draft PR of `@octokit/app.js`](https://github.com/octokit/app.js/pull/405) using this version looks good.

There are some small differences but I'm assuming these are differences/optimizations ESBuild does that Pika was doing in a different way.

oscard0m (dismissed):

oscard0m:
@wolfy1339 I had an idea on what we could do to make it easier the scaffolding for the rest of the repositories (and maintain it in an easy way): https://github.com/octokit/webhooks.js/pull/829

Let me know your thoughts here.

oscard0m (approved):

oscard0m:
@wolfy1339 I resolved the conflicts. Are we ready to merge? To trigger a release, should the PR title be `fix` instead?

wolfy1339:
The releases are kinda screwed up.

Since the `next` branch has a few releases that take up some patch versions, we'll need to merge that into `main` for it to trigger a proper release.

It's not ideal, as we didn't have alpha releases enabled and the beta release was taken up by `v11`

oscard0m:
> The releases are kinda screwed up.
> 
> Since the `next` branch has a few releases that take up some patch versions, we'll need to merge that into `main` for it to trigger a proper release.
> 
> It's not ideal, as we didn't have alpha releases enabled and the beta release was taken up by `v11`

Is not possible to have 10.7.1 in `latest` even we are in 10.7.3 in `next`?

wolfy1339:
Nope. You can only use a version number once, no matter the release channel.

You can overwrite a release, but I don't think semantic release is capable

oscard0m:
> Nope. You can only use a version number once, no matter the release channel.
> 
> You can overwrite a release, but I don't think semantic release is capable

Let's merge `next` in `main` then. The release notes look decent to me.

PR #829, opened 2023-04-07
refactor(esbuild): wrap all the logic inside a plugin

oscard0m:
# Description
Wrap all the logic of `scripts/build.mjs` in an in-house ESBuild plugin.

The idea would be to publish our ESBuild Octokit plugin and centralize the changes we want to do in the future.

We would need a `/scripts/build.mjs` which calls ESBuild with this plugin but that should remain static (and still would have the door open to passing options to our plugin in case we have a repo with special requirements).

@wolfy1339 What do you think about this?


wolfy1339:
This looks good and I like the idea.

We'd most likely have to get another repo and npm package set up for this

oscard0m:
> This looks good and I like the idea.
> 
> We'd most likely have to get another repo and npm package set up for this

I propose doing this: 
1. Let's land https://github.com/octokit/webhooks.js/pull/819 and we can see if this causes friction to users or not and land the appropriate fixes if necessary. Let's wait before scaffolding this to the rest of the repositories.
2. I'll update this PR to make it a refactor of what you land in step 1 and I will tag the core team to see if they like the idea and ask for that repo and npm package.
3. Depending if they take time to answer or not we can publish it in our personal GitHub until we get an answer but I hope to get an answer soon (and I hope they like the idea).

Sounds good to you? @wolfy1339

wolfy1339:
Absolutely sounds good 👍

scripts/esbuild-octokit-plugin.js
```diff
@@ -0,0 +1,100 @@
...
+
+  return {
+    name: "octokit-bundles",
+    async setup(build) {
+      // Run the build steps after the main build
+      build.onEnd(async (result) => {
+        if (result.errors.length) return;
+
+        // Start with a clean slate
+        await fs.rm("pkg", { recursive: true, force: true });
```
wolfy1339, line 23: This should be moved out of the `build.onEnd()` function. It would delete the build folder after the build has completed

wolfy1339:
This idea hasn't gotten any traction, and the PR is out of date.

PR #832, opened 2023-04-07
revert(action): update actions/add-to-project action to v0.5.0 (#827)"

oscard0m:
This reverts commit d64f69a450e4c7c9f9f1db7c322c829d9eda71a5.


wolfy1339:
This isn't the commit that needs to be reverted, as it doesn't trigger a release

You'll need to revert #830 as that is the commit that triggered a release

PR #835, opened 2023-04-08
Next

oscard0m:
Resolves #819


oscard0m:
I wonder if this branch has different protection rules for merging @wolfy1339 
I think the configuration is at repository level and not per branch but let's try.

wolfy1339 (dismissed):

wolfy1339:
The problem is that the `main` branch is protected.

wolfy1339 (dismissed):

wolfy1339:
Replaces #819

kfcampbell (approved):

PR #841, opened 2023-04-15
feat(types): use OpenAPI spec

wolfy1339:
Resolves #840

----

### Before the change?

* The types were made in house by using example payloads and extrapolated a schema from them

### After the change?

* The types use the official GitHub OpenAPI spec

### Other information

*

----

### Pull request checklist
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Added the appropriate label for the given change

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/main/community/breaking_changes.md) to help!

- [x] Yes (Please add the `Type: Breaking change` label)
- [ ] No

If `Yes`, what's the impact:

* Since we are switching the source of the types, I feel it better to release this as a breaking change

### Pull request type

Please add the corresponding label for change this PR introduces:
- Feature/model/API additions: `Type: Feature`

----

PR #849, opened 2023-05-13
ci: stop testing against NodeJS v14, v16

wolfy1339:
BREAKING CHANGE: Drop support for NodeJS v14, v16


gr2m (commented):
I think I'd remove both node 14 and node 16. Node 16 is nearly out of maintenance: https://nodejs.dev/en/about/releases/. If anyone complaints, they can still use the older versions and we can still accept patches for older versions. That will save us another breaking change release within a few months

Also when doing breaking changes, I'd always look for deprecations and remove them as well.

/cc @octokit/js

wolfy1339:
> I think I'd remove both node 14 and node 16. Node 16 is nearly out of maintenance: https://nodejs.dev/en/about/releases/.

I hadn't really thought of Node 16

> If anyone complaints, they can still use the older versions and we can still accept patches for older versions. That will save us another breaking change release within a few months

Good point. I'll update my octoherd script and run it again with these changes

G-Rath:
Inversely: why are we dropping support for v14? I know that it'll mean we might not have to transpile as much, but I wouldn't expect the wins to be that great so unless there's a particular feature or dependency we know we can be using with v14/v16, or there's another breaking change we want to land (which might be #841?), is there a good reason to drop support for these versions?

wolfy1339:
We will still support those versions with bug fixes for the near future.

Dropping support means easier transition to ESM, and no need to keep polyfills for the fetch API.

It just doesn't make much sense to support versions that are EOL

There was a discussion around this briefly, https://github.com/orgs/octokit/discussions/21

G-Rath:
> We will still support those versions with bug fixes for the near future.

In that case we should be testing against them? You either are or are not supporting the versions right? 🙂

> There was a discussion around this briefly, https://github.com/orgs/octokit/discussions/21

It doesn't look like anything was ever actually agreed from that discussion - I've posted a comment with my thoughts on it.

wolfy1339:
> In that case we should be testing against them? You either are or are not supporting the versions right? :slightly_smiling_face:

As gr2m said, we can support them with patches in maintenance releases. In the current release, we won't be supporting them
> If anyone complaints, they can still use the older versions and we can still accept patches for older versions.

kfcampbell (dismissed):

nickfloyd:
https://github.com/orgs/octokit/discussions/44

kfcampbell (approved):

PR #917, opened 2023-11-14
fix: handles verify error

nickfloyd:


gr2m:
we have a problem, the build did not work as expected. Compare

https://unpkg.com/browse/@octokit/webhooks@9.26.0/

and

https://unpkg.com/browse/@octokit/webhooks@9.26.1/

I feel like we have seen this before. I think @wolfy1339 figured it out in the past? I'm not sure. I know we moved away from pika because of its build problems, but we really need to get this one out correctly

wolfy1339:
We also had a patch to fix the `package.json` before we had switched away from pika.

I think that is what's needed here

gr2m:
can you find the patch? 🙏🏼 sorry I've to run in a second and won't be able to help until later today

gr2m:
I found this PR that replaces pika with the new build setup, but I don't see a pika patch in the previous setup?
https://github.com/octokit/request.js/pull/584/files

I also looked into the release build for 9.x and it looks good, as if all folders were created correctly:
https://github.com/octokit/webhooks.js/actions/runs/6879227572/job/18710678481#step:5:14

It also worked as expected locally

nickfloyd:
@wolfy1339 Was this it? https://github.com/octokit/webhooks.js/commit/e178758828bd0b2e60cecf25ce68547b4120e5e0

gr2m:
I tried to just trigger another release and also added an `ls pkg/` step for debugging, the files are there, but they are not published.

gr2m:
I did revert to 9.26.0 for the 9.x release channel

```
npm dist-tag add @octokit/webhooks@9.26.0 release-9.x
```

PR #920, opened 2023-11-15
Reduce json stringify

Uzlopak:
We can reduce the amount of JSON.stringify calls for the obvious cases. ;)

### DIFF
--- a/src/middleware/node/middleware.ts
+++ b/src/middleware/node/middleware.ts
@@ -27,11 +27,7 @@ export async function middleware(
     response.writeHead(422, {
       "content-type": "application/json",
     });
-    response.end(
-      JSON.stringify({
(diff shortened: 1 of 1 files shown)


G-Rath:
I don't think this is a good change - we're using `JSON.stringify` exactly what it's meant for: so we don't have to manually craft JSON or worry about escaping etc; while yes these cases might be safer to write out directly, it means going forward we should be making people decide "is this JSON simple enough to not use `JSON.stringify`?".

It's also possible refractors could result in dynamic content that needs to be escaped gets introduced, which `JSON.stringify` would handle for us - better to just always use `JSON.stringify` and not have to think about it

gr2m:
I agree with @G-Rath. I don't think the optimization is worth it for code readability / maintainability reasons. `JSON.stringify` is very efficient in modern runtimes.

PR #921, opened 2023-11-16
fix: only parse payload once

### LINKED ISSUE #922: Avoid parsing JSON payload twice.
n.T.

Uzlopak:
Closes #922 

We are parsing the JSON twice. Also it is quite dangerous to parse unverified data. So it is better to first verify and then parse it with the very low risk that it is broken payload. 

Improves the performance of the valid path in probot beta from 63 k ops/s to 71 k ops/s.

### DIFF
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -26,15 +26,7 @@ export function getPayload(request: IncomingMessage): Promise<string> {
     request.on("error", (error: Error) => reject(new AggregateError([error])));
     request.on("data", (chunk: string) => (data += chunk));
     request.on("end", () => {
-      try {
-        // Call JSON.parse() only to check if the payload is valid JSON
-        JSON.parse(data);
-        resolve(data);
-      } catch (error: any) {
-        error.message = "Invalid JSON";
-        error.status = 400;
-        reject(new AggregateError([error]));
-      }
+      resolve(data);
     });
   });
 }
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -25,8 +25,6 @@ export function getPayload(request: IncomingMessage): Promise<string> {
     // istanbul ignore next
     request.on("error", (error: Error) => reject(new AggregateError([error])));
     request.on("data", (chunk: string) => (data += chunk));
-    request.on("end", () => {
-      resolve(data);
-    });
+    request.on("end", () => resolve(data));
   });
 }
--- a/src/verify-and-receive.ts
+++ b/src/verify-and-receive.ts
@@ -36,11 +36,6 @@ export async function verifyAndReceive(
   } catch (error: any) {
     error.message = "Invalid JSON";
     error.status = 400;
-    let tmpStackTraceLimit = Error.stackTraceLimit;
-    Error.stackTraceLimit = 0;
-    const aggregateError = new AggregateError([error]);
-    Error.stackTraceLimit = tmpStackTraceLimit;
-    aggregateError.stack = error.stack;
-    throw aggregateError;
+    throw new AggregateError([error]);
   }
 }
--- a/src/verify-and-receive.ts
+++ b/src/verify-and-receive.ts
@@ -1,10 +1,11 @@
+import AggregateError from "aggregate-error";
 import { verify } from "@octokit/webhooks-methods";
 
 import type {
+  EmitterWebhookEvent,
   EmitterWebhookEventWithStringPayloadAndSignature,
   State,
 } from "./types";
-import AggregateError from "aggregate-error";
 
 export async function verifyAndReceive(
   state: State & { secret: string },
@@ -27,15 +28,18 @@ export async function verifyAndReceive(
     );
   }
 
+  let payload: EmitterWebhookEvent["payload"];
   try {
-    return state.eventHandler.receive({
-      id: event.id,
-      name: event.name,
-      payload: JSON.parse(event.payload),
-    });
+    payload = JSON.parse(event.payload);
   } catch (error: any) {
     error.message = "Invalid JSON";
     error.status = 400;
     throw new AggregateError([error]);
   }
+
+  return state.eventHandler.receive({
+    id: event.id,
+    name: event.name,
+    payload,
+  });
 }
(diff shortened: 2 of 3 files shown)


test/integration/node-middleware.test.ts
```diff
@@ -178,6 +178,8 @@ describe("createNodeMiddleware(webhooks)", () => {
     // @ts-expect-error complains about { port } although it's included in returned AddressInfo interface
     const { port } = server.address();
 
+    const payload = '{"name":"invalid"';
```
wolfy1339, line 181: Can you explain why the test needed to be updated?

src/verify-and-receive.ts
```diff
@@ -26,9 +27,20 @@ export async function verifyAndReceive(
...
+  try {
+    return state.eventHandler.receive({
+      id: event.id,
+      name: event.name,
+      payload: JSON.parse(event.payload),
+    });
+  } catch (error: any) {
+    error.message = "Invalid JSON";
+    error.status = 400;
+    let tmpStackTraceLimit = Error.stackTraceLimit;
```
wolfy1339: Can you explain this change? Adding comments to the code is useful for these sorts of things

Uzlopak: Reverted this because on second thoughts i think that nodes prepareStackTrace will get triggered and so being slower than reassigning the stacktrace from the original error to the aggregateError

test/integration/node-middleware.test.ts
Uzlopak: Before it was parsing and then validating. So in the test was an invalid sha256. So it would parse and fail on validation. 

I changed the sha256 to pass the validation but the payload will fail on parsing.

wolfy1339 (dismissed):

wolfy1339:
I'd like to wait for others to review as well

G-Rath:
I personally don't feel like this change is needed and really doubt this is an actual bottleneck for probot (if it is, I'd want to see a more detailed breakdown of the benchmarking before making exploring more of these kind of changes).

Having said that iirc `getPayload` is an internal function and it would be fair to have it just return `string` without any validation (it already does anyway if `request.body` is present).

src/verify-and-receive.ts
```diff
@@ -26,9 +27,15 @@ export async function verifyAndReceive(
...
-  return state.eventHandler.receive({
-    id: event.id,
-    name: event.name,
-    payload: JSON.parse(event.payload),
-  });
+  try {
+    return state.eventHandler.receive({
+      id: event.id,
+      name: event.name,
+      payload: JSON.parse(event.payload),
```
G-Rath: The `try/catch` should be around just the `JSON.parse` because if `eventHandler.receive` throws an error that might not be valid JSON (I'm not sure if this can end up calling user provided functions, but even if it doesn't the cost of just catching `JSON.parse` is very cheap)

src/verify-and-receive.ts
```diff
@@ -26,9 +27,15 @@ export async function verifyAndReceive(
...
+    });
+  } catch (error: any) {
+    error.message = "Invalid JSON";
+    error.status = 400;
+    throw new AggregateError([error]);
+  }
```
wolfy1339:
```suggestion
  let payload: any;
  try {
    payload = JSON.parse(event.payload);
  } catch (error: any) {
    error.message = "Invalid JSON";
    error.status = 400;
    throw new AggregateError([error]);
  }
  
  return state.eventHandler.receive({
      id: event.id,
      name: event.name,
      payload,
  }),
```

G-Rath: `payload` needs to be initialized outside of the `try/catch` to be used - you either need a `let` or to wrap it in a function that returns from within the `try/catch`

wolfy1339: Updated to initialize `payload` outside of the `try`/`catch`

src/verify-and-receive.ts
```diff
@@ -4,11 +4,12 @@ import type {
   EmitterWebhookEventWithStringPayloadAndSignature,
   State,
 } from "./types";
+import AggregateError from "aggregate-error";
```
gr2m: nit: move module imports to the top block below `import { verify } from "@octokit/webhooks-methods";`

we loosely follow the convention

- 1st block: `node:...` imports
- 2nd block: module imports
- 3rd block: relative imports

src/middleware/node/get-payload.ts
```diff
@@ -26,15 +26,7 @@ export function getPayload(request: IncomingMessage): Promise<string> {
...
-        // Call JSON.parse() only to check if the payload is valid JSON
-        JSON.parse(data);
-        resolve(data);
-      } catch (error: any) {
-        error.message = "Invalid JSON";
-        error.status = 400;
-        reject(new AggregateError([error]));
-      }
+      resolve(data);
     });
```
gr2m: I would shorten it to just 

```js
request.on("end", () => resolve(data));
```

gr2m (dismissed):
changes look good to me, just some nits

src/verify-and-receive.ts
Uzlopak: I oversaw your suggestion. Thank you for the suggestion ;)

Uzlopak:
@wolfy1339 
@G-Rath 
@gr2m 
@oscard0m 
PTAL ;)

src/verify-and-receive.ts
```diff
@@ -1,14 +1,16 @@
...
+  EmitterWebhookEvent,
   EmitterWebhookEventWithStringPayloadAndSignature,
   State,
 } from "./types";
 
 export async function verifyAndReceive(
   state: State & { secret: string },
   event: EmitterWebhookEventWithStringPayloadAndSignature,
-): Promise<any> {
+): Promise<void> {
```
oscard0m, line 13: Can you elaborate on why this becomes `void`?

Uzlopak: sure @oscard0m 

It is the return value of eventHandler.receive
see:
https://github.com/octokit/webhooks.js/blob/cc836d042288c5c1216089d3cc5f2afaa58dccb9/src/event-handler/index.ts#L29

oscard0m (approved):

gr2m (approved):

PR #923, opened 2023-11-16
fix(typescript): workaround for trouble with `aggregate-error` type export

G-Rath:
This wasn't copied across in #921 and I believe it's still a problem


gr2m (approved):
thank you!

PR #929, opened 2023-11-21
fix(types): limit recursion in types using ternary statements

### LINKED ISSUE #926: Memory leak/recursive loop in 12.0.7 in the types
@octokit/webhooks v. 12.0.7 seems to introduce some kind of memory leak/recursive loop. v 12.0.6 still works. 

_Originally posted by @Uzlopak in https://github.com/probot/probot/pull/1926#issuecomment-1819771230_

Uzlopak:
Resolves #926

@wolfy1339 

It seems that this results in building again properly on probot. 
But I dont know if this is what you expect.

I used `npm link` to test if it builds again in probot. And it builds again

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/src/event-handler/index.ts
+++ b/src/event-handler/index.ts
@@ -15,7 +15,7 @@ import {
 import { receiverHandle as receive } from "./receive.js";
 import { removeListener } from "./remove-listener.js";
 
-interface EventHandler<TTransformed = unknown> {
+interface EventHandler<TTransformed> {
   on<E extends EmitterWebhookEventName>(
     event: E | E[],
     callback: HandlerFunction<E, TTransformed>,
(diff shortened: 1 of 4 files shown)


src/event-handler/index.ts
```diff
@@ -15,12 +15,18 @@ import {
 import { receiverHandle as receive } from "./receive.js";
 import { removeListener } from "./remove-listener.js";
 
-interface EventHandler<TTransformed> {
+interface EventHandler<TTransformed = unknown> {
```
wolfy1339: Just wondering if it's necessary for this value to be optional? As this interface is never called on it's own, and `TTransformed` is always passed down

Uzlopak: i think it is an artifact from my testing
```suggestion
interface EventHandler<TTransformed> {
```

wolfy1339 (approved):

PR #931, opened 2023-11-24
chore: add serverless integration tests

### LINKED ISSUE #930: [FEAT]: Export createNodeHandler
### Describe the need

<s>If we support serverless in the middleware, we can promote probot to be serverless compatible... well maybe after some minor changes in probot.</s>

From the discussion we come to the conclusion, that serverless tests should be done in probot. But I think we need therefore export the Handler.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

Uzlopak:
Resolves #930 

Adds tests for serverless and ensures that we can claim that we are supporting serverless.

@wolfy1339 

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "jest": "^29.0.0",
         "prettier": "3.1.0",
         "prettier-plugin-packagejson": "^2.2.9",
+        "serverless-http": "^3.2.0",
         "ts-jest": "^29.0.0",
         "ts-node": "^10.0.0",
         "typescript": "^5.0.0"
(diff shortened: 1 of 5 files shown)


wolfy1339:
There is no need to always ping me. I get notifications on all JS Octokit repos, and I also get notifications for Probot repos I have access to

G-Rath:
It's not clear to me what this is covering that we've not already got covered - we've already required 100% code coverage, and this doesn't seem to be testing against every popular serverless runtime environment so I'm not sure how this change would allow us to claim we're serverless compatible anymore than we can now (which I think is fine for us to do without these tests), and not sure if it justifies the maintenance burden of having all the extra tests...

Uzlopak:
Ok, Cabal. ;)

PR #936, opened 2023-11-29
fix: handle gcf, verify already provided request.body

### LINKED ISSUE #935: [BUG]: google-cloud-function, can not verify payload
### What happened?

The request.body is already provided by gcf as an object. We expect that Lambda can provide request.body as a string but as it is an object in gcf it can not verify the request.body, because to verify the request.body is just using toString() of the object but not JSON.stringify.

### Versions

v. 12.0.8

### Relevant log output

_No response_

### Code of Conduct

Uzlopak:
Resolves #935 

As explained in #935 in gcf the payload is already deserialized to be an object, resulting that the payload can not be verified.

Can this be please fast-tracked?

@wolfy1339 
@gr2m 

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -17,7 +17,7 @@ export function getPayload(request: IncomingMessage): Promise<Buffer> {
 
     // istanbul ignore next
     request.on("error", (error: Error) => reject(new AggregateError([error])));
-    request.on("data", data.push.bind(data));
+    request.on("data", (chunk: string) => data.push(chunk);
     // istanbul ignore next
     request.on("end", () =>
       setImmediate(resolve, data.length === 1 ? data[0] : Buffer.concat(data)),
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -12,8 +12,6 @@ import AggregateError from "aggregate-error";
 type IncomingMessage = any;
 
 export function getPayload(request: IncomingMessage): Promise<Buffer> {
-  // If request.body already exists we can stop here
-  // See https://github.com/octokit/webhooks.js/pull/23
   return new Promise((resolve, reject) => {
     let data: Buffer[] = [];
 
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -11,20 +11,15 @@ import AggregateError from "aggregate-error";
 // }
 type IncomingMessage = any;
 
-export function getPayload(request: IncomingMessage): Promise<string> {
+export function getPayload(request: IncomingMessage): Promise<Buffer> {
   // If request.body already exists we can stop here
   // See https://github.com/octokit/webhooks.js/pull/23
-
-  if (request.body) return Promise.resolve(request.body);
-
   return new Promise((resolve, reject) => {
-    let data = "";
-
-    request.setEncoding("utf8");
+    let data: Buffer[] = [];
 
     // istanbul ignore next
     request.on("error", (error: Error) => reject(new AggregateError([error])));
-    request.on("data", (chunk: string) => (data += chunk));
-    request.on("end", () => resolve(data));
+    request.on("data", (chunk: Buffer) => data.push(chunk));
+    request.on("end", () => resolve(Buffer.concat(data)));
   });
 }
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -20,6 +20,8 @@ export function getPayload(request: IncomingMessage): Promise<Buffer> {
     request.on("data", (chunk: string) => data.push(chunk);
     // istanbul ignore next
     request.on("end", () =>
+      // setImmediate improves the throughput by reducing the pressure from
+      // the event loop
       setImmediate(resolve, data.length === 1 ? data[0] : Buffer.concat(data)),
     );
   });
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -17,7 +17,10 @@ export function getPayload(request: IncomingMessage): Promise<Buffer> {
 
     // istanbul ignore next
     request.on("error", (error: Error) => reject(new AggregateError([error])));
-    request.on("data", (chunk: Buffer) => data.push(chunk));
-    request.on("end", () => resolve(Buffer.concat(data)));
+    request.on("data", data.push.bind(data));
+    // istanbul ignore next
+    request.on("end", () =>
+      setImmediate(resolve, data.length === 1 ? data[0] : Buffer.concat(data)),
+    );
   });
 }
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -17,7 +17,7 @@ export function getPayload(request: IncomingMessage): Promise<Buffer> {
 
     // istanbul ignore next
     request.on("error", (error: Error) => reject(new AggregateError([error])));
-    request.on("data", (chunk: string) => data.push(chunk);
+    request.on("data", (chunk: Buffer) => data.push(chunk));
     // istanbul ignore next
     request.on("end", () =>
       // setImmediate improves the throughput by reducing the pressure from
--- a/src/middleware/node/middleware.ts
+++ b/src/middleware/node/middleware.ts
@@ -82,10 +82,10 @@ export async function middleware(
     "x-hub-signature-256": signature,
     "x-github-delivery": id,
   } = request.headers as {
-    'x-github-event': WebhookEventName;
-    'x-hub-signature-256': string;
-    'x-github-delivery': string;
-  }
+    "x-github-event": WebhookEventName;
+    "x-hub-signature-256": string;
+    "x-github-delivery": string;
+  };
   options.log.debug(`${name} event received (id: ${id})`);
 
   // GitHub will abort the request if it does not receive a response within 10s
--- a/src/middleware/node/middleware.ts
+++ b/src/middleware/node/middleware.ts
@@ -77,8 +77,8 @@ export async function middleware(
     return true;
   }
 
-  const eventName = request.headers["x-github-event"] as WebhookEventName;
-  const signatureSHA256 = request.headers["x-hub-signature-256"] as string;
+  const name = request.headers["x-github-event"] as WebhookEventName;
+  const signature = request.headers["x-hub-signature-256"] as string;
   const id = request.headers["x-github-delivery"] as string;
   options.log.debug(`${name} event received (id: ${id})`);
 
--- a/src/middleware/node/middleware.ts
+++ b/src/middleware/node/middleware.ts
@@ -93,12 +93,28 @@ export async function middleware(
   }, 9000).unref();
 
   try {
-    const payload = await getPayload(request);
+    let payload: Buffer;
+    let body: { [key: string]: any } | undefined;
+
+    const bodyType = typeof request.body;
+
+    // The body is a String (e.g. Lambda)
+    if (bodyType === "string") {
+      payload = request.body;
+      // The body is already an Object and rawBody is a Buffer (e.g. GCF)
+    } else if (bodyType === "object" && request.rawBody instanceof Buffer) {
+      body = request.body;
+      payload = request.rawBody;
+      // We need to load the payload from the request (normal case of Node.js server)
+    } else {
+      payload = await getPayload(request);
+    }
 
     await webhooks.verifyAndReceive({
       id: id,
       name: eventName as any,
       payload,
+      body,
       signature: signatureSHA256,
     });
     clearTimeout(timeout);
--- a/src/middleware/node/middleware.ts
+++ b/src/middleware/node/middleware.ts
@@ -77,11 +77,16 @@ export async function middleware(
     return true;
   }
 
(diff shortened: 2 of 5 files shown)


wolfy1339:
Can you make sure that this doesn't effect values that end in a zero decimal, ex: `123.10`

Uzlopak:
@wolfy1339 

I think I can do it even better.

Uzlopak:
@wolfy1339 
I could not follow, what you mean with handling numbers correctly. 

But now it is trying to avoid unecessary serializiation and deserialization based on the input. GCF has body and rawBody, while rawBody is the Buffer. 

Let me test this on GCF and give you feedback :).

Uzlopak:
@wolfy1339 

It works now properly. Can you please explain what your concern is regarding numbers?

wolfy1339:
NodeJS likes to remove trailing zeroes on numbers, and thus it changes the signature of the payload

Uzlopak:
@wolfy1339 

In this case, I just handle the specific case of GCF because there is no way I can ensure, that JSON.stringify and .parse will respect that.

PTAL

Uzlopak:
Tested on GCF and works as expected ;)

src/verify-and-receive.ts
```diff
@@ -29,18 +30,29 @@ export async function verifyAndReceive(
...
+  }
+
+  const payload =
+    typeof event.payload === "string"
+      ? event.payload
+      : event.payload.toString("utf8");
+
   try {
-    payload = JSON.parse(event.payload);
+    return state.eventHandler.receive({
```
wolfy1339: This changes the behaviour of the try/catch

We don't want to catch errors from the `state.eventHandler.receive()` function,
We want to make sure that it is valid JSON

src/middleware/node/get-payload.ts
```diff
@@ -11,20 +11,15 @@ import AggregateError from "aggregate-error";
 // }
 type IncomingMessage = any;
 
-export function getPayload(request: IncomingMessage): Promise<string> {
+export function getPayload(request: IncomingMessage): Promise<Buffer> {
   // If request.body already exists we can stop here
   // See https://github.com/octokit/webhooks.js/pull/23
-
-  if (request.body) return Promise.resolve(request.body);
-
```
wolfy1339: The comment should be moved, updated or removed if it isn't necessary any more.

Uzlopak: done

src/verify-and-receive.ts
Uzlopak: done

src/verify-and-receive.ts
```diff
@@ -29,9 +30,22 @@ export async function verifyAndReceive(
...
+  // The body is already an Object (e.g. GCF) and can be passed directly to the EventHandler
+  if (event.body) {
+    return state.eventHandler.receive({
+      id: event.id,
+      name: event.name,
+      payload: event.body as EmitterWebhookEvent["payload"],
+    });
+  }
+
+  let payload;
```
wolfy1339:
```suggestion
  let payload: EmitterWebhookEvent["payload"];
```

src/verify-and-receive.ts
```diff
@@ -29,9 +30,22 @@ export async function verifyAndReceive(
...
 
-  let payload: EmitterWebhookEvent["payload"];
+  // The body is already an Object (e.g. GCF) and can be passed directly to the EventHandler
+  if (event.body) {
+    return state.eventHandler.receive({
+      id: event.id,
+      name: event.name,
+      payload: event.body as EmitterWebhookEvent["payload"],
+    });
+  }
```
wolfy1339: Is there any way to avoid having to add an extra parameter?

Uzlopak: For performance reasons no ;)

wolfy1339: And what are the performance reasons?
What are the different performance results?

Simply saying for performance reasons isn't going to get this approved. If you find a performance improvement, explain how it's impacting performance with data.

Uzlopak: Fair point.will provide Benchmarks

src/middleware/node/get-payload.ts
```diff
@@ -11,20 +11,13 @@ import AggregateError from "aggregate-error";
...
-    request.setEncoding("utf8");
+    let data: Buffer[] = [];
 
     // istanbul ignore next
     request.on("error", (error: Error) => reject(new AggregateError([error])));
-    request.on("data", (chunk: string) => (data += chunk));
-    request.on("end", () => resolve(data));
+    request.on("data", (chunk: Buffer) => data.push(chunk));
+    request.on("end", () => resolve(Buffer.concat(data)));
```
Uzlopak:
```suggestion
    request.on("data", data.push.bind(data));
    request.on("end", setImmediate(resolve, Buffer.concat(data)));
}
```

Uzlopak: @kibertoad
It seems that the setImmediate improves the performance. Seems like without it the event loop is clogging? Maybe it helps to relax the verification later, which is doing a createHmac and is IO heavy and thus hurting the performance of the event loop?

kibertoad: Buffer concatenation happens immediately in this code, so the only thing deferred is the promise resolution.

ChatGPT has this to say: 
image

Uzlopak: Oh right. If I do () => setImmediate it works.

src/middleware/node/middleware.ts
```diff
@@ -77,11 +77,16 @@ export async function middleware(
...
-  options.log.debug(`${eventName} event received (id: ${id})`);
+  const {
+    "x-github-event": name,
+    "x-hub-signature-256": signature,
+    "x-github-delivery": id,
+  } = request.headers as {
+    "x-github-event": WebhookEventName;
+    "x-hub-signature-256": string;
+    "x-github-delivery": string;
+  };
```
wolfy1339: Why do it this way, it's much more clear and concise the way it was before. The variable names were more descriptive

Uzlopak:
@wolfy1339 

It is hard to provide a benchmark for body attribute, as it is gcf specific. Passing the body object is only relevant for gcf. In other cases it is just undefined and does not effect significantly the performance. In gcf it actually avoids one JSON.parse. 

I really need this PR to land, so i beg you to approve and merge it. 

My benchmarks show a significant performance gain from 73k ops/ 10s to 83k ops/ 10s.

before:
```
┌─────────┬──────┬──────┬───────┬──────┬────────┬─────────┬───────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%  │ Avg    │ Stdev   │ Max   │
├─────────┼──────┼──────┼───────┼──────┼────────┼─────────┼───────┤
│ Latency │ 0 ms │ 1 ms │ 4 ms  │ 5 ms │ 0.8 ms │ 1.08 ms │ 31 ms │
└─────────┴──────┴──────┴───────┴──────┴────────┴─────────┴───────┘
┌───────────┬────────┬────────┬────────┬────────┬────────┬─────────┬────────┐
│ Stat      │ 1%     │ 2.5%   │ 50%    │ 97.5%  │ Avg    │ Stdev   │ Min    │
├───────────┼────────┼────────┼────────┼────────┼────────┼─────────┼────────┤
│ Req/Sec   │ 3287   │ 3287   │ 7767   │ 7919   │ 7255.7 │ 1364.07 │ 3286   │
├───────────┼────────┼────────┼────────┼────────┼────────┼─────────┼────────┤
│ Bytes/Sec │ 411 kB │ 411 kB │ 971 kB │ 990 kB │ 907 kB │ 171 kB  │ 411 kB │
└───────────┴────────┴────────┴────────┴────────┴────────┴─────────┴────────┘

Req/Bytes counts sampled once per second.
# of samples: 10

73k requests in 10.02s, 9.07 MB read
```

after:

```sh
┌─────────┬──────┬──────┬───────┬──────┬─────────┬─────────┬───────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%  │ Avg     │ Stdev   │ Max   │
├─────────┼──────┼──────┼───────┼──────┼─────────┼─────────┼───────┤
│ Latency │ 0 ms │ 0 ms │ 4 ms  │ 4 ms │ 0.46 ms │ 0.99 ms │ 29 ms │
└─────────┴──────┴──────┴───────┴──────┴─────────┴─────────┴───────┘
┌───────────┬────────┬────────┬─────────┬─────────┬─────────┬─────────┬────────┐
│ Stat      │ 1%     │ 2.5%   │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min    │
├───────────┼────────┼────────┼─────────┼─────────┼─────────┼─────────┼────────┤
│ Req/Sec   │ 4711   │ 4711   │ 8623    │ 8959    │ 8254.8  │ 1231.47 │ 4708   │
├───────────┼────────┼────────┼─────────┼─────────┼─────────┼─────────┼────────┤
│ Bytes/Sec │ 589 kB │ 589 kB │ 1.08 MB │ 1.12 MB │ 1.03 MB │ 154 kB  │ 589 kB │
└───────────┴────────┴────────┴─────────┴─────────┴─────────┴─────────┴────────┘

Req/Bytes counts sampled once per second.
# of samples: 10

83k requests in 10.02s, 10.3 MB read
```

src/middleware/node/middleware.ts
```diff
@@ -93,13 +98,30 @@ export async function middleware(
...
+    let payload: Buffer;
+
+    if ("body" in request) {
+      if (
+        typeof request.body === "object" &&
+        "rawBody" in request &&
+        request.rawBody instanceof Buffer
+      ) {
+        // The body is already an Object and rawBody is a Buffer (e.g. GCF)
+        payload = request.rawBody;
```
wolfy1339, line 104: If we are sending the raw body as a Buffer from GCF, then why do we need the JSON obect in a new `body` parameter?

src/verify-and-receive.ts
```diff
@@ -15,6 +15,7 @@ export async function verifyAndReceive(
   // verify will validate that the secret is not undefined
   const matchesSignature = await verify(
     state.secret,
+    // @ts-expect-error verify uses createHmac, which can take Strings and Buffers
```
wolfy1339, line 18: I'd rather not. It's better to fix this at the source

src/middleware/node/middleware.ts
Uzlopak: There is no new body parameter. I removed it to satisfy this for now. We can discuss this when I have more time ;)

src/verify-and-receive.ts
Uzlopak: see https://github.com/octokit/webhooks-methods.js/pull/211

src/middleware/node/middleware.ts
Uzlopak:
```suggestion
  const eventName = request.headers["x-github-event"] as WebhookEventName;
  const signatureSHA256 = request.headers["x-hub-signature-256"] as string;
  const id = request.headers["x-github-delivery"] as string;
```

wolfy1339:
Please let other maintainers the time to review this. Can you split the performance patches from this PR and only handle GCF compatibility please

src/middleware/node/get-payload.ts
```diff
@@ -11,20 +11,16 @@ import AggregateError from "aggregate-error";
...
-    let data = "";
-
-    request.setEncoding("utf8");
+    let data: Buffer[] = [];
 
     // istanbul ignore next
     request.on("error", (error: Error) => reject(new AggregateError([error])));
-    request.on("data", (chunk: string) => (data += chunk));
-    request.on("end", () => resolve(data));
+    request.on("data", data.push.bind(data));
```
gr2m: It's shorter and it's unlikely to break in future, but I'd still prefer we make it more explicit. The usage of `.bind` disappears from code since the use of arrow functions, fewer folks know what it is, how it works, and its drawbacks

```suggestion
    request.on("data", (chunk: string) => data.push(chunk);
```

src/middleware/node/get-payload.ts
```diff
@@ -11,20 +11,16 @@ import AggregateError from "aggregate-error";
...
+    // istanbul ignore next
+    request.on("end", () =>
+      setImmediate(resolve, data.length === 1 ? data[0] : Buffer.concat(data)),
```
gr2m, line 25: why is the setImmediate needed here?

wolfy1339: I saw an explanation in this previous conversation, https://github.com/octokit/webhooks.js/pull/936#discussion_r1409945521

gr2m: @Uzlopak Can you please add a comment to the code about it?

Uzlopak:
```suggestion
      // setImmediate improves the throughput by reducing the pressure from
      // the event loop
      setImmediate(resolve, data.length === 1 ? data[0] : Buffer.concat(data)),
```

wolfy1339:
We have 3 active PRs, of which 2 were split off from this one.

Can you incorporate the feedback of this PR into the other PRs?

#939 for the switch to using `Buffer` instead of `string`
#937 for the changes to fix compatibility with Google Cloud Functions.

Let's continue on in those more specific PRs

Uzlopak:
k

PR #939, opened 2023-11-30
fix: prevent splitting multibyte characters in `getPayload()`

### LINKED ISSUE #938: [BUG]: bodyparser in get-payload potentially buggy
### What happened?

I think @addaleax tweeted long time ago about this issue of bodyparser.

If you expect that the payload is using utf8 or utf16 or any other multybyte encoding, then you have to concat the buffers first and in the final step you can serialize the buffer to a string with the desired encoding. 

The issue will arise, when with some "luck" a multibyte character is split between two chunks.

### Versions

12.0.8

### Relevant log output

_No response_

### Code of Conduct

Uzlopak:
Resolves #938 

Resolves #ISSUE_NUMBER

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/test/integration/get-payload.test.ts
+++ b/test/integration/get-payload.test.ts
@@ -1,5 +1,5 @@
 import EventEmitter from "node:events";
-import { getPayload } from "../../src/middleware/node/get-payload";
+import { getPayload } from "../../src/middleware/node/get-payload.ts";
 
 describe("getPayload", () => {
   it("returns a promise", () => {
--- a/test/integration/get-payload.test.ts
+++ b/test/integration/get-payload.test.ts
@@ -1,0 +1,77 @@
+import EventEmitter from "node:events";
+import { getPayload } from "../../src/middleware/node/get-payload";
+
+describe("getPayload", () => {
+  it("returns a promise", () => {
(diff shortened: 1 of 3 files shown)


wolfy1339:
Can you please add a test to make sure we don't break this behaviour in the future

wolfy1339:
Please take the time to fill out the PR templates and use a more descriptive PR title

Uzlopak:
@wolfy1339 
provided tests as requested. I also run a similar test against the original implementation and got for the multibyte test two broken characters.

Uzlopak:
@wolfy1339 
adapted requested changes

test/integration/get-payload.test.ts
```diff
@@ -0,0 +1,77 @@
+import EventEmitter from "node:events";
+import { getPayload } from "../../src/middleware/node/get-payload";
```
wolfy1339:
```suggestion
import { getPayload } from "../../src/middleware/node/get-payload.ts";
```

wolfy1339 (approved):

PR #962, opened 2024-01-31
fix: handle preparsed body payloads in verifyAndReceive accordingly

### LINKED ISSUE #961: [BUG]: Preparsed payloads are not handled correctly, here Vercel
### What happened?

https://github.com/probot/probot/issues/1955#issuecomment-1919925057

The issue arises from the fact, that vercel preparses the body. So `verify` can not verify the body, as it is not a string.

You can get it to work for now, by setting the  `NODEJS_HELPERS` environment variable to 0 in vercel.

### Versions

current

### Relevant log output

_No response_

### Code of Conduct

Uzlopak:
fixes #961 

https://github.com/probot/probot/issues/1955#issuecomment-1919925057


src/verify-and-receive.ts
```diff
@@ -15,7 +15,9 @@ export async function verifyAndReceive(
   // verify will validate that the secret is not undefined
   const matchesSignature = await verify(
     state.secret,
-    event.payload,
+    typeof event.payload === "string"
+      ? event.payload
+      : JSON.stringify(event.payload),
```
AaronDewes, line 20: This could cause problems if GitHub ever changes payload serialization, so the signature would not match...

We should check if it's possible to disable this parsing in Vercel.

Uzlopak: It is possible https://github.com/probot/probot/issues/1955#issuecomment-1919925057

G-Rath: Seems like you don't even need to do that: https://vercel.com/guides/how-do-i-get-the-raw-body-of-a-serverless-function

AaronDewes: https://github.com/octokit/webhooks.js/pull/963 should be a better implementation based on that guide

Uzlopak: I have no preference. But you should add a test to your PR.

wolfy1339:
> @Uzlopak from the discussion in #963, I think we might need to add the typeof condition here: https://github.com/probot/probot/blob/1eb03896e9772312c84f0be3da7bf3c0910797b6/src/context.ts#L72
> 
> I'm sorry if I'm wrong, but looking at the source code, my mind tells me that it is the place from where the request body is passed to this SDK.

That is the `Context` class, it doesn't call the webhooks at all. It only augments it.

The webhooks aren't really called anywhere, we just pass through the data from the integrated web server and do the handliin this library:
https://github.com/probot/probot/blob/1eb03896e9772312c84f0be3da7bf3c0910797b6/src/create-node-middleware.ts#L14

vedantmgoyal2009:
I had a look again, and I still don't get it. I was thinking to open a PR, but seems like I can only depend on you guys to resolve it.

wolfy1339:
Closing as per discussion in #963

PR #963, opened 2024-02-01
fix: Get raw body properly on Vercel

### LINKED ISSUE #961: [BUG]: Preparsed payloads are not handled correctly, here Vercel

AaronDewes:
Resolves #961

----

### Before the change?

Bodies were not of type string on Vercel, which broke Probot.

### After the change?

Bodies are now read as string.

### Pull request checklist

(Tests will be added later)

- [ ] Yes
- [x] No

----

### DIFF
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -11,7 +11,16 @@ import AggregateError from "aggregate-error";
 // }
 type IncomingMessage = any;
 
-export function getPayload(request: IncomingMessage): Promise<string> {
+
+async function buffer(readable: Readable) {
+  const chunks = [];
+  for await (const chunk of readable) {
+    chunks.push(typeof chunk === 'string' ? Buffer.from(chunk) : chunk);
+  }
+  return Buffer.concat(chunks);
+}
+
+export async function getPayload(request: IncomingMessage): Promise<string> {
   if ("body" in request) {
     if (
       typeof request.body === "object" &&
@@ -19,10 +28,13 @@ export function getPayload(request: IncomingMessage): Promise<string> {
       request.rawBody instanceof Buffer
     ) {
       // The body is already an Object and rawBody is a Buffer (e.g. GCF)
-      return Promise.resolve(request.rawBody.toString("utf8"));
+      return request.rawBody.toString("utf8");
+    } else if(Symbol.asyncIterator in request || Symbol.iterator in request) {
+      const buf = await buffer(req);
+      return buf.toString('utf8');
     } else {
       // The body is a String (e.g. Lambda)
-      return Promise.resolve(request.body);
+      return request.body;
     }
   }
 
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -29,6 +29,7 @@ export async function getPayload(request: IncomingMessage): Promise<string> {
     ) {
       // The body is already an Object and rawBody is a Buffer (e.g. GCF)
       return request.rawBody.toString("utf8");
+    // request is a readable stream (e.g. vercel)
     } else if(Symbol.asyncIterator in request || Symbol.iterator in request) {
       const buf = await buffer(req);
       return buf.toString('utf8');


src/middleware/node/get-payload.ts
```diff
@@ -11,18 +11,30 @@ import AggregateError from "aggregate-error";
...
   if ("body" in request) {
     if (
       typeof request.body === "object" &&
       "rawBody" in request &&
       request.rawBody instanceof Buffer
     ) {
       // The body is already an Object and rawBody is a Buffer (e.g. GCF)
-      return Promise.resolve(request.rawBody.toString("utf8"));
+      return request.rawBody.toString("utf8");
+    } else if(Symbol.asyncIterator in request || Symbol.iterator in request) {
```
Uzlopak, line 33:
```suggestion
    // request is a readable stream (e.g. vercel)
    } else if(Symbol.asyncIterator in request || Symbol.iterator in request) {
```

Is it enough to check for the Symbols to assume it is a readable stream?

G-Rath (changes requested):
I'm on the fence about this because it adds a sizable function and we can't expect to account for every behaviour.

If I understand correctly, downstream users should be able to handle this and assign the result to `rawBody` before passing the request to us - so instead we could just document that and even include a test to ensure that works instead.

(Either way though, you need to write tests)

src/middleware/node/get-payload.ts
AaronDewes: This can handle some other cases where request can be an iterator over either strings or buffers. Whatever is behind it does not really matter.

vedantmgoyal2009:
> If I understand correctly, downstream users should be able to handle this and assign the result to `rawBody` before passing the request to us - so instead we could just document that and even include a test to ensure that works instead.

How can I do that? Vercel doesn't support disabling request body parsing for a specific function, and I can't sacrifice either Probot or my other functions.

G-Rath:
> How can I do that? Vercel doesn't support disabling request body parsing for a specific function

Yes it does, via the `api: { bodyParser: boolean }` option - see https://vercel.com/guides/how-do-i-get-the-raw-body-of-a-serverless-function and https://github.com/vercel/next.js/discussions/44212.

Note that that would be required in order for this to work as well.

src/middleware/node/get-payload.ts
G-Rath: > This can handle some other cases where request can be an iterator over either strings or buffers

What other cases? This is the first time I know of that this has been raised, and this is a lot more work than just checking for `rawBody`.

vedantmgoyal2009:
Okay, I tested it, and it turns out that it doesn't work because I'm using Probot, and not this SDK directly. It might be really good to merge #962 or #963. Although, I would prefer #962 since it's more direct and does not include a "sizable function". Many cloud frameworks do provide this feature, so I don't see any harm in supporting the case where the request body is pre-parsed.

G-Rath:
You should be able to since Probot doesn't directly run in Vercel i.e. you've got _some_ custom code that starts up Probot, but you're probably not having to do anything with the request already so it could be a lot of work to handle this.

However, I think it would be better if Probot supported providing a form of "request preprocessor" (if it doesn't already) instead of us adding more and more platform-specific workarounds here; Probot would then also be free to ship as many platform specific preprocessors it likes.

AaronDewes:
> You should be able to since Probot doesn't directly run in Vercel i.e. you've got _some_ custom code that starts up Probot, but you're probably not having to do anything with the request already so it could be a lot of work to handle this.
> 
> However, I think it would be better if Probot supported providing a form of "request preprocessor" (if it doesn't already) instead of us adding more and more platform-specific workarounds here; Probot would then also be free to ship as many platform specific preprocessors it likes.

We have some of them for GitHub actions, Azure Functions and AWS Lambda: https://github.com/probot/?q=adapter&type=all&language=&sort=.

@gr2m Can you create a new repo probot/adapter-vercel so we can close this PR please?

vedantmgoyal2009:
Are we having a whole new project setup for just doing `JSON.stringify()`?

Uzlopak:
@vedantmgoyal2009 
Could you be less arrogant?

AaronDewes:
> Are we having a whole new project setup for just doing `JSON.stringify()`?

Doing JSON.stringify is not safe and can break, as I explained here: https://github.com/octokit/webhooks.js/pull/962#discussion_r1473821735.

Also, we should not try to bother Octokit maintainers too much with Probot problems we can solve ourselves.

AaronDewes:
Closing this now. @gr2m Before you create a new adapter repo, I'll try something else in Probot first and will let you know if my other idea won't work.

PR #967, opened 2024-02-11
feat: v13

### LINKED ISSUE #271: Restore aggregate-error external dependency

wolfy1339:
BREAKING CHANGE: Use GitHub's OpenAPI spec
BREAKING CHANGE: Output ESM instead of CJS

Fixes #271
Fixes #897
Fixes #974


gr2m (approved):
outstanding work!

PR #968, opened 2024-02-11
feat: migrate to ESM

wolfy1339:
BREAKING CHANGE: Migrate to ESM builds

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@octokit/webhooks",
   "version": "0.0.0-development",
+  "type": "module",
   "description": "GitHub webhook events toolset for Node.js",
   "keywords": [],
   "repository": "github:octokit/webhooks.js",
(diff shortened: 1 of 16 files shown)


tsconfig.json
```diff
@@ -8,6 +8,9 @@
     }
   },
   "compilerOptions": {
+    "lib": ["ESNext"],
```
wolfy1339, line 10: These can be moved to `@octokit/tsconfig` later on

package.json
```diff
@@ -43,14 +44,13 @@
     "@octokit/request-error": "^5.0.0",
     "@octokit/webhooks-methods": "^4.0.0",
     "@wolfy1339/openapi-webhooks-types": "5.1.3",
-    "aggregate-error": "^3.1.0"
+    "aggregate-error": "^5.0.0"
```
wolfy1339, line 47: While `AggregateError` was introduced natively in NodeJS v15, it isn't generic. So the types cannot be extended to use our custom error

src/types.ts
```diff
@@ -75,20 +76,3 @@ export interface WebhookEventHandlerError<TTransformed = unknown>
     ? EmitterWebhookEvent
     : EmitterWebhookEvent & TTransformed;
 }
-
```
wolfy1339, line 78: The types were fixed in version 3.0.1 and these are out of date versus v5 of `aggregate-error`

test/integration/event-handler-test.ts
```diff
@@ -91,12 +91,12 @@ describe("when a handler throws an error", () => {
     } catch (error: any) {
       expect(error.message).toMatch(/oops/);
 
-      const errors = Array.from(error);
+      const errors = Array.from(error.errors);
```
wolfy1339, line 94: The `AggregateError` class doesn't extend `Iterable<T>` anymore, you have to use the `.errors` property

nickfloyd (approved):
All of this looks good @wolfy1339 - thank you for pushing this forward ❤️.  Just a clarifying question  - are you planning a follow along set of changes in octokit.js and app.js for imports and such since those are still cjs modules?

wolfy1339:
Yes, I will be doing the other modules as well.

PR #1010, opened 2024-04-27
fix: explicitly check that `request.body` is a string

### LINKED ISSUE #1009: [BUG]: webhooks.js v12.0.10 onward breaks compatibility with @fastify/middie middleware library
### What happened?

Due to a conditional change within `getPayload()` from `if(request.body)` to `if("body" in request)` compatibility with the middleware library `@fastify/meddie` library is broken. Although this may be an upstream issue with how Middie parses the request, this is an easy fix that was an existing behaviour prior to v12.0.10[0].

If `request.body` is undefined, Middie parses this as a defined key with undefined value as apposed to not indulging the body property like `node:http` and `express` do.

I've only inspected middleware request `body` handling with `node:http`, `express`, `@fastify/express` and `@fastify/middie` but this may impact other libraries other than `@fastify/middie`.

[0] https://github.com/octokit/webhooks.js/compare/v12.0.9...v12.0.10

### Versions

webhooks.js 13.2.6

### Relevant log output

_No response_

### Code of Conduct

----
UPDATE: This may be being fixed upstream within `@fastify/middie`, if attempting to reproduce this bug please use a Middie <=8.3.0. I think the `@octokit/webhooks` patch should still be completed as this could be quite difficult to diagnose if the user is on a <=8.3.0 of Middie or if they have constraints for upgrading it.

didley:
Resolves #1009 

----

### Before the change?

* I am reverting a conditional that changes the handling of `request.body` that was made within v12.0.10. This broke webhooks for libraries that define the request `body` key, if `body` is undefined such as `@fastify/middie` rather than not including the `body` property like `node:http` and `express`.

### After the change?

* `getPayload()` will resolve with a string rather than the undefined if the request body key is defined but has a value undefined.

### Pull request checklist

- [ ] Yes
- [x] No

----

UPDATE: This may be being fixed upstream within `@fastify/middie`, if attempting to reproduce this bug please use a Middie <=8.3.0. I think the `@octokit/webhooks` patch should still be completed as this could be quite difficult to diagnose if the user is on a <=8.3.0 of Middie or if they have constraints for upgrading it.

### DIFF
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -12,7 +12,7 @@ import AggregateError from "aggregate-error";
 type IncomingMessage = any;
 
 export function getPayload(request: IncomingMessage): Promise<string> {
-  if ("body" in request) {
+  if (request.body) {
     if (
       typeof request.body === "object" &&
       "rawBody" in request &&
--- a/src/middleware/node/get-payload.ts
+++ b/src/middleware/node/get-payload.ts
@@ -12,18 +12,16 @@ import AggregateError from "aggregate-error";
 type IncomingMessage = any;
 
 export function getPayload(request: IncomingMessage): Promise<string> {
-  if (request.body) {
-    if (
-      typeof request.body === "object" &&
-      "rawBody" in request &&
-      request.rawBody instanceof Buffer
-    ) {
-      // The body is already an Object and rawBody is a Buffer (e.g. GCF)
-      return Promise.resolve(request.rawBody.toString("utf8"));
-    } else {
-      // The body is a String (e.g. Lambda)
-      return Promise.resolve(request.body);
-    }
+  if (
+    typeof request.body === "object" &&
+    "rawBody" in request &&
+    request.rawBody instanceof Buffer
+  ) {
+    // The body is already an Object and rawBody is a Buffer (e.g. GCF)
+    return Promise.resolve(request.rawBody.toString("utf8"));
+  } else if (typeof request.body === "string") {
+    // The body is a String (e.g. Lambda)
+    return Promise.resolve(request.body);
   }
 
   // We need to load the payload from the request (normal case of Node.js server)
--- a/test/integration/get-payload.test.ts
+++ b/test/integration/get-payload.test.ts
@@ -74,4 +74,19 @@ describe("getPayload", () => {
 
     expect(await promise).toEqual("foo");
   });
+
+  it("resolves with a string if the body key of the request is defined but value is undefined", async () => {
+    const request = new EventEmitter();
+    // @ts-ignore body is not part of EventEmitter, which we are using
+    // to mock the request object
+    request.body = undefined;
+
+    const promise = getPayload(request);
+
+    // we emit data, to ensure that the body attribute is preferred
+    request.emit("data", "bar");
+    request.emit("end");
+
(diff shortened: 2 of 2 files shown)


wolfy1339:
Is it possible to test that this won't break in a future release

src/middleware/node/get-payload.ts
```diff
@@ -12,7 +12,7 @@ import AggregateError from "aggregate-error";
 type IncomingMessage = any;
 
 export function getPayload(request: IncomingMessage): Promise<string> {
-  if ("body" in request) {
+  if (request.body) {
```
G-Rath: @wolfy1339 I feel like we should probably remove this check entirely, and change our `else` branch to be `else if typeof 'string'`, as this change means an empty string won't get treated as a string anymore - while in practice I doubt it'll break anything, I think we should ensure we have a clear official logic which I think should be "if object, we assume; if string, we assume" (i.e. regardless of actual value)

what do you think?

didley: Thanks Gareth, yes that sounds like it could be a good approach, although I've just briefly tested that this morning and it breaks quite a few unit tests. I can investigate it further this afternoon.

didley:
@wolfy1339 I have added a unit test covering this case. Is there anything else you had in mind to ensue this won't break a future release?

src/middleware/node/get-payload.ts
didley: After further investigation  `if (request.body === undefined || typeof request.body === "string") // parsing block` gets the unit tests close to passing, but unfortunately I don't think is the right approach. 

The existence of a body indicates it has been parsed even if it was a `string`. Many of the unit tests make this assumption[0]. This is also how `node:http` treats the body, the `body` key isn't created unless by a body parser or external logic. 

The crux of the issue is `@fastify/middie` attaches the `body` key even if it's `undefined`. I think `if (request.body !== undefined)` will be a good approach, it will solve the case for Middie while also aligning the handling of `body` with `node:http`, and allows for the `null` and `''` case.

[0] such as `resolves with the body, if passed via the request` within get-payload.test.ts.

G-Rath: I'm a bit confused by this - just to be sure, when you say "... gets the unit tests close to passing", you are accounting for the nested conditions right? because otherwise yes the tests will fail since you'll be disabling the "is the request.body already a parsed body?" flow.

What I'm getting at is our logic is currently this:

```
if ('body' in request) {
  if (<something about body>) {
    return ...;
  } else {
    return ...;
  }
}
```

and with your change, it strikes me that we should be able to do this instead:

```
if(<something about the body>) {
  return ...;
} else if(typeof request.body === 'string') {
  return ...;
}
```

From what I'm seeing that should be equivalent because we explicitly type the return of this function as a `string` and our comment says it should be a `string` even though we're technically not checking it is a `string` - if what you're saying is that we actually are sometimes _not_ returning a `string` due to that lack of check, then that's a whole other concerning thing.

didley: Apologies, I misunderstood with my initial attempt. I've just pushed a change that I think you're meaning. Yes this works well, it fixes the issue with the compatibility with `@fastify/middie` and all unit tests are passing as expected.

G-Rath (approved):

didley:
Thank you @G-Rath and everyone else involved!

PR #1036, opened 2024-07-14
chore: remove unused deps

Uzlopak:


package.json
```diff
@@ -50,12 +50,10 @@
     "@octokit/tsconfig": "^3.0.0",
     "@types/node": "^20.0.0",
     "@vitest/coverage-v8": "^2.0.2",
-    "axios": "^1.0.0",
     "esbuild": "^0.23.0",
     "express": "^4.17.1",
     "glob": "^11.0.0",
     "prettier": "3.3.3",
-    "prettier-plugin-packagejson": "^2.2.9",
```
wolfy1339: This is used by the CLI in the prettier workflow. you won't see it used in any scripts

Uzlopak: according to the documentation we never configured it accordingly

https://github.com/matzkoh/prettier-plugin-packagejson/blob/master/README.md#installation

Uzlopak: but if you are 100% sure that this plugin is in use i wll of course revert

wolfy1339: it was added in https://github.com/octokit/webhooks.js/pull/412

wolfy1339: You are correct that as it is, it is currently unused, due to Prettier v3.

The configuration should then be updated to add the plugin

Uzlopak: Done :)

wolfy1339 (approved):

PR #1038, opened 2024-07-15
feat: use native AbortError

### LINKED ISSUE #1013: [MAINT]: drop aggregate-error dependency
### Describe the need

The project's tsconfig targets `es2023` and the stated node support is (at least?) [node>16](https://github.com/octokit/webhooks.js/releases/tag/v12.0.0), but it still takes a dependency on [aggregate-error](https://www.npmjs.com/package/aggregate-error), which is made redundant by [a standard global for ~4 years now](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError#browser_compatibility), introduced in versions older than these.

This dependency caused me trouble in an arcane monorepo rollup build that is mostly self-inflicted and probably experienced by nobody else, so I won't elaborate on the details.  Instead I'll just appeal to the general principle that unneeded dependencies are bad 🌞.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

Uzlopak:
fixes #1013

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,8 +11,7 @@
       "dependencies": {
         "@octokit/openapi-webhooks-types": "8.3.0",
         "@octokit/request-error": "^6.0.1",
-        "@octokit/webhooks-methods": "^5.0.0",
-        "aggregate-error": "^5.0.0"
+        "@octokit/webhooks-methods": "^5.0.0"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


Uzlopak:
@isker 

PTAL

wolfy1339 (approved):

Issue #1055, opened 2024-09-20
[BUG]: Public interface types should expect strings


jyasskin:
### What happened?

I tried to call `verifyAndReceive` with `name`=`request.headers["x-github-event"]`, and got a type error because the header's type is `string`, but `verifyAndReceive` expects `WebhookEventName`, which isn't even exported from the library. It turns out that https://github.com/octokit/webhooks.js/blob/6531c971903ac2cc41bd2bd5b8c40ac7741002d5/src/middleware/node/middleware.ts#L78 is casting without checking (and then https://github.com/octokit/webhooks.js/blob/6531c971903ac2cc41bd2bd5b8c40ac7741002d5/src/middleware/node/middleware.ts#L98 unnecessarily casts to `any` on top of that), which indicates that `verifyAndReceive` should probably just take a `string`.

Alternatively, the library could expose a function to validate the event name and maybe payload structure, but that seems like more work than just loosening up the types.

A secondary question is whether `receive` should also take looser types for `name` and `payload`. https://github.com/octokit/webhooks.js/blob/6531c971903ac2cc41bd2bd5b8c40ac7741002d5/src/verify-and-receive.ts#L41-L45 casts without checking, which seems to indicate that it should, but doing that takes more surgery on `BaseWebhookEvent` to make it correctly infer the payload type when the event name does happen to be constrained.

### Versions

Typescript 5.6

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

wolfy1339:
Here's an example of a validation function for event names
```ts
import { emitterEventNames } from "@octokit/webhooks";
import type { EmitterWebhookEventName } from "@octokit/webhooks/types";

export function validateEventName(event: string | string[]): event is EmitterWebhookEventName | EmitterWebhookEventName[] {
  if (Array.isArray(event)) {
    return event.every(e => emitterEventNames.includes(e as EmitterWebhookEventName));
  }
  return emitterEventNames.includes(event as EmitterWebhookEventName);
}

```

wolfy1339:
Here's a set of functions to check if the payload is of proper structure, checks if the event name is a known event. Both functions resolve the types.

```ts
import { emitterEventNames } from "@octokit/webhooks";
import type { EmitterWebhookEventName, EmitterWebhookEvent } from "@octokit/webhooks/types";

/**
 * Validates if the provided event name or names are a know event name.
 *
 * @param {string | string[]} event The event name or an array of event names to validate.
[...]

PR #1056, opened 2024-09-20
fix: make verifyAndReceive accept a string name.

jyasskin:
As naturally comes from the relevant HTTP header.

Resolves the easy half of #1055

----

### Before the change?

* Callers have to do an unchecked cast from `string` to either `any` or the unexported `WebhookEventName` type.

### After the change?

* Callers can pass the HTTP header directly into `verifyAndReceive()`

### Pull request checklist
Kinda: I removed the cast from the middleware, so it'll break if the stricter type is re-introduced.
This type constraint doesn't seem to have been documented.

- [ ] Yes
- [x] No

----

### DIFF
--- a/src/middleware/node/middleware.ts
+++ b/src/middleware/node/middleware.ts
@@ -2,8 +2,6 @@
 // see https://github.com/octokit/octokit.js/issues/2075#issuecomment-817361886
 import type { IncomingMessage, ServerResponse } from "node:http";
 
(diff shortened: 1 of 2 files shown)


wolfy1339:
The `WebhookEventName` is actually exported in the `types.ts` file.

```ts
import type { WebhookEventName } from "@octokit/webhooks/types";
```

**Note**: there is no file extension on the import path

wolfy1339 (approved):

PR #1080, opened 2024-11-29
Use verifyWithFallback to support secret rotation

### LINKED ISSUE #770: [FEAT]: Facilitate webhook secret rotation

bismip:
Resolves #770 

----

### Before the change?

Only one secret is supported

### After the change?

Multiple fallback secrets are supported

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -54,7 +54,7 @@ class Webhooks<TTransformed = unknown> {
 
     const state: State & {
       secret: string;
-      additionalSecrets?: undefined | string[];
+      additionalSecrets?: string[] | undefined
       eventHandler: EventHandler<TTransformed>;
     } = {
       eventHandler: createEventHandler(options),
--- a/src/index.ts
+++ b/src/index.ts
@@ -54,10 +54,12 @@ class Webhooks<TTransformed = unknown> {
 
     const state: State & {
       secret: string;
+      additionalSecrets?: undefined | string[];
       eventHandler: EventHandler<TTransformed>;
     } = {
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -54,10 +54,12 @@ class Webhooks<TTransformed = unknown> {
 
     const state: State & {
       secret: string;
+      additionalSecrets?: undefined | string[];
```
wolfy1339:
```suggestion
      additionalSecrets?: string[] | undefined
```

src/types.ts
```diff
@@ -38,6 +38,7 @@ interface BaseWebhookEvent<TName extends WebhookEventName> {
 
 export interface Options<TTransformed = unknown> {
   secret?: string;
+  additionalSecrets?: undefined | string[];
```
wolfy1339:
```suggestion
  additionalSecrets?: string[] | undefined;
```

wolfy1339 (commented):
Just a little style convention, otherwise it looks good

bismip:
@wolfy1339 Thank you, fixed

wolfy1339 (dismissed):

bismip:
@wolfy1339 I trusted your suggestions 😆 Added the missing semicolon, should pass now.

wolfy1339 (approved):

PR #1089, opened 2025-01-08
Update `createNodeMiddleware()` documentation

arnitdo:
Update `createNodeMiddleware()` - `path` option documentation, to reflect exact setting of value for custom routing options

----

### Before the change?

* Documentation does not *fully* reflect the correct behavior of `path` in `createNodeMiddleware()`

### After the change?

* Documentation updated, informing users that setting the `path` value to exactly `"/"` enables the created middleware to be used as a handler

### Pull request checklist
- [x] Tests for the changes have been added (for bug fixes / features) *[No Changes]*

- [ ] Yes
- [x] No

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -551,7 +551,7 @@ The middleware returned from `createNodeMiddleware` can also serve as an
         </em>
       </td>
       <td>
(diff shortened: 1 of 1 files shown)


arnitdo:
Had to dig into the source code for this exact behavior, since not everyone is running `http.createServer()` usually - the generated middleware can technically be used with `Express.js` as well

PR #932, opened 2023-11-25
fix: export createNodeHandler

Uzlopak:
Resolves #930 

I would like to have the middleware separated from the handler to use it in probot. 

Resolves #ISSUE_NUMBER

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #1094, opened 2025-01-28
fix: improve header validation

Uzlopak:
takes the header validation logic from #932


G-Rath:
This doesn't really seem like an improvement to me as we're now only reporting a single missing header, and using more code to do so?

Uzlopak:
@G-Rath 

The benefit is to avoid creating arrays and strings, over and over.

Uzlopak:
@G-Rath 

What about now? It avoids as I wrote unneccessary array and string generation.

wolfy1339:
I don't think we need to use binary here. Let's keep this simple....

Uzlopak:
I actually want more performance.

wolfy1339:
While more performance is always good. At what point do we really need to sacrifice readability and understanding for performance.
Let's keep it stupid simple

PR #1096, opened 2025-01-30
No bind

Uzlopak:
Avoid the unnecessary binding


wolfy1339:
This is a breaking change, do we really need to do it like that

Uzlopak:
Closing for now, because I have something else also in mind.

PR #1115, opened 2025-02-24
maint: set node versions to `lts/*`, replace `ts-node` with `tsx`

wolfy1339:
Resolves #ISSUE_NUMBER

`ts-node` is not very well maintained currently and needs workarounds for ESM usage.
`tsx` just works out of the box

----

### Before the change?

* Workflows would use pinned versions of Node, while they weren't dependant on any specific version
* Used `ts-node` to run scripts

### After the change?

* Workflows will now use the latest lts release of Node
* Use `tsx` to run script

### Pull request checklist

- [ ] Yes
- [ ] No

----

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           cache: npm
-          node-version: 20
+          node-version: lts/*
       - run: npm ci
       - run: npm run validate:ts
     if: ${{ always() }}
--- a/.github/workflows/update-prettier.yml
+++ b/.github/workflows/update-prettier.yml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           cache: npm
-          node-version: 20
+          node-version: lts/*
       - run: npm ci
       - run: npm run lint:fix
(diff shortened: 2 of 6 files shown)


nickfloyd (dismissed):
❤️

.github/workflows/test.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           cache: npm
-          node-version: 20
+          node-version: lts/*
```
oscard0m, line 38: Do we want to rely on `lts/*` and not control when we are ready to upgrade node?

wolfy1339: This test isn't reliant on any particular Node version, and using `lts/*` avoids the need to change it.
It's simply validating the types.

oscard0m: Could we potentially get different `tsc` results depending on the node version we use for `npm run validate:ts`. Do we want to run this with a matrix of node versions?

Approving in the meantime as this should be a different issue (in case we want).

oscard0m (approved):

.github/workflows/test.yml
wolfy1339: Typescript is supposed to give the same output no matter the NodeJS version

PR #1143, opened 2025-05-09
refactor: write tests to also work with deno and bun

Uzlopak:
My first goal was to ensure, that the "webmiddleware" is tested properly.  But then I realized, that the web middleware is kind of better than the old node logic. So that was used as base for the unification.

So anyhow. I tried first tried to get it work with Deno. So tried to remove node specific code as much as possible. So instead of using Buffer I used Uint8Array.

I basically needed to rewrite all the unit tests. Why keeping vitest? `node --test` works only reliable with node 24.  But I think it is now very cool, as the tests run additionally on deno and bun.


.github/workflows/test.yml
```diff
@@ -38,15 +40,34 @@ jobs:
...
+
+  test-bun:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: oven-sh/setup-bun@v2
+      - run: bun install
+      - run: bun test
+            
+  validate-typescript:
```
wolfy1339, line 62: Because this job was renamed, it will require the intervention of an admin to merge this

Uzlopak: is it better now?

wolfy1339: yeah, that works

test/helpers/find-free-port.ts
```diff
@@ -0,0 +1,40 @@
...
+  while (tries < maxTries) {
+    if (await isFreePort(port)) {
+      return port;
+    }
+    port = getRandomPort();
+
+    tries++;
+  }
+  throw new Error("No available ports");
+}
```
gr2m: we can add https://github.com/sindresorhus/get-port as a devdependency instead of adding this code

src/middleware/node/get-payload.ts
```diff
@@ -36,8 +36,7 @@ export function getPayloadFromRequestStream(
     request.on("data", data.push.bind(data));
     request.on("end", () => {
       const result = concatUint8Array(data);
-      // Switch to queue microtask when we want to support bun and deno
-      setImmediate(resolve, result);
+      queueMicrotask(() => resolve(result));
```
gr2m, line 39: TIL, fascinating https://developer.mozilla.org/en-US/docs/Web/API/Window/queueMicrotask

gr2m (dismissed):
Looks good to me, just a nit

test/helpers/find-free-port.ts
Uzlopak: we could but the problem is, that we run into race conditions.

gr2m: I don't see how but I won't die on my hill 😁

Uzlopak: nevermind

Uzlopak: :D

gr2m (approved):
looks good, just give the PR a proper title explaining what you've done. You can use `refactor: ...` as prefix.

Uzlopak:
@gr2m acceptable?

PR #1144, opened 2025-05-10
Use ts imports in src

Uzlopak:
Splitting up my refactor PR into multiple PRs so that it can be easier reviewed.

This PR changes the imports to .ts files in the src folder. 

With this, you can also test the code in deno. Also it seems that --experimental-strip-types can work better with that on node24.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,20 @@ name: Test
       - opened
       - synchronize
 jobs:
+  test-building:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          cache: npm
+          node-version: lts/*
+      - run: npm ci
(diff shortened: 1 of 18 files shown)


package-lock.json
wolfy1339, line 1: I don't think this is supposed to be updated here

wolfy1339 (dismissed):

package-lock.json
Uzlopak: Yes, looks strange indeed. I will check it when I am back home. I guess i was using node 18 npm and it removed the integrity and resolved fields

Uzlopak: Seems like it was connected with https://stackoverflow.com/questions/75224019/resolved-and-integrity-go-missing-in-package-lock-json as I had to cleanthe cache to generate a proper package-lock.json.  But the change in package-lock.json was not necessary so I reverted it ;)

wolfy1339 (approved):

PR #1146, opened 2025-05-10
fix: use `Uint8Array` instead of `Buffer`

Uzlopak:
We should avoid the use of Buffer and instead use Uint8Array.


gr2m (approved):
> We should avoid the use of Buffer and instead use Uint8Array.

I agree.

But does this introduce a breaking change to the end user?

Uzlopak:
No, the unit test was wrong in the first place, because it passes a string, despite that the streams should use Buffers/Uint8Arrays anyway. It was just working, because somewhere the string got transformed to Buffer. If I would have written that unit test long time ago correct, it would not triggered that failing unit test today. :D

test/integration/get-payload.test.ts
```diff
@@ -85,9 +85,20 @@ describe("getPayload", () => {
     const promise = getPayload(request);
 
     // we emit data, to ensure that the body attribute is preferred
-    request.emit("data", "bar");
+    request.emit("data", Buffer.from("bar"));
```
Uzlopak, line 88: It was working, because of 
https://github.com/octokit/webhooks.js/blob/9a9593a4325a0e4bc3e05409382c9bf3cd80ddd7/src/middleware/node/get-payload.ts#L38

So the string "bar" would be then "bar".toString()

I guess in the moment where we emit two times strings, it would error because Buffer.concat expects only Uint8Array and Buffers.

gr2m:
Okay I say we ship it, thanks for all the great work @Uzlopak! 

But I suggest we do a release even though this is technically a refactor as far as I can tell. Or we do a feature release saying it introduces compatibility with `Uint8Array`? I'm not sure what the better option is, but I would like us to do a release just in case the changes cause problems in some obscure JS runtime environment, it will make it easier to track down the problem and to revert the change if necessary.

Uzlopak:
@gr2m 

You mean this change as last minor of the current major?

Uzlopak:
@gr2m 

Sure, makes sense.

I wrote following test case in main branch.

```typescript
  it("resolves with a string if the body key of the request is defined but value is undefined", async () => {
    const request = new EventEmitter();
    // @ts-ignore body is not part of EventEmitter, which we are using
    // to mock the request object
    request.body = undefined;

    const promise = getPayload(request);

    // we emit data, to ensure that the body attribute is preferred
    request.emit("data", "foo");
    request.emit("data", "bar");
    request.emit("end");

    expect(await promise).toEqual("foobar");
  });
```

And I got following result:

```sh
 FAIL  test/integration/get-payload.test.ts > getPayload > resolves with a string if the body key of the request is defined but value is undefined
TypeError: The "list[0]" argument must be an instance of Buffer or Uint8Array. Received type string ('foo')
 ❯ EventEmitter.<anonymous> src/middleware/node/get-payload.ts:39:20
     37|         data.length === 1
     38|           ? data[0].toString("utf8")
     39|           : Buffer.concat(data).toString("utf8"),
       |                    ^
     40|       ),
     41|     );
 ❯ test/integration/get-payload.test.ts:105:13
```

So definetly the test was wrong by passing a string.

gr2m (approved):

PR #1147, opened 2025-05-10
build: remove `glob`

### LINKED ISSUE #862: [FEAT]: Export type MiddlewareOptions
### Describe the need

I need a little wrapper around `createNodeMiddleware` and for that use case, it would be nice if the `MiddlewareOptions` type were exported explicitly from the main module.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

Uzlopak:
Remove glob

### DIFF
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -2,7 +2,6 @@ import esbuild from "esbuild";
 import { dirname } from "node:path";
 import { mkdirSync, writeFileSync } from "node:fs";
 import { copyFile, readFile, writeFile, rm } from "node:fs/promises";
-import { glob } from "glob";
 
 /**
  * Internal function to rewrite relative import extensions.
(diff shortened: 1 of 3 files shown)


G-Rath:
Can you split this up, since one is a build change and the other is a feature

Uzlopak:
G-Rath

I removed in this PR glob. In a follow up we can export MiddlewareOptions

wolfy1339 (approved):

scripts/build.mjs
```diff
@@ -108,13 +107,8 @@ async function main() {
   });
 
   // Remove the types file from the dist-src folder
-  const typeFiles = await glob([
-    "./pkg/dist-src/**/types.js.map",
```
G-Rath, line 112: Is this file not being generated anymore?

wolfy1339: Nope, map files haven't been generated for the source files
https://github.com/octokit/webhooks.js/blob/main/scripts%2Fbuild.mjs#L106

G-Rath (approved):
Love removing a dependency like this!

PR #1148, opened 2025-05-11
feat: allow to specify `timeout` option in middleware, refactor middleware

Uzlopak:
This PR is builds on #1146.

As you can see, the unit test pass. 

In a follow up PR we can adapt the new tests I wrote, and then we can test for deno and bun ;)


Uzlopak:
@gr2m 

Sure, makes sense.

I wrote following test case in main branch.

```typescript
  it("resolves with a string if the body key of the request is defined but value is undefined", async () => {
    const request = new EventEmitter();
    // @ts-ignore body is not part of EventEmitter, which we are using
    // to mock the request object
    request.body = undefined;

    const promise = getPayload(request);

    // we emit data, to ensure that the body attribute is preferred
    request.emit("data", "foo");
    request.emit("data", "bar");
    request.emit("end");

    expect(await promise).toEqual("foobar");
  });
```

And I got following result:

```sh
 FAIL  test/integration/get-payload.test.ts > getPayload > resolves with a string if the body key of the request is defined but value is undefined
TypeError: The "list[0]" argument must be an instance of Buffer or Uint8Array. Received type string ('foo')
 ❯ EventEmitter.<anonymous> src/middleware/node/get-payload.ts:39:20
     37|         data.length === 1
     38|           ? data[0].toString("utf8")
     39|           : Buffer.concat(data).toString("utf8"),
       |                    ^
     40|       ),
     41|     );
 ❯ test/integration/get-payload.test.ts:105:13
```

So definetly the test was wrong by passing a string.

Uzlopak:
@wolfy1339 
@gr2m 

Is this now acceptable?

gr2m (approved):
Looks good to me. I would suggest to do a new release due to the significance of the changes. It looks like we should make this a feature release since `timeout` is now an argument for the Node.js middleware. Should we document it at https://github.com/octokit/webhooks.js?tab=readme-ov-file#createnodemiddleware, or is that just exposed for testing?

Uzlopak:
THe timeout is not node specific.for  Web is also possible to configure a timeout. 

I created it for testing. I dont think people need to configure the timeout.

Uzlopak:
Should we merge this first and then the tests or how do we want to proceed?

PR #1149, opened 2025-05-12
ci: stop testing against NodeJS v18

wolfy1339:
BREAKING CHANGE: Drop support for NodeJS v18


Uzlopak:
why 20.17?

wolfy1339:
It's just an error.
If anything, it should be `20.19` and not `20.17`, for `require(esm)`

oscard0m (dismissed):

PR #1157, opened 2025-06-04
feat: add new `verifyAndParse()` method

### LINKED ISSUE #379: `webhooks.verifyAndParse` and standalone `verifyAndParse` method
**What’s missing?**

There is currently no simple way to parse and verify an incoming http request object at the same time. It's only possible as part of the event API of a `webhooks` instance, which might be overkill / not the best fit for a serverless/function environment

This would permit for code such as this:

```ts
export async function handler(event) {
  try {    
    const githubEvent = await verifyAndParse(secret, event)
    // handle valid event
    return { body: text, statusCode: 200 };
  } catch (error) {
    // handle errors
    return { body: 'oh noes!', statusCode: 500 }
  }
}
```

**Why?**

This is a follow up to https://github.com/octokit/webhooks.js/pull/372#issuecomment-739575379, see the discussion for more reasoning.

In a nutshell, @G-Rath's use case is to separate receiving and verifying the event request from running any event handlers to avoid AWS Lambda timeouts.

**Alternatives you tried**

```js
export async function handler(event) {
  const webhooks = new Webhooks(options)

  webhooks.on(['check_run', 'code_scanning_alert', 'commit_comment'], async (event) => {
    // handle valid events
  })

  try {
    await webhooks.verifyAndReceive(awsEventToOctokitEvent(event))
    return { statusCode: 200 };
  } catch (error) {
    // handle error
    return { body: 'oh noes!', statusCode: 500 }
  }
}
```

wolfy1339:
This method is almost identical to `verifyAndReceive()`, the difference being that this function won't trigger the event handler, so it can be executed at a later time.

The code has been refactored so that this function is used internally to remove duplication.

Fixes #379

----

### Before the change?

* Users could not simply get an `EmitterWebhookEvent` object, the included functions would fire the event handler for webhook events

### After the change?

* Users can now get an `EmitterWebhookEvent` object without the event handler being utilised. This is especially useful for serverless platforms

### Pull request checklist

- [ ] Yes
- [ ] No

----

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,6 +16,7 @@ import type {
   WebhookEventHandlerError,
   EmitterWebhookEventWithStringPayloadAndSignature,
 } from "./types.ts";
+import { verifyAndParse } from "./verify-and-parse.ts";
 
 export { createNodeMiddleware } from "./middleware/node/index.ts";
 export { createWebMiddleware } from "./middleware/web/index.ts";
@@ -47,6 +48,9 @@ class Webhooks<TTransformed = unknown> {
   public verifyAndReceive: (
     options: EmitterWebhookEventWithStringPayloadAndSignature,
   ) => Promise<void>;
+  verifyAndParse: (
+    event: EmitterWebhookEventWithStringPayloadAndSignature,
+  ) => Promise<EmitterWebhookEvent | WebhookError>;
 
   constructor(options: Options<TTransformed> & { secret: string }) {
     if (!options || !options.secret) {
@@ -73,6 +77,15 @@ class Webhooks<TTransformed = unknown> {
     this.removeListener = state.eventHandler.removeListener;
     this.receive = state.eventHandler.receive;
     this.verifyAndReceive = verifyAndReceive.bind(null, state);
+    this.verifyAndParse = async (
+      event: EmitterWebhookEventWithStringPayloadAndSignature,
+    ) => {
+      return verifyAndParse(
+        state.secret,
+        event,
+        state.additionalSecrets,
+      ) as Promise<EmitterWebhookEvent | WebhookError>;
+    };
   }
 }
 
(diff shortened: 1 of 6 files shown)


G-Rath:
wow blast from the past! I'd forgotten about this, which hopefully shows we've all done good work in our code since I do still have lambdas consuming GH events that I've obviously not had to think about in 5 years 😄

I'm not sure though how useful this actually is especially as it doesn't seem like there's been a lot of request for it (but I could easily be wrong there) - looking at my code, I think this would be replacing this custom function:

```ts
export const getEvent = async (event: ApiEvent): Promise<GithubEvent> => {
  const rawBody = event.body ?? '{}';
  const signature = event.headers['X-Hub-Signature-256'] ?? '';
  const name = event.headers['X-GitHub-Event'];
  const { GH_WEBHOOK_SECRET } = process.env;

  if ((await verify(GH_WEBHOOK_SECRET, rawBody, signature)) && name) {
    const payload = JSON.parse(rawBody) as GithubEvent['payload'];

    return { name, payload } as GithubEvent;
  }

  throw new Error('event did not come from github');
};
```

which is pretty small and contained whereas this addition is twice the size - having said that, I'm pretty sure we could easily have `verifyAndReceive` use this function too which would cut most of the additional size to the package and make it a lot easier to justify

timoschinkel:
Just some unsolicited comment; I was actually looking for a function like this. I plan on running this logic in an AWS lambda environment, and having a nested handler seems a bit unnecessary.

> which is pretty small and contained whereas this addition is twice the size

I understand, but having the library offer this instead of having consumers write/copy this logic when needed is a much better developer experience.

wolfy1339:
The comment about the size of this addition should be moot, as I refactored the internals to use this new function in https://github.com/octokit/webhooks.js/commit/23e044335c779de9c58a4299887f0d2e73ac7f37

PR #1166, opened 2025-07-30
chore: avoid Object.assign

Uzlopak:
Resolves #ISSUE_NUMBER

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [ ] No

----

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -70,7 +70,7 @@ export interface State extends Options<any> {
  */
 export type WebhookError = Error &
   Partial<RequestError> & {
-    event?: EmitterWebhookEventWithStringPayloadAndSignature;
+    event?: EmitterWebhookEventWithStringPayloadAndSignature | undefined;
   };
 
 export interface AggregateWebhookError extends AggregateError {
--- a/src/types.ts
+++ b/src/types.ts
@@ -68,7 +68,10 @@ export interface State extends Options<any> {
 /**
  * Error object with optional properties coming from `octokit.request` errors
  */
-export type WebhookError = Error & Partial<RequestError>;
+export type WebhookError = Error &
+  Partial<RequestError> & {
+    event?: EmitterWebhookEventWithStringPayloadAndSignature;
+  };
 
 export interface AggregateWebhookError extends AggregateError {
   errors: WebhookError[];
--- a/src/verify-and-receive.ts
+++ b/src/verify-and-receive.ts
@@ -23,11 +23,12 @@ export async function verifyAndReceive(
   if (!matchesSignature) {
     const error = new Error(
       "[@octokit/webhooks] signature does not match event payload and secret",
-    );
+    ) as WebhookError;
 
-    return state.eventHandler.receive(
-      Object.assign(error, { event, status: 400 }) as WebhookError,
-    );
+    error.event = event;
+    error.status = 400;
+
+    return state.eventHandler.receive(error);
   }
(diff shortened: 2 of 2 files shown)


G-Rath (dismissed):
this change does more than just remove the `Object.assign` - could you retitle your PR to be either a `fix` or `feat`, and provide more explanation on why you think this change is needed (even better if you could add a test related to it)

Uzlopak:
Why is it doing more than removing Object.assign? It is the same behaviour than before. event and status are not non-wrtieable or non-configurable propertie on a generic Error. And Object.assign makes them also enumerable. Obviously Object.assign is slightls slower than just assigning the  properties.

G-Rath:
* you're adding a new property to the types
* you're making the properties not enumerable (which is technically a breaking change, but it's probably fine)
* performance improvements are not without actual numbers to back them up

I'm not saying this change shouldn't be landed, but there is a reason we have a PR template

Uzlopak:
Ok, I have now to check if assigning with Object.assign is different then directly assigning the properties. Thank you for this precious hint. No sarcasm or so. I am quiet impressed regarding this aspect I did not see.

Uzlopak:
@G-Rath 

I checked it, and directly assigning makes them also enumerable. So there is no breaking change. 


```js
const bla = { a: 1}

Object.assign(bla, { b: 2 });

bla.c = 3;

Object.defineProperty(bla, 'd', {
  value: 4,
  enumerable: false,
});

console.log(bla.propertyIsEnumerable('a')); // true
console.log(bla.propertyIsEnumerable('b')); // true
console.log(bla.propertyIsEnumerable('c')); // true
console.log(bla.propertyIsEnumerable('d')); // false
```

And yes, mutating is always faster. Object. assign is actually mutating the Object, thus if you would use a Proxy  you could see that the setters are triggered on the Object. 

The type change actually just demonstrates, that the properties were missing on the type. The Object.assign actually hid the the fact that those propertes were not documented.

wolfy1339:
This would be a bug fix for the types.

Is it really faster to not use Object.assign ?
At what point does it really matter?

Uzlopak:
You are right. We have other bottlenecks. Actually verification could/should be improved. 

When I started working on #1167 i opened my local project and saw these changes, and i thought, that I should propose it. Object.assign swallowed the issue of lack of typings. So this was the reason why i wrote that it avoids Object.assign. The slightly faster remark in my first answer was just a clarifying that it has also some kind of perf effect. But it was not my main target.

I can add back the Object.assign, if you like.

wolfy1339 (dismissed):
I believe that the fact that this fixes a type issue and fixes the fact that the properties were hidden from typescript and thus the type issue not detected, make this PR just fine

src/types.ts
```diff
@@ -68,7 +68,10 @@ export interface State extends Options<any> {
 /**
  * Error object with optional properties coming from `octokit.request` errors
  */
-export type WebhookError = Error & Partial<RequestError>;
+export type WebhookError = Error &
+  Partial<RequestError> & {
+    event?: EmitterWebhookEventWithStringPayloadAndSignature;
```
wolfy1339:
```suggestion
    event?: EmitterWebhookEventWithStringPayloadAndSignature | undefined;
```
Explicit undefined for optional types

wolfy1339 (approved):

PR #1199, opened 2026-05-09
fix: accept string event names in verifyAndReceive

### LINKED ISSUE #1055: [BUG]: Public interface types should expect strings

AmSach:
Fixed the bug described in issue #1055.

**What was wrong:**
The `EmitterWebhookEventWithStringPayloadAndSignature` interface had `name: string` (a plain, wide string type), while `EmitterWebhookEvent` had `name` as a union of literal webhook event names (`EmitterWebhookEventName`). When calling `verifyAndReceive` with an event whose `name` was a string literal (like `'push'` from HTTP headers), TypeScript couldn't narrow it to the literal union type needed for `receive()`, causing type errors even though the library accepted such values at runtime.

**How I fixed it:**
1. Changed `EmitterWebhookEventWithStringPayloadAndSignature.name` from `string` to `EmitterWebhookEventName` — the same union type used by `EmitterWebhookEvent`. This makes the type system consistent with what the library actually accepts at runtime.

2. Updated `createMiddleware` in `src/middleware/create-middleware.ts` to import and use `EmitterWebhookEventName` instead of the internal-only `WebhookEventName` for the header extraction. This removes the redundant `as any` cast that was masking the type mismatch.

**Files changed:**
- `src/types.ts` — `EmitterWebhookEventWithStringPayloadAndSignature.name` now typed as `EmitterWebhookEventName` instead of `string`
- `src/middleware/create-middleware.ts` — `getRequestHeader<EmitterWebhookEventName>` replacing the previous `WebhookEventName` import and cast

**Tested by:**
- TypeScript compilation passes with no errors (`npx tsc --noEmit` exits 0)
- Verified that calling `verifyAndReceive` with a plain string event name (as real HTTP headers would provide) works correctly — signature verification proceeds normally

Closes #1055

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2170,7 +2170,6 @@
       "integrity": "sha512-5gTmgEY/sqK6gFXLIsQNH19lWb4ebPDLA4SdLP7dsWkIXHWlG66oPuVvXSGFPppYZz8ZDZq0dYYrbHfBCVUb1Q==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=12"
       },
@@ -2213,7 +2212,6 @@
       "integrity": "sha512-I7AIg5boAr5R0FFtJ6rCfD+LFsWHp81dolrFD8S79U9tb8Az2nGrJncnMSnys+bpQJfRUzqs9hnA81OAA3hCuQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "prettier": "bin/prettier.cjs"
       },
@@ -2704,7 +2702,6 @@
       "integrity": "sha512-ytQKuwgmrrkDTFP4LjR0ToE2nqgy886GpvRSpU0JAnrdBYppuY5rLkRUYPU1yCryb24SsKBTL/hlDQAEFVwtZg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "esbuild": "~0.25.0",
         "get-tsconfig": "^4.7.5"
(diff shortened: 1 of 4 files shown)


wolfy1339:
The reason that it was set to string in the first place, is because you can pass any value and the runtime code will throw an error if it's invalid.

Headers from http request libraries are often only typed as string

